PR #3, opened 2018-10-02
Initial PAPR CI check-in

baude:
Test conmon binary build on Fedora 28 in a container.  Then make a SRPM from
the current source tree and rpmbuild it.  Finally, verify it installs cleanly
and the packaging worked.

Signed-off-by: baude <bbaude@redhat.com>

### DIFF
--- a/.papr.yml
+++ b/.papr.yml
@@ -1,0 +1,27 @@
+branches:
+  - master
+  - auto
+  - try
+
+container:
+    image: registry.fedoraproject.org/fedora:28
+    
+packages:
+    - make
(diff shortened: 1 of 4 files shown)


mheon:
This LGTM, but the tests don't seem to be running. Do we need to merge it to start them going, or is there other setup missing?

baude:
As the commit message says, it only tests build, build rpm, and install.  The actual CRIO and libpod tests will be follow on PRs

mheon:
I see the papr tests now - they weren't there when I first checked. Seems fine. Merging.

PR #5, opened 2018-11-11
Do not overflow local buffer

isimluk:
Do not use full width of buf, as we intend to assign

    buf[num_read] = '\0';

at the end of the buffer, afterwards. In case we have just read full BUF_SIZE,
we will write '\0' to a memory location that is outside of buf. This can be
referred to as CWE-119 although, it seems pretty harmless in this case.

Signed-off-by: Šimon Lukašík <isimluk@fedoraproject.org>

### DIFF
--- a/src/conmon.c
+++ b/src/conmon.c
@@ -1680,7 +1680,7 @@ int main(int argc, char *argv[]) {
        * Read from container stderr for any error and send it to parent
        * We send -1 as pid to signal to parent that create container has failed.
        */
-      num_read = read(masterfd_stderr, buf, BUF_SIZE);
+      num_read = read(masterfd_stderr, buf, BUF_SIZE - 1);
       if (num_read > 0) {
         buf[num_read] = '\0';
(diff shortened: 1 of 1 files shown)


rhatdan:
bot, add author to whitelist
bot, test pull request

rhatdan:
LGTM
@giuseppe @baude PTAL
We need this in the conmon that is under CRI-O also.

giuseppe:
@isimluk could you open the PR against https://github.com/kubernetes-sigs/cri-o/blob/master/conmon/conmon.c ?

The separate conmon project here is outdated, we will need to sync it anyway.

giuseppe:
in any case, the patch LGTM

isimluk:
> @isimluk could you open the PR against https://github.com/kubernetes-sigs/cri-o/blob/master/conmon/conmon.c ?

Sure.

isimluk:
I think this can be merged. PTAL.

PR #6, opened 2018-11-26
Cirrus: Add CRI-O Integration testing for RHEL

cevich:
Enable CRI-O Integration testing for RHEL and migrate PAPR package testing to Cirrus-CI.  This work is largely based on the same layout/construction of ``libpod`` testing under Cirrus-CI.  Both base-image and cache-image production is documented and functional.  The output images from both are utilized in testing for this PR.   Setup for running the integration tests under Fedora 28 and 29 is also included, however inactive.  They will be enabled via a separate PR.

***CIRRUS: REBUILD IMAGES***


mheon:
@cevich IMO, merge this once it's ready - the test suite looks fine on a once-over.
Might want @runcom to take a look and make sure the CRI-O bits look right though.

cevich:
Removed testing commit - though I expect tests will fail in ``network.bats`` again.  Once we're past that, there's another two or three failing tests, but otherwise the majority are passing.

cevich:
@mrunalp any idea why cri-o integration tests may be failing?  I'm pretty sure I correctly transcribed the Ansible tasks in bash, but perhaps I missed something?

cevich:
note to me:
    * Make sure packages are updated at test-time
    * Grab the VM and reproduce first failing test (then fix it)

rhatdan:
@cevich whats the scoop on this PR?

cevich:
@rhatdan It's the failure I e-mailed about.

Since then I've switched to using the official (released)  RHEL OCP 3.11 RPMs for cri-o, matching upstream source, but that same integration test still fails.  Chatted with @giuseppe and @runcom about it too, consensus was that we need to involve @dcbw (test author).

$10Bn question: How did / does this pass upstream?

Answer: :man_shrugging:   (it's possible it was not tested or did not pass and went unreported - ***All TBD***)

My inexperienced, gut-feeling guess: At this point I don't think there's an actual bug/problem in released code (until proven otherwise).  It feels like a test-bug at worse, fragile-test at best.

The good news is, I can readily reproduce VMs that reproduce the problem :smile:

cevich:
@baude @rhatdan 

:zap: :zap: **tests pass!** :zap: :zap:

Full-Transparency: [I disabled one broken test (by patching it away)](https://github.com/kubernetes-sigs/cri-o/issues/2030)

rhatdan:
LGTM

cevich:
From #9 

> but I would like an ack from @mrunalp or another of the core CRI-O maintainers to make sure they agree that the tests are ready... I'm not really comfortable saying that for them

**ping** :smile:

cevich:
@mrunalp @runcom @giuseppe PTAL.  Hint: The skeleton of this work is from what we're already using in libpod.  The major important differences are in ``lib.sh`` and ``integration_test.sh``.  All/most of the other details I obtained by manually "translating" steps from CRI-O's CI (in Ansible) into bash.

giuseppe:
LGTM, I don't think the failing test has anything to do with this PR

mrunalp:
LGTM

mrunalp:
@rhatdan can you merge this? Also, @giuseppe and I need merge rights here.

cevich:
Thanks guys.  I will proceed with adding support for testing Fedora, then follow up with additional tests next week.  For now I'm finding my self lost in Golang-land and going deep with some BDD to help along the way.

PR #9, opened 2019-02-13
Fix Makefile to rebuild when sources are touched (and other fixes)

filbranden:
Dependencies were not fully specified, so touching a source file didn't trigger a rebuild and `make clean` was necessary; fix that.

Furthermore, use `$(CFLAGS)` and `$(LDFLAGS)` correctly. Don't pass a header file to the linker step, use a dependency-order rule (GNU make specific) for creating the target bin/ directory and mark a few more rules as `.PHONY:`.

Also remove executable bit from a header file and update `.gitignore` to include the path to the built `conmon` executable.


rhatdan:
LGTM
@giuseppe PTAL

giuseppe:
bot, add author to whitelist

@rh-atomic-bot r+

giuseppe:
I am worried about this repo not being in sync with CRI-O, where the "real" implementation of conmon is.

@mrunalp are you fine if I set conmon in the CRI-O repo to be a git submodule (and update the code here)?

rhatdan:
I think we need to do this.  Are we running the full CRI-O Test suite against conmon still?

giuseppe:
@rhatdan yes, we do that

mrunalp:
@giuseppe Sounds good if that allows us to still open PRs to conmon in CRI-O to be able to run full suite and update the repo here automatically.

giuseppe:
that won't be possible, only to update the submodule itself (and possibly make it point to somewhere else).

Any other idea on how we can improve this?  If conmon is so tied to CRI-O maybe it makes no sense to have its own repository?

Or alternatively, but this requires more work, should we run CRI-O tests on conmon PRs?

mheon:
@cevich is working on getting the CRI-O tests running over here, but there were obstacles - some of the tests weren't passing

cevich:
@mheon @giuseppe Yep, [over here](#6) :smile:  I managed to get it down to a single failing integration test, but couldn't get adequate support for debugging/fixing it, so I disabled it (here).  I have cards open to work on adding the other cri-o tests (here) but didn't want to continue given the unknown/uncertain overall status of this repo.

IMHO and design-wise, having conmon in it's own repo. makes _way_ more sense than having it tied to cri-o.  However, I'm not so much a fan of porting the cri-o (or podman) tests for use here.  That's not going to be long-term sustainable.  Ideally conmon should have it's own dedicated / specialized set of tests.

cevich:
Idea:  as a one-off, I could cherry-pick this into #6 to see if the tests pass.  Let me know if that would be helpful (it's not asking for much work).

filbranden:
My 2c:

Even without integration tests here, it *should* be fine to turn conmon into a git submodule in the cri-o repository.

Because when an updated conmon is pushed there, that is done through a PR and that will end up running the full cri-o tests against the updated conmon.

The risk there is that one such update will fail, in which case we'll have to come back to conmon and either revert the commits or push other fixes to keep the integration with cri-o working. So the worst case is "more churn." But I wouldn't expect this to be so bad, given I don't really expect that many changes around here to be breaking the cri-o tests all the time...

Anyways, I'm happy to help however I can, just let me know :smile: 

Cheers,
Filipe

cevich:
> hat is done through a PR and that will end up running the full cri-o tests against the updated conmon.

@filbranden unless I not understanding the workflow, there would also need to be a PR opened in CRI-O that updates the submodule sha.  Then it would get tested :(

Though I seem to remember a way to make git submodules always grab the latest head of a branch.  I can't remember for sure on that.

cevich:
> such update will fail, in which case we'll have to come back to conmon

Also a desirable thing to avoid.  So having cri-o (and podman) pin a conmon git submodule down to a specific commit, is also not a terrible idea.

Still, it would be good to at least have some even-minimal CI setup here, to catch the "simple" mistakes :smile:

filbranden:
@cevich Yes there would be a PR on CRI-O, once all the changes to conmon are merged/submitted and it's ready for integration back to CRI-O. Yes, it would get tested then, but I think that is *good* because we only start using the new conmon changes in CRI-O once they've been actually tested there...

As far as I can tell, git submodule will always pin to a specific SHA-1. Which, I argue, is the right thing to do. (In that sense, it's not much different from how Go dependencies work, with godep or vndr.)

I agree that it's bad having to come back here and fix some mistakes we only realize at integration time... But given the size of this project and I don't really expect huge refactors on it right now, I think it's *probably* fine, I'd expect us to "get it right" most of the time and integration test in CRI-O will just confirm that...

Of course, having some CI here would be great and the more the better. Perhaps pushing the subset of tests you have that are currently passing would be a great start already. Later on, if we do find problems integrating back to CRI-O, we can always come back and add more specific tests here to prevent that problem from breaking us again and iterate from there...

cevich:
@filbranden yep on all accounts, we're listening to the same music.

With it passing using a subset of tests, there's not much else but a "/lgtm" holding back #6

Best guess based on what I heard from Matt/Brent is "time on more urgent tasks" is probably the only thing in the way.  So...could just be we need to do more "hurry up and wait" :smiley:

mheon:
@cevich Honestly, I'm fine merging #6 whenever, but I would like an ack from @mrunalp or another of the core CRI-O maintainers to make sure they agree that the tests are ready... I'm not really comfortable saying that for them

cevich:
@mheon oh ok, I completely understand.  I'll ping 'em in that PR.

giuseppe:
@rhatdan seems like my r+ didn't work.  Either I am not an admin or the bot ignored me :-)

Could you try?

mrunalp:
@rh-atomic-bot r+

PR #10, opened 2019-02-13
Add Meson build configuration

filbranden:
Replace the rules in a Makefile with a Meson based build.

The build is essentially equivalent, except that the target directory for the conmon binary is `$(LIBEXECDIR)/podman` instead of `$(LIBEXECDIR)/crio`.

---

This is in some ways an alternative to #9 (even though merging both is probably fine.)

Getting it right on Makefiles is just too hard... Meson is much nicer. (If using `make`, at least adopting automake would make the Makefiles more reliable, but Meson is still a much cleaner approach!)

This is somewhat incomplete, since the `.rpmbuild/Makefile` and `*.spec.in` need updating, and the change needs a counterpart in the Fedora RPM for podman... But let me know your initial thoughts, I'd be happy to update those as well if you like the idea of adopting Meson.

Cheers,
Filipe

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,0 +1,48 @@
+
+project('conmon', 'c',
+        version : '0.1',
+        license : 'Apache-2.0',
+        default_options : [
+                'c_std=c99',
+        ],
+        meson_version : '>= 0.46',
+)
+
+git_commit = ''
+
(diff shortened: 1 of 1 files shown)


rhatdan:
bot, add author to whitelist

PR #12, opened 2019-02-22
journald log driver

haircommander:
Changed the log-path to allow for a colon delimited <DRIVER>:<PATH>, with two options for the <DRIVER>:
k8s-file: Write output to <PATH> logs in kubernetes .log format (default, mimics old behavior)
journald: Write output to systemd journal, and ignore <PATH>
    Note: journald output can be filtered by container ID, as it is printed in the journal's MESSAGE_ID field.
    i.e: 'journalctl MESSAGE_ID=<ctr_id>'
both k8s-file and journald can be flagged
Added conditional code in Makefile to allow for conmon to compile on systems without libsystemd. Flagging --log-path=journald: will error if libsystemd isn't found during compilation.
Added ctr_logging and utils libraries. Ctr_logging will now be the home for all things related to container logging. Utils was added as the shared library between conmon and ctr_logging.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,21 @@ override LIBS += $(shell pkg-config --libs glib-2.0)
 CFLAGS ?= -std=c99 -Os -Wall -Wextra
 override CFLAGS += $(shell pkg-config --cflags glib-2.0) -DVERSION=\"$(VERSION)\" -DGIT_COMMIT=\"$(GIT_COMMIT)\"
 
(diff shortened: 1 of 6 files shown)


haircommander:
@rhatdan @giuseppe PTAL, now that this is merged into CRI-O, I'd like to get it into our other tools as well

rhatdan:
LGTM

haircommander:
closing per #16

PR #15, opened 2019-04-03
Fix CRI-O repo. references and image names

cevich:
Signed-off-by: Chris Evich <cevich@redhat.com>


haircommander:
LGTM, I would merge but I don't have write access here :\

PR #13, opened 2019-02-26
Cirrus: Enable CRI-O intg. testing w/ Fedora

cevich:

PR #16, opened 2019-04-04
Update to cri-o

haircommander:
c/conmon has fallen quite behind upstream cri-o. Let's catch up :)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,21 @@ override LIBS += $(shell pkg-config --libs glib-2.0)
 CFLAGS ?= -std=c99 -Os -Wall -Wextra
 override CFLAGS += $(shell pkg-config --cflags glib-2.0) -DVERSION=\"$(VERSION)\" -DGIT_COMMIT=\"$(GIT_COMMIT)\"
 
-bin/conmon: src/conmon.o src/cmsg.o | bin
+# Conditionally compile journald logging code if the libraries can be found
+# if they can be found, set USE_JOURNALD macro for use in conmon code.
+#
+# "pkg-config --exists" will error if the package doesn't exist. Make can only compare
+# output of commands, so the echo commands are to allow pkg-config to error out, make to catch it,
+# and allow the compilation to complete.
+ifeq ($(shell pkg-config --exists libsystemd-journal && echo "0" || echo "1"), 0)
+	override LIBS += $(shell pkg-config --libs libsystemd-journal)
+	override CFLAGS += $(shell pkg-config --cflags libsystemd-journal) -D USE_JOURNALD=0
+else ifeq ($(shell pkg-config --exists libsystemd && echo "0" || echo "1"), 0)
+	override LIBS += $(shell pkg-config --libs libsystemd)
+	override CFLAGS += $(shell pkg-config --cflags libsystemd) -D USE_JOURNALD=0
+endif
+
+bin/conmon: src/conmon.o src/cmsg.o src/ctr_logging.o src/utils.o | bin
 	$(CC) $(LDFLAGS) -o $@ $^ $(LIBS)
 
 %.o: %.c
@@ -24,7 +38,11 @@ bin/conmon: src/conmon.o src/cmsg.o | bin
 
 src/cmsg.o: src/cmsg.c src/cmsg.h
 
-src/conmon.o: src/conmon.c src/cmsg.h src/config.h
+src/utils.o: src/utils.c src/utils.h
+
+src/ctr_logging.o: src/ctr_logging.c src/ctr_logging.h src/utils.h
+
+src/conmon.o: src/conmon.c src/cmsg.h src/config.h src/utils.h src/ctr_logging.h
 
 bin:
 	mkdir -p bin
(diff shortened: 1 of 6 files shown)


TomSweeneyRedHat:
@haircommander you've some test issues.

haircommander:
@TomSweeneyRedHat ah yes here is the trouble with testing a tool with tools that depend on it that @cevich was talking about. I have a fix in mind for this, but in the meantime, where is the cri-o version used in cirrus defined?

TomSweeneyRedHat:
@cevich, you're up to bat for @haircommander 's cri-o version in cirrus question.

cevich:
For this, since RHEL is used for CI, and because of OpenShift / cri-o version locking, we [handle it at runtime according to what's installed.](https://github.com/containers/conmon/blob/master/contrib/cirrus/lib.sh#L266-L275)

I'm planning something similar for Fedora using modularity.  There's a little bit of [monkey-patching need also](https://github.com/containers/conmon/blob/master/contrib/cirrus/integration_test.sh#L17-L21) since that test is broken in this particular version.  Though not ideal (we REALLY need dedicated/specialized conmon-tests here), this is the only way I could find to get the cri-o integration tests to pass :confused: 

If you have a better idea, I'm open to suggestions.  Took weeks to get ^^^ working reliably :confounded:

haircommander:
@cevich my plan is to tackle conmon unit tests next sprint actually. You'll be gone so that'll give me some time to get something vaguely comprehensive. Then when you get back I'll have a nice plate of tests you can hand out to our cirrus bots :)

cevich:
Lol, here's a hair-brained thought...I bet the changes you've got here will make my #13 pass (while simultaneously breaking RHEL).  As a test, try this (awful, unholy) thing:

```diff
diff --git a/.cirrus.yml b/.cirrus.yml
index fc3f108..a39d0d1 100644
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -20,8 +20,8 @@ env:
     ####
     #### Cache-image names to test with
     ###
-    PRIOR_RHEL_CACHE_IMAGE_NAME: "rhel-7-conmon-34e54ce8"
-    FEDORA_CACHE_IMAGE_NAME: "fedora-29-conmon-34e54ce8"
+    PRIOR_RHEL_CACHE_IMAGE_NAME: 'fedora-29-conmon-75ea13be'
+    FEDORA_CACHE_IMAGE_NAME: 'fedora-29-conmon-75ea13be'
     #PRIOR_FEDORA_CACHE_IMAGE_NAME: "fedora-28-notready"
     #RHEL_CACHE_IMAGE_NAME: "rhel-8-notready"
     #CENTOS_CACHE_IMAGE_NAME: "centos-7-notready"
```

haircommander:
like that?

cevich:
Ya, no idea if it'll work :confused: 

I have half a mind to drop running RHEL here all together, but that could suck if we ever want to test a needed backport for cri-o.  Either way, being pinned to the Fedora version of cri-o is probably much better for upstream development (here) than using RHEL.

We'll figure something out in any case.  Though heads-up: I'm on PTO for the next two weeks.

cevich:
Eek! gosh, okay, well that most certainly did not work.  If you can make heads/tails why, more power to you.  All I see is a wall of unintelligible gobbledygook :confounded: 

Instead of using RPMs, if we pin conmon to a hard-coded cri-o SHA, then the CI here must compile cri-o.  Based on my experience with libpod, using the RPMs is much better from a dependency-tracking and real-world compatibility standpoint.

Honestly I don't have any good solutions here I'm afraid.  Did I mention it would be nice to only rely on some specialized conmon-tests? :smile:

haircommander:
I think conmon tests are the move

cevich:
If you have a mind to, please.  I'm not beholden to the current conmon CI setup in any way/form.  In fact, running specialized CI [tests direct from a docker file would be a fine and very simple solution.](https://cirrus-ci.org/guide/docker-builder/#dockerfile-as-a-ci-environment).  We could get rid of all the VMs and image-building and related scripts.  Would be fast.  Easy to maintain.  And trivial to expand.

haircommander:
I think I'll need podman to create the containers that will then have be run directly with conmon with both runc and crun. Basically: check if conmon is being a good translator for image->runtime. I'll try to get that going with a dockerfile

haircommander:
For the time being, I fixed the backwards compatibility problem. tests should pass and I'd like if this merged. PTAL @rhatdan @giuseppe @mrunalp @mheon

rhatdan:
LGTM

rhatdan:
Well this looks good.
 find /sys/kernel/ | wc
find: '/sys/kernel/security': Permission denied
find: '/sys/kernel/debug': Permission denied
find: '/sys/kernel/config': Permission denied
   5615    5615  213373
So I think for rootless we can ignore the mounts.

rhatdan:
Wrong PR

rhatdan:
@giuseppe @mrunalp @mheon PTAL

haircommander:
@rhatdan @mheon needs manual merge, no bot here :)

haircommander:
(and I don't have write access)

mheon:
Done

PR #18, opened 2019-05-10
Add VERSION file

haircommander:
This PR comes in anticipation of conmon exec. I'd like to get podman to use conmon here rather than in cri-o, and we should really start using some semantic versioning for conmon.

Because conmon exec will almost definitely include breaking API changes, we will want to tag a new version once I finish with that.

I was unsure of whether to call what currently exists 0.1.0 or 1.0.0, and am willing to do either. I'd certainly consider conmon to have a stable API as of now, but the meson.build labels it as 0.1.0.

because podman remote is blocked on conmon exec, I am deferring adding unit tests to this repository until then, but I have figured out a way to do it without relying on podman/CRI-O.


meson.build
```diff
@@ -1,6 +1,6 @@
 
 project('conmon', 'c',
-        version : '0.1',
+        version : '0.1.0',
```
rhatdan: Is there any way to get this in a different way so we don't need to define it twice?

Something like
```
version = run_command(
    'cat', files('VERSION'),
).stdout().strip()


project('conmon', 'c',
        version : version,
        license : 'Apache-2.0',
        default_options : [
                'c_std=c99',
        ],
        meson_version : '>= 0.46',
)
...
```

rhatdan: That is what googling of mesa got me.

@saschagrunert PTAL

haircommander: good call, I modified it (and tested it)

rhatdan:
LGTM

rhatdan:
@giuseppe @mheon @vrothberg PTAL

mheon:
LGTM

haircommander:
@rhatdan manual merge please :)

PR #21, opened 2019-05-17
Try to fix CI

haircommander:
Instead of basing off of RPM version, try manually tagging current CRI-O version


contrib/cirrus/lib.sh
```diff
@@ -267,8 +267,10 @@ match_crio_tag() {
     req_env_var "
         GOSRC $GOSRC
     "
-    export CRIO_VER="$(rpm -q --qf '%{V}' cri-o)"
-    echo "Checking out CRI-O tag v$CRIO_VER to match installed rpm $(rpm -q cri-o)"
+    # export CRIO_VER="$(rpm -q --qf '%{V}' cri-o)"
+    # echo "Checking out CRI-O tag v$CRIO_VER to match installed rpm $(rpm -q cri-o)"
+    # instead of going off of an old RPM, update to latest and greatest CRI-O
+    export CRIO_VER="1.13.8"
```
vrothberg: Can you elaborate what the problem with extracting the version from the package is? Both approaches seem a bit exotic. Intuitively, I'd always test the latest release or even the latest commit of CRI-O to detect regressions as early as possible. But I may be missing an important part.

haircommander: @vrothberg the problem is CRI-O is originally downloaded vis rpm, and has proven to be 1.11. I was trying to update it but I need a different approach, as the tests ans binary are mismatched if i do what is here

PR #23, opened 2019-05-20
Remove Makefile git shell commands

saschagrunert:
The git commands leads into annoying failures when not running within a
git repository (only on downloaded sources). It should be fine to remove
them since they're not used anywhere.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,3 @@
-include Makefile.inc
-
-GIT_BRANCH := $(shell git rev-parse --abbrev-ref HEAD 2>/dev/null)
-GIT_BRANCH_CLEAN := $(shell echo $(GIT_BRANCH) | sed -e "s/[^[:alnum:]]/-/g")
 VERSION := $(shell cat VERSION)
 PREFIX ?= ${DESTDIR}/usr/local
 BINDIR ?= ${PREFIX}/bin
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM

haircommander:
we have atomic bot here?

bot, add author to whitelist

rhatdan:
bot, add author to whitelist

rhatdan:
bot, test pull request

haircommander:
the git commit does print in conmon version, which could be useful in the future, so I'm hesitant to remove them. I do see what you mean about it being annoying, though.

saschagrunert:
> the git commit does print in conmon version, which could be useful in the future, so I'm hesitant to remove them. I do see what you mean about it being annoying, though.

Okay, so I could define a function within the Makefile which pre-checks if git is available I think the same would apply for the `Makefile.inc`. WDYT?

haircommander:
@saschagrunert perfect SGTM

saschagrunert:
I changed the implementation to be a bit more flexible, not sure if we want to stick to the `unknown` part but this actually works. :)

haircommander:
btw how did you run into this? Changes LGTM

saschagrunert:
> btw how did you run into this? Changes LGTM

I played around with packaging the project into a RPM and used the source code tarball which does not include the repository. I think we have the same issue in CRI-O as well, so we cannot reuse the makefile targets during packaging. I guess we can fix it the same way like we do here. :)

haircommander:
@saschagrunert that stinks because the case of shipping is the most vital as its the typical case where a user can encounter a problem and we'd want to know which version of conmon they're using.

saschagrunert:
> @saschagrunert that stinks because the case of shipping is the most vital as its the typical case where a user can encounter a problem and we'd want to know which version of conmon they're using.

Hm, I think in this case here we still would have the `VERSION` information available. I think we should do it like in CRI-O and always use a `VERSION` with `-dev` suffix and override it when needed (maybe via `make VERSION=0.2.0`?)

I updated the version to be 0.2.0-dev.

haircommander:
Ok i'd like to get a couple more prs in, then i'll make a commit bumping to 0.2.0, then i'll merge this. Klingt gut?

saschagrunert:
> Ok i'd like to get a couple more prs in, then i'll make a commit bumping to 0.2.0, then i'll merge this. Klingt gut?

Klingt sehr gut! Thanks :)

saschagrunert:
Rebased on top of the latest master branch.

PR #26, opened 2019-05-21
Add .clang-format file

saschagrunert:
This PR adds the .clang-format file to the repository, which
includes all available Options based on the style from the CRI-O
repository.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,116 @@
+---
+Language:        Cpp
+# BasedOnStyle:  LLVM
+AccessModifierOffset: -2
+AlignAfterOpenBracket: Align
+AlignConsecutiveAssignments: false
+AlignConsecutiveDeclarations: false
+AlignEscapedNewlines: DontAlign
+AlignOperands:   true
+AlignTrailingComments: true
+AllowAllParametersOfDeclarationOnNextLine: false
+AllowShortBlocksOnASingleLine: false
+AllowShortCaseLabelsOnASingleLine: false
+AllowShortFunctionsOnASingleLine: false
+AllowShortIfStatementsOnASingleLine: false
+AllowShortLoopsOnASingleLine: false
+AlwaysBreakAfterDefinitionReturnType: None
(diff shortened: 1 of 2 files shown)


.clang-format
```diff
@@ -0,0 +1,116 @@
+---
+Language:        Cpp
```
haircommander: Isn't the language C?

saschagrunert: Yes, but clang-format does not have this option:

```
Language (LanguageKind)
Language, this format style is targeted at.

Possible values:

LK_None (in configuration: None) Do not use.
LK_Cpp (in configuration: Cpp) Should be used for C, C++.
LK_CSharp (in configuration: CSharp) Should be used for C#.
LK_Java (in configuration: Java) Should be used for Java.
LK_JavaScript (in configuration: JavaScript) Should be used for JavaScript.
LK_ObjC (in configuration: ObjC) Should be used for Objective-C, Objective-C++.
LK_Proto (in configuration: Proto) Should be used for Protocol Buffers (https://developers.google.com/protocol-buffers/).
LK_TableGen (in configuration: TableGen) Should be used for TableGen code.
LK_TextProto (in configuration: TextProto) Should be used for Protocol Buffer messages in text format (https://developers.google.com/protocol-buffers/).
```

So we have to use `Cpp`, but this should not harm.

haircommander: :+1:

haircommander:
Can you add a make target make lint or something to run these checks?

haircommander:
other than that LGTM

haircommander:
also needs rebase to fix CI :)

saschagrunert:
> Can you add a make target make lint or something to run these checks?

Added a `make fmt` test target to be run in the CI. I guess we could also run clang static analysis or cppcheck later on in the CI so I decided to stick to `fmt`.

> also needs rebase to fix CI :)

Rebased on top of the latest master branch.

rhatdan:
LGTM

PR #35, opened 2019-05-28
Add podman and crio make target

haircommander:
Users can now directly install to a "podman" directory, instead of needing to install to crio, then link to podman

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -97,7 +97,7 @@ fedora_packaging_task:
         - make -f .rpmbuild/Makefile
         - rpmbuild --rebuild conmon-*.src.rpm
         - dnf -y install ~/rpmbuild/RPMS/x86_64/conmon*.x86_64.rpm
-        - ls -l /usr/libexec/crio/conmon
+        - ls -l /usr/bin/conmon
 
     timeout_in: '20m'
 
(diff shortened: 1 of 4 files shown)


rhatdan:
Can we just install it to /usr/bin/conmon?

haircommander:
@rhatdan that sounds like it should be a different make target. Unless we're not worried about backwards compatibility (for instance libpod.conf will need to add a different place to look for conmon)

saschagrunert:
> @rhatdan that sounds like it should be a different make target. Unless we're not worried about backwards compatibility (for instance libpod.conf will need to add a different place to look for conmon)

Would creating a symlink be an option? I mean so we would avoid copying stuff around.

rhatdan:
We eventually want to get conmon to be its own package so it should be installed in a location where any tool could use it.  /usr/sbin/conmon for example.   I would think this is the default location, then we adjust libpod.conf to look to this location.

rhatdan:
I would like to reverse the dependency here. conmon should be independant of CRI-O and Podman, Right now the dependency seems backwords.

haircommander:
I went for $(BINDIR) for make install, and left make podman and make crio. PTAL @rhatdan @giuseppe @saschagrunert

PR #36, opened 2019-05-29
Add warning if read events failed

saschagrunert:
We now add a warning if the read of events fd failed. This also fixes a warning that the return value of `read` is not considered in that place.

### DIFF
--- a/src/conmon.c
+++ b/src/conmon.c
@@ -572,7 +572,10 @@ static gboolean oom_cb_cgroup_v2(int fd, GIOCondition condition, G_GNUC_UNUSED g
 	gboolean ret = G_SOURCE_REMOVE;
 
 	/* Drop the inotify events.  */
-	read(fd, &events, sizeof(events));
+	ssize_t num_read = read(fd, &events, sizeof(events));
+	if (num_read < 0) {
+		pwarn("Failed to read events");
+	}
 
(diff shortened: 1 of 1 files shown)


src/conmon.c
```diff
@@ -572,7 +572,10 @@ static gboolean oom_cb_cgroup_v2(int fd, GIOCondition condition, G_GNUC_UNUSED g
 	gboolean ret = G_SOURCE_REMOVE;
 
 	/* Drop the inotify events.  */
-	read(fd, &events, sizeof(events));
+	ssize_t num_read = read(fd, &events, sizeof(events));
+	if (num_read < 0) {
+		nwarn("Failed to read events");
```
vrothberg: Can we use `pwarn` instead? `pwarn` is using strerror(3) to translate errno into something human readable, which could come in handy for debugging.

saschagrunert: Yes sure, I adapted it as suggested.

vrothberg:
@giuseppe @haircommander PTAL

vrothberg:
LGTM

haircommander:
LGTM, @giuseppe was there a reason you opted to drop the warning for v2?

giuseppe:
> LGTM, @giuseppe was there a reason you opted to drop the warning for v2?

I didn't add it because the event itself is not used, we are only draining the fd.

PR #39, opened 2019-06-08
fix: separate DESTDIR variable from PREFIX

CameronNemo:
Build systems that specify PREFIX typically do not include
the DESTDIR value.
So, the DESTDIR value should be used independently to ensure
that it is used at all.


haircommander:
LGTM

rhatdan:
LGTM

PR #41, opened 2019-06-11
Add cmd/config/config.go

haircommander:
Now, we can configure conmon from within the conmon directory.

This will allow tools to vendor conmon and import the values they need, rather than having the configuration parameters in different repositories

This PR also comes with the eventual goal of coupling containers/conmon with the go code that calls it, so libraries that use conmon don't have to worry too much about conflicting versions

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -20,9 +20,11 @@ env:
     TRAVIS: "true"
 
     ####
-    #### Cache-image names to test with
+    #### Image names to test with
     ###
     FEDORA_CACHE_IMAGE_NAME: 'fedora-29-conmon-75ea13be'
+    FEDORA_CONTAINER_FQIN: 'registry.fedoraproject.org/fedora:29'
+    PRIOR_FEDORA_CONTAINER_FQIN: 'registry.fedoraproject.org/fedora:28'
 
     ####
     #### Variables for composing new cache-images (used in PR testing) from
@@ -63,6 +65,9 @@ timeout_in: '120m'
 # testing for every platform
 integration_task:
 
+    depends_on:
+        - 'config'
+
     gce_instance:
         # Generate multiple parallel tasks, covering all possible
         # 'matrix' combinations.
@@ -85,8 +90,8 @@ fedora_packaging_task:
     # Runs within Cirrus's "community cluster"
     container:
         matrix:
-            image: "registry.fedoraproject.org/fedora:28"
-            image: "registry.fedoraproject.org/fedora:29"
+            image: "${FEDORA_CONTAINER_FQIN}"
+            image: "${PRIOR_FEDORA_CONTAINER_FQIN}"
         cpu: 4
         memory: 12
 
@@ -101,6 +106,26 @@ fedora_packaging_task:
 
     timeout_in: '20m'
 
+# Verify calls to bin/config were saved
+config_task:
+    env:
+        GOSRC: $CIRRUS_WORKING_DIR
+    # Runs within Cirrus's "community cluster"
+    container:
+        matrix:
+               # fedora:28 doesn't have go mod by default
+               # and we should only need one check to make sure
+               # config changes were synced
+            image: "${FEDORA_CONTAINER_FQIN}"
+        cpu: 4
+        memory: 12
+
+    script:
+        - dnf install -y make glib2-devel git gcc rpm-build golang
+        - cd $CIRRUS_WORKING_DIR
+        - GO111MODULE=on go mod init github.com/containers/conmon
+        - make config
+        - ./hack/tree_status.sh
 
 # Test building of new cache-images for future PR testing, in this PR.
 # Output images will be stored only for a very short time, then automaticly deleted.
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,10 @@ VERSION := $(shell cat VERSION)
 PREFIX ?= /usr/local
 BINDIR ?= ${PREFIX}/bin
 LIBEXECDIR ?= ${PREFIX}/libexec
+GO ?= go
+PROJECT := github.com/containers/conmon
+
+
 
 .PHONY: all git-vars
 all: git-vars bin bin/conmon
@@ -43,6 +47,10 @@ bin/conmon: src/conmon.o src/cmsg.o src/ctr_logging.o src/utils.o | bin
 
 %.o: %.c
 	$(CC) $(CFLAGS) -o $@ -c $<
+
+config: git-vars cmd/conmon-config/conmon-config.go runner/config/config.go runner/config/config_unix.go runner/config/config_windows.go
+	$(GO) build $(LDFLAGS) -tags "$(BUILDTAGS)" -o bin/config $(PROJECT)/cmd/conmon-config
+		( cd src && $(CURDIR)/bin/config )
 
 src/cmsg.o: src/cmsg.c src/cmsg.h
 
--- a/.gitignore
+++ b/.gitignore
@@ -51,3 +51,5 @@ modules.order
 Module.symvers
 Mkfile.old
 dkms.conf
+
+bin/
--- a/cmd/conmon-config/conmon-config.go
+++ b/cmd/conmon-config/conmon-config.go
@@ -1,0 +1,24 @@
+package main
+
+import (
+	"fmt"
+	"io/ioutil"
+
+	"github.com/containers/conmon/runner/config"
+)
+
+func main() {
+	output := `
+#if !defined(CONFIG_H)
+#define CONFIG_H
+
+#define BUF_SIZE %d
+#define STDIO_BUF_SIZE %d
+#define DEFAULT_SOCKET_PATH "%s"
+
+#endif // CONFIG_H
+`
+	if err := ioutil.WriteFile("config.h", []byte(fmt.Sprintf(output, config.BufSize, config.BufSize, config.ContainerAttachSocketDir)), 0644); err != nil {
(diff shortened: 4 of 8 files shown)


oci/oci_config.go
```diff
@@ -0,0 +1,8 @@
+package oci
+
+const (
+	// BufSize is the size of buffers passed in to sockets
+	BufSize = 8192
+	// TODO Fix for unsupported (windows)
```
saschagrunert: Is this still a valid TODO?

Makefile
```diff
@@ -2,6 +2,10 @@ VERSION := $(shell cat VERSION)
 PREFIX ?= /usr/local
 BINDIR ?= ${PREFIX}/bin
 LIBEXECDIR ?= ${PREFIX}/libexec
+GO ?= go
+PROJECT := github.com/containers/conmon
+
```
saschagrunert: I would add here:
```suggestion
BUILDTAGS := containers_image_ostree_stub
```

cmd/config/config.go
```diff
@@ -0,0 +1,26 @@
...
+#if !defined(CONFIG_H)
+#define CONFIG_H
+
+#define BUF_SIZE %d
+#define STDIO_BUF_SIZE %d
+#define DEFAULT_SOCKET_PATH "%s"
+
+#endif // CONFIG_H
+`
+	if err := ioutil.WriteFile("config.h", []byte(fmt.Sprintf(output, oci.BufSize, oci.BufSize, lib.ContainerAttachSocketDir)), 0700); err != nil {
```
saschagrunert: I would create that file with 0644 since it is no executable.

cmd/config/config.go
```diff
@@ -0,0 +1,26 @@
+package main
+
+import (
+	"fmt"
+	"io/ioutil"
+
+	"github.com/cri-o/cri-o/lib"
```
saschagrunert: We need to vendor these dependencies, right?

Makefile
haircommander: why that tag in particular?

saschagrunert: Because it does not build on my machine without the dependency installed :innocent:

haircommander: this patch or cri-o? if cri-o, I meant to not have any cri-o imports in this repo (first version of this PR was a bit sloppy :\ ), and the code here should not have containers/image associated with it

cmd/config/config.go
haircommander: I didn't even mean to have these dependencies...

Makefile
saschagrunert: Ah alright, so then we do not need it. Yes I guess it's because of CRI-O.

oci/oci_config.go
haircommander: nope :D

haircommander:
I added a CI check for a dirty config (one that was changed in oci/oci_config* but the config wasn't added to the PR). PTAL @cevich

haircommander:
@saschagrunert @rhatdan @mheon PTAL

mheon:
The C and Go look fine. I can't comment on the testing YAML or bash bits.

cevich:
looking...

.cirrus.yml
```diff
@@ -101,6 +101,26 @@ fedora_packaging_task:
...
+            image: "registry.fedoraproject.org/fedora:29"
+        cpu: 4
+        memory: 12
+
+    script:
+        - dnf install -y make glib2-devel git gcc rpm-build golang
+        - cd $CIRRUS_WORKING_DIR
+        - GO111MODULE=on go mod init github.com/containers/conmon
+        - make config
+        - ./hack/tree_status.sh
```
cevich: My preference would be to have these commands inside something like `contrib/cirrus/config_verify.sh`.  Not only to cut down the number of lines in the YAML (it's quite big), but also because then running the script manually becomes possible if/when needed.

cevich: (deleted)

.cirrus.yml
```diff
@@ -101,6 +101,26 @@ fedora_packaging_task:
...
+config_task:
+    env:
+        GOSRC: $CIRRUS_WORKING_DIR
+    # Runs within Cirrus's "community cluster"
+    container:
+        matrix:
+               # fedora:28 doesn't have go mod by default
+               # and we should only need one check to make sure
+               # config changes were synced
+            image: "registry.fedoraproject.org/fedora:29"
```
cevich: Adding another container image will become a maintenance and debugging burden. I suggest adding any packages you need into `contrib/gate/Dockerfile`.  Then re-use "quay.io/libpod/gate:latest" here, which is automatically built.  The default entrypoint (literally `entrypoint.sh`) runs `make $@` (as in gating task above).

.cirrus.yml
```diff
@@ -101,6 +101,26 @@ fedora_packaging_task:
 
     timeout_in: '20m'
 
+# Verify calls to bin/config were saved
+config_task:
```
cevich: Need to set up dependencies here and in the other tasks (to depend on this one passing).

cevich: Let's do:

```yaml
config_task:

    depends_on:
        - "gating"
```

then add `config` (no "task" suffix) into the dependency lists for:

* meta_task
* testing_task
* success_task

.cirrus.yml
cevich: Note: This will make PR testing break, so we'll need to merge manually.  There's no automated workflow for testing auto-built images, but as-is we can see it working (until you fix it).  Make sense?

cevich: "this will make"...meaning, my suggested change.

.cirrus.yml
haircommander: we don't a gating task as of now, will this change add one?

haircommander: nor meta, testing, or success tasks. I could add the dependency to integration and and fedora_packaging tasks?

cevich:
Ugggg....my wires are totally crossed, taking a fresh look...

cevich:
Okay, remembering now this is actually the conmon repo.  :smile: I think my maintenance concern is still valid, let's see if this can be made more "DRY" (untested - not sure this will work):

```diff
diff --git a/.cirrus.yml b/.cirrus.yml
index 07c7b1c..3c9fb59 100644
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -20,9 +20,11 @@ env:
     TRAVIS: "true"
 
     ####
-    #### Cache-image names to test with
+    #### Image names to test with
     ###
     FEDORA_CACHE_IMAGE_NAME: 'fedora-29-conmon-75ea13be'
+    FEDOTA_CONTAINER_FQIN: 'registry.fedoraproject.org/fedora:29'
+    PRIOR_FEDORA_CONTAINER_FQIN: 'registry.fedoraproject.org/fedora:28'
 
     ####
     #### Variables for composing new cache-images (used in PR testing) from
@@ -63,6 +65,9 @@ timeout_in: '120m'
 # testing for every platform
 integration_task:
 
+    depends_on:
+        - 'config'
+
     gce_instance:
         # Generate multiple parallel tasks, covering all possible
         # 'matrix' combinations.
@@ -85,8 +90,8 @@ fedora_packaging_task:
     # Runs within Cirrus's "community cluster"
     container:
         matrix:
-            image: "registry.fedoraproject.org/fedora:28"
-            image: "registry.fedoraproject.org/fedora:29"
+            image: "${FEDOTA_CONTAINER_FQIN}"
+            image: "${PRIOR_FEDORA_CONTAINER_FQIN}"
         cpu: 4
         memory: 12
 
@@ -111,7 +116,7 @@ config_task:
                # fedora:28 doesn't have go mod by default
                # and we should only need one check to make sure
                # config changes were synced
-            image: "registry.fedoraproject.org/fedora:29"
+            image: "${FEDOTA_CONTAINER_FQIN}"
         cpu: 4
         memory: 12
 
```

cevich:
Reason I suggest making 'integration' depend on 'config is:  "Fail fast".  But I'm not actually sure if the config test failing would make the integration test results invalid or not.

cevich:
Otherwise, this LGTM :smile:

haircommander:
let's try it out :D (it could make integration tests invalid, but worse we'd spend money on spin cycles that would fail anyway)

cevich:
ya, z'ktly, and clearly there are no actual gating tests here, so it's a start.

cevich:
Neat it works, okay, well there ya go!   Sorry for my confusion...it must have been very confusing :grin:

haircommander:
(testing to verify it stops jobs if config fails)

haircommander:
@saschagrunert I'm tagging you to merge this if it looks good to you :)

Makefile
```diff
@@ -44,6 +48,10 @@ bin/conmon: src/conmon.o src/cmsg.o src/ctr_logging.o src/utils.o | bin
 %.o: %.c
 	$(CC) $(CFLAGS) -o $@ -c $<
 
+config: git-vars cmd/conmon-config/conmon-config.go runner/config/config.go runner/config/config_unix.go runner/config/config_windows.go
```
saschagrunert: I guess we have to put it as dependency to either `all` or `src/conmon.o`, right?

haircommander: I'd rather we not have the hard dependency on go. The tool chain is pretty opinionated. I'd rather keep config.h in sync with config.go (by the gating test) so one can presumably compile conmon with just the c tool chain. WDYT

saschagrunert: Yeah I also thought that this would be a valuable solution. SGTM

PR #46, opened 2019-06-21
conmon: force unlink attach socket

giuseppe:
before we attempt to create the attach socket, make sure it doesn't
already exist otherwise bind(2) will fail with EEXIST.

Closes: https://github.com/containers/libpod/issues/3381

### DIFF
--- a/src/conmon.c
+++ b/src/conmon.c
@@ -845,6 +845,9 @@ static char *setup_attach_socket(void)
 	if (fchmod(attach_socket_fd, 0700))
 		pexit("Failed to change attach socket permissions");
 
+	if (unlink(attach_addr.sun_path) == -1 && errno != ENOENT)
+		pexitf("Failed to remove existing attach socket: %s", attach_addr.sun_path);
+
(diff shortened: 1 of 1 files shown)


haircommander:
Nice, LGTM.
BTW: I see the handling of unlinking the attach socket is a bit different than that of the console socket:
```
if (unlink(csname) < 0)
    pexit("Failed to unlink temporary random path");
```
Is it desirable to handle unlinking them the same?

giuseppe:
thanks for the quick review.  Do you mean taking care of `ENOENT`?  We should not fail if the file doesn't exist.

haircommander:
Ok, I'll look into it. That can be a separate PR though

haircommander:
@giuseppe unfortunately, we still haven't merged c/conmon into cri-o. Can also make a PR there?

giuseppe:
> Ok, I'll look into it. That can be a separate PR though

I mean, we should not fail in this PR, for `unlink(csname)` we still need to fail as the file was created by us

haircommander:
ah, understood, thanks for explaining!

giuseppe:
I've opened a PR for CRI-O here: https://github.com/cri-o/cri-o/pull/2515

PR #47, opened 2019-06-21
Fix linker error due to missing source files

debarshiray:
Otherwise, the build would fail with:
  /usr/bin/ld: conmon/builddir/../src/conmon.c:499: undefined
    reference to `use_syslog'

Signed-off-by: Debarshi Ray <rishi@fedoraproject.org>


giuseppe:
bot, add author to whitelist

LGTM

haircommander:
Thanks @debarshiray LGTM

PR #49, opened 2019-06-24
Conmon exec

haircommander:
Change the conmon -e endpoint to optionally take a --attach option, as well as an attach_sync_fd. This allows us to attach to an exec session with conmon

Also attempt to change CI to build from CRI-O master instead of the latest release on the fedora (needed to actually merge this into libpod in a timely manner)

### DIFF
--- a/contrib/cirrus/lib.sh
+++ b/contrib/cirrus/lib.sh
@@ -192,6 +192,11 @@ install_crio_repo() {
     echo "Cloning current CRI-O Source for faster access later"
     sudo rm -rf "$GOSRC"  # just in case
     ooe.sh git clone $CRIO_REPO $GOSRC
+
+    # Install CRI-O
+    cd crio
+    ooe.sh make PREFIX=/usr
+    ooe.sh sudo make install PREFIX=/usr
 }
 
 install_testing_deps() {
@@ -239,17 +244,6 @@ selinux_permissive(){
     echo "Entering SELinux Permissive mode, will switch to enforcing upon shell exit"
     trap "setenforce 1" EXIT
     setenforce 0
-}
-
-match_crio_tag() {
-    req_env_var "
-        GOSRC $GOSRC
-    "
-    export CRIO_VER="$(rpm -q --qf '%{V}' cri-o)"
-    echo "Checking out CRI-O tag v$CRIO_VER to match installed rpm $(rpm -q cri-o)"
-    cd $GOSRC
-    ooe.sh git remote update
-    git checkout v$CRIO_VER | tail -1
 }
 
 build_and_replace_conmon() {
--- a/src/conmon.c
+++ b/src/conmon.c
@@ -60,10 +60,11 @@ static char *opt_container_pid_file = NULL;
 static char *opt_conmon_pid_file = NULL;
 static gboolean opt_systemd_cgroup = FALSE;
 static gboolean opt_no_pivot = FALSE;
+static gboolean opt_attach = FALSE;
 static char *opt_exec_process_spec = NULL;
 static gboolean opt_exec = FALSE;
 static char *opt_restore_path = NULL;
-static gchar **opt_restore_args = NULL;
+static gchar **opt_runtime_opts = NULL;
 static gchar **opt_runtime_args = NULL;
 static gchar **opt_log_path = NULL;
 static char *opt_exit_dir = NULL;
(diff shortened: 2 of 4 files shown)


contrib/cirrus/lib.sh
```diff
@@ -192,6 +192,11 @@ install_crio_repo() {
     echo "Cloning current CRI-O Source for faster access later"
     sudo rm -rf "$GOSRC"  # just in case
     ooe.sh git clone $CRIO_REPO $GOSRC
+
+	# Install CRI-O
```
rhatdan: Tabbing issues.

haircommander:
Okay this is ready for a final review. I set it so conmon now has an --updated flag that sets the use of the new API. this allows it to be totally backward compatible. That was tested against CRI-O e2e at https://github.com/cri-o/cri-o/pull/2540, as well as the integration tests here. The new API in use is being tested at https://github.com/cri-o/cri-o/pull/2377 . @giuseppe @rhatdan @mrunalp @mheon @saschagrunert PTAL

mheon:
What's the process on merging this one, Peter - Conmon PR first, then CRI-O
and others?

haircommander:
@mheon  CRI-O and conmon can merge in any order (now that it's backward compatible). Podman is only reliant on conmon, though i like keeping the two conmons in sync

src/conmon.c
```diff
@@ -415,8 +424,15 @@ static void check_child_processes(GHashTable *pid_to_handler)
 
 		/* If we got here, pid > 0, so we have a valid pid to check.  */
 		cb = g_hash_table_lookup(pid_to_handler, &pid);
-		if (cb)
+		if (cb) {
 			cb(pid, status, 0);
+		} else if (opt_updated) {
+			ndebugf("couldn't  find cb for pid %d", pid);
```
giuseppe: nit, two whitespaces before "find"

src/conmon.c
```diff
@@ -572,10 +588,7 @@ static gboolean oom_cb_cgroup_v2(int fd, GIOCondition condition, G_GNUC_UNUSED g
 	gboolean ret = G_SOURCE_REMOVE;
 
 	/* Drop the inotify events.  */
-	ssize_t num_read = read(fd, &events, sizeof(events));
-	if (num_read < 0) {
-		pwarn("Failed to read events");
-	}
+	read(fd, &events, sizeof(events));
```
giuseppe: is this intentional?

src/conmon.c
```diff
@@ -811,24 +824,37 @@ static void container_exit_cb(G_GNUC_UNUSED GPid pid, int status, G_GNUC_UNUSED
...
 		return;
 
-	if (opt_exec)
-		res_key = "exit_code";
-	else
-		res_key = "pid";
-
 	if (message) {
 		escaped_message = escape_json_string(message);
 		json = g_strdup_printf("{\"%s\": %d, \"message\": \"%s\"}\n", res_key, res, escaped_message);
```
giuseppe: another way to solve the versioning could be to specify both `data,exit_code` or `data,pid` in the outputted JSON and have:
```
json = g_strdup_printf("{\"data\": \"%s\", \"%s\": %d, \"message\": \"%s\"}\n", res, res_key, res, escaped_message);
```

haircommander: I think this solution is much cleaner, but I am going to stick with using --api-version instead. There are a couple of other cases that --api-version covers (for instance, making sure we only send on the sync pipe once in the legacy api). --attach could be used for this, but I found it hacky and difficult to explain (which means confusing to read).

I don't think the cost of sending and formatting the longer JSON is worth it if we can't get rid of the --api-version flag.

WDYT

src/conmon.c
haircommander: ah this was in the crio version and not here. fixing

src/conmon.c
```diff
@@ -97,6 +102,7 @@ static GOptionEntry opt_entries[] = {
 	{"conmon-pidfile", 'P', 0, G_OPTION_ARG_STRING, &opt_conmon_pid_file, "Conmon daemon PID file", NULL},
 	{"systemd-cgroup", 's', 0, G_OPTION_ARG_NONE, &opt_systemd_cgroup, "Enable systemd cgroup manager", NULL},
 	{"exec", 'e', 0, G_OPTION_ARG_NONE, &opt_exec, "Exec a command in a running container", NULL},
+	{"updated", 0, 0, G_OPTION_ARG_NONE, &opt_updated, "Use the updated conmon API", NULL},
```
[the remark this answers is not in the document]
haircommander: I opted for conmon exec to be version 1 to match the semantic versioning being used in this repo.

src/conmon.c
giuseppe: yes sure, if we cannot take rid of it completely then there is no advantage in changing the JSON format

giuseppe:
thanks, LGTM

PR #55, opened 2019-07-11
release 1.0.0-rc1

haircommander:
in preparation for https://github.com/containers/libpod/pull/3143, bump conmon to v1.0.0-rc1. That way, if we find issues, we can still change it before releasing 1.0


haircommander:
@saschagrunert @rhatdan @mrunalp @mheon @giuseppe PTAL

PR #57, opened 2019-07-18
conmon: check it is valid pid before kill'ing it

giuseppe:
always check it is a valid PID before passing it down to kill(2).

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1730919


haircommander:
LGTM

PR #62, opened 2019-08-02
conmon: close FDs before creating exit file

giuseppe:
fix a race with Podman restart where it tries to bind ports
immediately after the exit file was created.

### DIFF
--- a/src/conmon.c
+++ b/src/conmon.c
@@ -1098,7 +1098,7 @@ static void do_exit_command()
 
 int main(int argc, char *argv[])
 {
-	int fd, ret;
+	int ret;
 	char cwd[PATH_MAX];
 	_cleanup_free_ char *default_pid_file = NULL;
 	_cleanup_free_ char *csname = NULL;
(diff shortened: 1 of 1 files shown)


src/conmon.c
```diff
@@ -1616,23 +1617,35 @@ int main(int argc, char *argv[])
...
-			continue;
-		if (close(fd) < 0 && errno == EBADF)
-			break;
+	fdsdir = opendir("/proc/self/fd");
+	if (fdsdir != NULL) {
+		int fd;
+		int dfd = dirfd(fdsdir);
+		struct dirent *next;
+
+		for (next = readdir(fdsdir); next; next = readdir(fdsdir)) {
```
haircommander: I would prefer to declare next here instead of above. feels more idiomatic

haircommander:
I the /proc/self/fd approach better! besides the nit LGTM

rhatdan:
@giuseppe Make @haircommander change and then we can merge.

src/conmon.c
giuseppe: I am fine with doing the change, but the current version follows more the rest of the `conmon.c` file where we don't use any C99 for loop (in facts we are using only `//` comments from C99)

haircommander: Hm okay i'd be fine with leaving it then, and instead update it all together in a future PR

PR #64, opened 2019-08-09
Bump 2.0

haircommander:
given complications with conmon shipping with CRI-O version, it is preferred to version conmon like this to make clear conmon-1.15.x is older than conmon with exec


haircommander:
:cookie:

PR #65, opened 2019-08-19
conmon: propagate error code from the container

giuseppe:
at exit propagate the return code from the watched container.  This is
useful when running in systemd so it can properly detect a failure and
restart the container if necessary.

### DIFF
--- a/src/conmon.c
+++ b/src/conmon.c
@@ -1072,8 +1072,31 @@ static void setup_oom_handling(int pid)
 
 static void do_exit_command()
 {
+	pid_t exit_pid;
 	gchar **args;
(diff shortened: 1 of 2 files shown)


vrothberg:
@giuseppe, compiler isn't happy:
```
src/conmon.c:1079:31: error: ordered comparison of pointer with integer zero [-Werror=extra]
 1079 |  if (signal(SIGCHLD, SIG_DFL) < 0) {
      |                               ^
```

TomSweeneyRedHat:
LGTM assuming happy tests.

rhatdan:
LGTM

PR #68, opened 2019-08-30
Makefile: don't fail if clean is called without a build

agners:
Do not fail when calling the clean target without building first.

Signed-off-by: Stefan Agner <stefan@agner.ch>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,8 +66,7 @@ bin:
 
 .PHONY: clean
 clean:
-	rm -f bin/conmon src/*.o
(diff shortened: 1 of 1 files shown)


agners:
This is helpful to create a OpenEmbedded recipe which seems to call the clean target unconditionally.

haircommander:
thanks for the PR @agners LGTM

haircommander:
bot, add author to whitelist

rhatdan:
LGTM
Thanks @agners

PR #74, opened 2019-09-17
Add CONN_SOCK_BUF_SIZE to config

haircommander:
instead of having a note that says to sync the value

### DIFF
--- a/runner/config/config.go
+++ b/runner/config/config.go
@@ -3,4 +3,7 @@ package config
 const (
 	// BufSize is the size of buffers passed in to sockets
 	BufSize = 8192
+	// ConnSockBufSize is the size of the socket used for
+	// to attach to the container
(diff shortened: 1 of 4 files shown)


runner/config/config.go
```diff
@@ -3,4 +3,7 @@ package config
 const (
 	// BufSize is the size of buffers passed in to sockets
 	BufSize = 8192
+	// ConnSockBufSize is the size of the socket used for
+	// to attach to the container
+	ConnSockBufSize = 32768
```
rhatdan: Why doesn't this use the constant defined in config.h?  I guess same goes for BufSize.

rhatdan: Oops I guess this is because one is Go and the other is C

rhatdan:
LGTM

PR #76, opened 2019-10-09
conmon: close the sync pipe before exit command

giuseppe:
before calling the exit command be sure the sync pipe is closed so
that the waiting process can stop waiting.

It solves an issue with Podman where the cleanup process could hang
since the container is already locked by the podman process waiting
on the sync pipe.  The reproducer here:

podman --runtime /bin/false run --rm  alpine true

### DIFF
--- a/src/conmon.c
+++ b/src/conmon.c
@@ -42,6 +42,7 @@
 #define CGROUP2_SUPER_MAGIC 0x63677270
 #endif
 
+static int sync_pipe_fd = -1;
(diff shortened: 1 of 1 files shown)


giuseppe:
@mheon it should solve the issue I've seen with Podman

haircommander:
LGTM

mheon:
LGTM!

giuseppe:
LGTM comment addressed

PR #78, opened 2019-10-15
Add 'trace' log level support

saschagrunert:
The trace log level is currently not supported by conmon, but seems to be
a valid level for container runtimes like CRI-O and podman. This results
that the container currently won't start if the log level is set to
'trace'. Beside this, we now case insensitively compare the log level
string to make it more fail safe.


haircommander:
LGTM

PR #82, opened 2019-11-08
 Handle window resize events in a different fifo 

haircommander:
The terminal_ctrl_fd was being overloaded with events. CRI-O wants to reopen the log files at any time, and podman wants to not drop window resize events on exec.

But we can have both.

Introduce winsz_fd_*, a special fifo pair for window resize events. The terminal_ctrl passes window resize events to the write side, and the read side can process them (after masterfd_stdout is initialized, so we have somewhere to write to and we don't drop events).

Also, add ctrl event types to conmon config

### DIFF
--- a/cmd/conmon-config/conmon-config.go
+++ b/cmd/conmon-config/conmon-config.go
@@ -16,10 +16,20 @@ func main() {
 #define STDIO_BUF_SIZE %d
 #define CONN_SOCK_BUF_SIZE %d
 #define DEFAULT_SOCKET_PATH "%s"
+#define WIN_RESIZE_EVENT %d
+#define REOPEN_LOGS_EVENT %d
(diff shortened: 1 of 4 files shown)


haircommander:
replaces https://github.com/containers/conmon/pull/80

haircommander:
@saschagrunert can you PTAL and let me know if this solves the issue #80 solves?

saschagrunert:
I tested it locally with critest and it seems to work, thanks @haircommander!

saschagrunert:
I think we should release a new version afterwards :)

rhatdan:
@giuseppe @vrothberg PTAL

rhatdan:
This looks good, but I want to get one more person to review.

PR #89, opened 2019-11-27
Initialize the endpoint for attach before calling fork.

DaanDeMeyer:
This way, when an error occurs in the attach endpoint logic, conmon can exit immediately without having to clean up the child process. Because the exit happens before forking, the child process doesn't have to be cleaned up at all.

This doesn't solve the core problem that no cleanup is being done on exit but after looking into the code making sure everything is cleanup up correctly is a bit out of my league and is better done by someone more familiar with the code. 

As an aside (and this is absolutely not meant in a bad way), I found the code rather hard to understand with the massive single conmon.c file with very a large main function, lots of globals/callbacks and no clearly defined point of cleanup. I was also expecting a lot more inline documentation on what conmon does and how its implementation works. I find it hard to make any kind of substantial changes to a project without this information at hand.

I didn't test the change yet locally, I'm assuming CI will kick off once I make the PR to verify if this breaks anything.

### DIFF
--- a/src/conmon.c
+++ b/src/conmon.c
@@ -1480,6 +1480,20 @@ int main(int argc, char *argv[])
 	add_argv(runtime_argv, opt_cid, NULL);
 	end_argv(runtime_argv);
 
+	/* Setup endpoint for attach */
+	_cleanup_free_ char *attach_symlink_dir_path = NULL;
+	if (opt_bundle_path != NULL) {
+		attach_symlink_dir_path = setup_attach_socket();
+		dummyfd = setup_terminal_control_fifo();
+		setup_console_fifo();
+
+		if (opt_attach) {
+			ndebug("sending attach message to parent");
+			write_sync_fd(attach_pipe_fd, 0, NULL);
+			ndebug("sent attach message to parent");
+		}
+	}
+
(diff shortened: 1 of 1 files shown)


haircommander:
Thanks for the PR @DaanDeMeyer 
> As an aside (and this is absolutely not meant in a bad way), I found the code rather hard to understand with the massive single conmon.c file with very a large main function, lots of globals/callbacks and no clearly defined point of cleanup. I was also expecting a lot more inline documentation on what conmon does and how its implementation works. I find it hard to make any kind of substantial changes to a project without this information at hand.

I (one of the conmon maintainers) absolutely agree. It's high on my list of wants to refactor conmon to be more readable. I appreciate you digging through the code

Change LGTM at a glance

giuseppe:
LGTM

I've restarted the failing tests

rhatdan:
Thanks @DaanDeMeyer 
Any one wanting to rewrite/refactor conmon would be appreciated.

saschagrunert:
> Thanks @DaanDeMeyer
> Any one wanting to rewrite/refactor conmon would be appreciated.

Sounds like another great opportunity to rewrite conmon in Rust. :innocent:

Issue #84, opened 2019-11-14
Provide logging behavior policies applied by conmon to stdout/stderr


portante:
Today `conmon` is engaging in a byte-capture of `stdout`/`stderr` pipes, recording data as a series of JSON documents in a log file.  All bytes from both file descriptors are captured without discrimination.  We also see that `conmon` tries to write to disk as fast as possible (but sequentially) with what it reads from both pipes.

There is a natural back-pressure presented by `conmon` to the processes in the container writing to `stdout`/`stderr` that results from the rate the disk can accepts `write()`s. It is usually the case that all `conmon` processes server containers write to the same disk.  Which means it is possible for "noisy" containers to take up much of the available disk bandwidth leading to unwanted impacts of one container on another.

Further, the rate at which containers write to disk is independent of the rate at which log file readers (or scrapers; e.g. fluentd, fluent-bit, rsyslog, syslog-ng, filebeat, etc.) can read.  Because log file rotation is performed independent of the readers and `conmon` on most platforms, it is possible for `conmon` to write more data than what a reader can take in before a log file is deleted behind the reader's back with it ever seeing it.

These two problems suggest that log behavior policies applied by `conmon` to the log stream will give administrators the chance to solve them.

There are three policy behaviors we can begin to consider, each applied to `stdout`/`stderr` independently:

1. `back-pressure`
  - Given a rate specified in "bytes per interval", stop reading from the given pipe if the number of bytes read so far exceeds the limit for that interval, continuing to read again at the beginning of the next interval
 - Suggested interval could just be 1 second by default, and/or allow it to be configured along with the # of bytes for that interval 
2. `drop`
 - Given a rate like specified for `back-pressure`, drop bytes read over the limit for the remainder of the interval, accepting them again at the start of the next interval
3. `ignore`
 - Ignore all bytes read from a given pipe without writing them to disk.

This will allow an administrator to apply a policy to all `conmon` processes.  Since `conmon` processes don't "know" about each other, coordination of the setting of that policy would have to be managed externally to `conmon`.

It is possible that `conmon` processes could periodically poll for configuration changes, or accept some sort of signal to re-evaluate the change.

rhatdan:
@giuseppe Could you see about adding these flags.

giuseppe:
I've a few questions.

- Is `ignore` the equivalent of `drop` with a bytes-per-interval=0?

- Are these limits supposed to be changed at runtime?  e.g. can I change bytes-per-interval while the container is running?

[...]

PR #92, opened 2019-12-09
WIP: logging behavior policies and rated-limited logging

syedriko:
A PR to address https://github.com/containers/conmon/issues/84

### DIFF
--- a/src/log_rate.c
+++ b/src/log_rate.c
@@ -59,8 +59,8 @@ bool log_rate_parse_rate_limit(const char* rate_limit_string, size_t* rate_limit
 	}
 	char* endptr;
 	errno = 0;
-	size_t unscaled_rate_limit = strtol(rate_limit_string, &endptr, 10);
-	if (errno != 0) {
+	long unscaled_rate_limit = strtol(rate_limit_string, &endptr, 10);
+	if (errno != 0 || unscaled_rate_limit <= 0) {
 		return false;
 	}
 	size_t scale = 1;
--- a/src/log_rate.c
+++ b/src/log_rate.c
@@ -1,0 +1,110 @@
+#define _GNU_SOURCE
+#include <stdio.h>
+#include <unistd.h>
+#include <string.h>
+#include <time.h>
+#include <stdlib.h>
+#include <errno.h>
+
+#include "utils.h"
+#include "log_rate.h"
+#include "ctr_logging.h"
+
+#define IO_BUF_SIZE      65536
+#define BYTES_PER_PERIOD 1024
+#define SECS_PER_PERIOD  1
+#define BILLION          1000000000
+
+static const struct timespec secs_per_period = {SECS_PER_PERIOD, 0};
+static size_t bytes_written_this_period = 0;
+static struct timespec start_of_this_period;
+
+static struct timespec subtract_timespecs(const struct timespec* first, const struct timespec* second);
+static int64_t subtract_timespecs_nano(const struct timespec* first, const struct timespec* second);
+static void write_io_bufs(stdpipe_t pipe, char* buf, ssize_t count);
+static void sleep_for_the_rest_of_this_period(const struct timespec* start_of_this_period);
+static void start_new_period();
+
+void log_rate_init() {
+	start_new_period();
+}
+
+bool log_rate_write_to_logs(stdpipe_t pipe, char *buf, ssize_t num_read) {
+	char* buf_start = buf;
+	ssize_t bytes_remaining = num_read;
+	struct timespec now;
+	clock_gettime(CLOCK_MONOTONIC, &now);
+	int64_t diff_nano = subtract_timespecs_nano(&now, &start_of_this_period);
+	if (diff_nano < SECS_PER_PERIOD * BILLION) {
+		ssize_t bytes_we_can_write = BYTES_PER_PERIOD - bytes_written_this_period;
+		if (num_read <= bytes_we_can_write) {
+			write_io_bufs(pipe, buf_start, num_read);
+			bytes_written_this_period += num_read;
+			return true;
+		} else {
+			write_io_bufs(pipe, buf_start, bytes_we_can_write);
+			bytes_written_this_period += bytes_we_can_write;
+			buf_start += bytes_we_can_write;
+			bytes_remaining = num_read - bytes_we_can_write;
+			sleep_for_the_rest_of_this_period(&start_of_this_period);
+			start_new_period();
+		}
+	} else {
+		start_new_period();
+	}
+
+	ssize_t chunks = bytes_remaining / BYTES_PER_PERIOD;
+	ssize_t remainder = bytes_remaining % BYTES_PER_PERIOD;
+
+	for (ssize_t i = 0; i < chunks; ++i) {
+		write_io_bufs(pipe, buf_start + i * BYTES_PER_PERIOD, BYTES_PER_PERIOD);
+		sleep_for_the_rest_of_this_period(&start_of_this_period);
+		start_new_period();
+	}
+	if (remainder != 0) {
+	    if (bytes_written_this_period + remainder > BYTES_PER_PERIOD) {
+		    sleep_for_the_rest_of_this_period(&start_of_this_period);
+		    start_new_period();
+	    }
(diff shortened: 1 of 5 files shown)


src/log_rate.c
```diff
@@ -0,0 +1,110 @@
...
+	int ret;
+	struct timespec now, sleep, diff;
+	clock_gettime(CLOCK_MONOTONIC, &now);
+	diff = subtract_timespecs(&now, &start_of_this_period);
+	sleep = subtract_timespecs(&secs_per_period, &diff);
+	if (sleep.tv_sec < 0 || sleep.tv_nsec < 0)
+		nexit("negative sleep");
+	do {
+		ret = nanosleep(&sleep, &sleep);
+	} while (ret == -1 && errno == EINTR);
```
portante: Not sure you want a loop that doesn't check the time change.  Shouldn't we be checking to see if the absolute time has reached or passed the calculated absolute time for the sleep?

src/log_rate.c
```diff
@@ -0,0 +1,110 @@
...
+}
+
+void start_new_period() {
+	bytes_written_this_period = 0;
+	clock_gettime(CLOCK_MONOTONIC, &start_of_this_period);
```
portante: Isn't this going to cause the periods to drift?  Instead, could we just use math to calculate the periods from some initial startup time?

src/log_rate.c
```diff
@@ -0,0 +1,110 @@
...
+bool log_rate_write_to_logs(stdpipe_t pipe, char *buf, ssize_t num_read) {
+	char* buf_start = buf;
+	ssize_t bytes_remaining = num_read;
+	struct timespec now;
+	clock_gettime(CLOCK_MONOTONIC, &now);
+	int64_t diff_nano = subtract_timespecs_nano(&now, &start_of_this_period);
+	if (diff_nano < SECS_PER_PERIOD * BILLION) {
+		ssize_t bytes_we_can_write = BYTES_PER_PERIOD - bytes_written_this_period;
+		if (num_read <= bytes_we_can_write) {
+			write_io_bufs(pipe, buf_start, num_read);
```
portante: Does this assume that after a write we are still in the same period?

It seems that we have to re-evaluate what period we are in following each `write_to_logs()` call.

src/log_rate.c
```diff
@@ -0,0 +1,110 @@
...
+			bytes_written_this_period += num_read;
+			return true;
+		} else {
+			write_io_bufs(pipe, buf_start, bytes_we_can_write);
+			bytes_written_this_period += bytes_we_can_write;
+			buf_start += bytes_we_can_write;
+			bytes_remaining = num_read - bytes_we_can_write;
+			sleep_for_the_rest_of_this_period();
```
giuseppe: I think we shouldn't block the thread execution here.

We should either propagate the condition back (with the remaining time) and set a glibc timer, or we need to use a much shorter period so that it is acceptable, in the worst case, to hang for that duration

syedriko: This code implements the backpressure policy, I'm still adding the others. 
I'm not sure I follow your concern or how backpressure is going to be maintained if we return from here without blocking. If we do, aren't we going to quickly enough be reading from the container's pipe again?

src/log_rate.c
syedriko: My reading of nanosleep(2) is that it sleeps for relative time intervals and when interrupted by a signal, it will store the remaining time to sleep in the timespec pointed to by the second argument, which we will use on the next pass through the loop as the relative time to sleep.

And I think I'm not doing the right thing for the negative sleep - if it's negative, it means we're in a new period and we should just return without sleeping. We start a new period after every sleep anyway.

src/log_rate.c
syedriko: Could you illustrate with a bit of code how you propose to calculate the periods?

src/log_rate.c
```diff
@@ -0,0 +1,210 @@
...
+	return true;
+}
+
+void log_rate_init(log_policy_t policy, size_t rate_limit) {
+	log_policy = policy;
+	bytes_per_period = rate_limit;
+	start_new_period();
+}
+
+bool log_rate_write_to_logs(stdpipe_t pipe, char *buf, ssize_t num_read) {
```
portante: Perhaps it'd be worth a comment here stating that this function's call signature must be kept the same as `write_to_logs()`?

portante: Also, it is not clear this is the interface we want to rate limit.  The SRE needs to be able to control the final behavior of `conmon`, after it has added all the metadata to the logs.  So it would seem we need to add the rate limit to the file writes to disk, or the journald writes.

src/log_rate.c
```diff
@@ -0,0 +1,210 @@
...
+	}
+}
+
+bool log_rate_parse_rate_limit(const char* rate_limit_string, size_t* rate_limit) {
+	if (rate_limit_string == NULL) {
+		rate_limit = 0;
+		return true;
+	}
+	char* endptr;
+	errno = 0;
```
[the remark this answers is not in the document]
syedriko: It's a shortcut, also used in https://github.com/containers/conmon/blob/c8f7443640371822b4fa83c2df94733d650c4daf/src/conmon.c#L247. I'll fix it according to strtol(3).

src/log_rate.c
```diff
@@ -0,0 +1,210 @@
...
+			if (diff_nano < 0) {
+				return true;
+			} else {
+				dropping = false;
+				start_new_period();
+			}
+		}
+		break;
+	case IGNORE:
+		return true;
```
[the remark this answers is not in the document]
syedriko: We can collect the metrics, but what should we do with them?

src/log_rate.c
```diff
@@ -0,0 +1,210 @@
...
+	case 'K':
+		scale = (size_t)1024;
+		break;
+	case 'M':
+		scale = (size_t)1024 * 1024;
+		break;
+	case 'G':
+		scale = (size_t)1024 * 1024 * 1024;
+		break;
+	case 'T':
```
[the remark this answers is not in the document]
syedriko: For the rate limit suffixes, I was going by what pv(1) does. We can call it future-proofing. Or I can get rid of T.

As far as a config file goes, conmon receives all it's config on the command line. I think this is something that needs to be looked at at the podman and cri-o level, which I don't have a good grasp of yet.

src/log_rate.c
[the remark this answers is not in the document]
syedriko: Oh, and with nanosleep() we don't because we only look at errno if nanosleep() returns an error.

src/log_rate.c
syedriko: Yep, that's true. Stay tuned...

src/log_rate.c
alanconway: My first reaction to the sleep was also "yeuch" but given that conmon is written as a sync read()/write() loop there's no other way to idle but to sleep.

If conmon was written as a poll/epoll loop then you could set a timer and flip the read fd out of the loop till it fired. That would be neater and more efficient, but a bigger overhaul of conmon.

src/log_rate.c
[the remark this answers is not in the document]
haircommander: to be returned up the stack they'd have to be collected and written to a file here, and read by the caller. Alternatively a pipe fd could be sent from the parent similar to how sync_fd is in main. Seems like a reasonable extension but a bit out of scope here

src/log_rate.c
alanconway: Cautionary note: clock_gettime() is relatively expensive in tight inner loops. I'm not saying this code is calling it needlessly, but keep in mind as you fiddle with the timing calculations. A common strategy is to cache the time for re-use and only refresh it after blocking IO calls.

syedriko: I think this code is pretty close to that idea. It finds out the time the first thing on entering this code and then the usual sequence is
write()
clock_gettime() // how long to sleep for?
nanosleep() // sleep
clock_gettime() // when does the new period start
Not sure if it's worth/good idea eliminating the last clock_gettime() and just adding the sleep time to the previous timestamp. Seems like a topic for a profiling session once we get to that point.

syedriko:
The work on making log collection more reliable is ongoing, but this branch as it stands is unlikely to be directly usable.

PR #93, opened 2019-12-09
Write exit file to container persistent directory

mrunalp:
Write the exit files under the container bundle directory so
they are available even when we restart the node. Also, keeping
it scoped there makes the cleanup easier.

Also, we hardlink this file to the exit directory location
used for inotify.

### DIFF
--- a/src/conmon.c
+++ b/src/conmon.c
@@ -57,6 +57,7 @@ static char *opt_cuuid = NULL;
 static char *opt_name = NULL;
 static char *opt_runtime_path = NULL;
 static char *opt_bundle_path = NULL;
+static char *opt_persist_path = NULL;
 static char *opt_container_pid_file = NULL;
 static char *opt_conmon_pid_file = NULL;
 static gboolean opt_systemd_cgroup = FALSE;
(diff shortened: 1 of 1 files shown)


haircommander:
LGTM

src/conmon.c
```diff
@@ -1750,13 +1750,25 @@ int main(int argc, char *argv[])
...
+	/*
+	 * Writing to this directory helps if a daemon process wants to monitor all container exits
+	 * using inotify.
+	 */
 	if (opt_exit_dir) {
-		_cleanup_free_ char *status_str = g_strdup_printf("%d", exit_status);
 		_cleanup_free_ char *exit_file_path = g_build_filename(opt_exit_dir, opt_cid, NULL);
-		if (!g_file_set_contents(exit_file_path, status_str, -1, &err))
-			nexitf("Failed to write %s to exit file: %s", status_str, err->message);
+		if (link(ctr_exit_file_path, exit_file_path) == -1) {
```
giuseppe: this works only if both paths are on the same file system.  Can we be sure about it?

It would be safer to fallback to `g_file_set_contents` if the `link(2)` fails

mrunalp: Good point @giuseppe, I will add the fallback on failure.

mrunalp:
Updated. I got rid of the link because in our most common use case, we will have exit dir in `/var/run` and persistent directory under `/var/lib`.

giuseppe:
LGTM

src/conmon.c
```diff
@@ -1750,8 +1752,18 @@ int main(int argc, char *argv[])
 		closedir(fdsdir);
 	}
 
+	_cleanup_free_ char *status_str = g_strdup_printf("%d", exit_status);
+
+	/* Always write the exit file to container directory */
+	_cleanup_free_ char *ctr_exit_file_path = g_build_filename(opt_persist_path, "exit", NULL);
```
haircommander: We need to check if opt_persist_path is defined, as it's default is NULL

mrunalp:
Updated with NULL check.

haircommander:
Is there a reason to mandate the perisitent dir? I'd rather not to break backwards compatibility. We can just check if it's null before writing to it

mheon:
I'm also a bit concerned about this - we're going to break any existing Podman with this. Falling back if the new flag is not specified seems like a backwards-compatible alternative.

mrunalp:
I am fine making this optional. Modified the PR to write only if persist directory path is set.

haircommander:
LGTM

TomSweeneyRedHat:
LGTM, but would like a head nod from @mheon before merging.

mheon:
LGTM

PR #97, opened 2019-12-11
Persist the oom file when --persist-dir is specified

mrunalp:
We want to ensure we save the oom file as well for reboot scenario.

@giuseppe @haircommander PTAL.


haircommander:
LGTM

giuseppe:
LGTM

wking:
you want this for the cgroups v2 handler too, right?

haircommander:
> you want this for the cgroups v2 handler too, right?

good point, opened https://github.com/containers/conmon/pull/102

PR #98, opened 2019-12-12
conmon: check oom counter on cgroup v1

giuseppe:
when running on cgroup v1, also check whether the OOM counter is > 0
before writing the "oom" file.

Previously, conmon was detecting an OOM every time an event fd was
triggered.

Unfortunately that is not enough, as an event on the eventfd is also
triggered when the cgroup is deleted.  That could cause a race
condition where conmon sees the event from the cgroup deletion before
the SIGCHLD+SIGUSR1.  That could happen for example when systemd
CollectMode is set for failed units.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1782601

### DIFF
--- a/src/conmon.c
+++ b/src/conmon.c
@@ -42,6 +42,7 @@
 #define CGROUP2_SUPER_MAGIC 0x63677270
 #endif
 
+static int memory_cgroup_v1_oom_control_fd = -1;
 static int sync_pipe_fd = -1;
 static volatile pid_t container_pid = -1;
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
LGTM but one of the tests isn't happy.

rhatdan:
LGTM

haircommander:
LGTM

mrunalp:
Awesome :+1:

PR #100, opened 2019-12-12
conmon: use splice(2) to copy from stdin

giuseppe:
the first advantage is that it is much faster as data is not copied
from the kernel to the userspace and the back to the kernel, and the
second advantage is that we don't block on writ'ing on the fd all the
data we've buffered.

Closes: https://github.com/containers/conmon/issues/99


rhatdan:
LGTM
@haircommander @mrunalp PTAL

haircommander:
LGTM

giuseppe:
don't merge yet, I might have found an issue

haircommander:
I'll cut 2.0.7 for cri-o then, and we can get this in 2.0.8 :smile:

giuseppe:
fixed  :-)

PR #104, opened 2019-12-13
Revert "conmon: check oom counter on cgroup v1"

haircommander:
This reverts commit 4100fb20bc156fdd602f8684affa27e6b139d0ad.

We found issues where the reading of the memory control file was racing with the cgroup manager's write. while we want to prevent false positives, not having false negatives is more important. Revert this commit until we find a more complete solution.

### DIFF
--- a/src/conmon.c
+++ b/src/conmon.c
@@ -42,7 +42,6 @@
 #define CGROUP2_SUPER_MAGIC 0x63677270
 #endif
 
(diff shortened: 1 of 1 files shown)


haircommander:
@giuseppe @mrunalp @saschagrunert @mheon @rhatdan PTAL

mrunalp:
LGTM. @giuseppe ptal.

giuseppe:
LGTM

PR #106, opened 2019-12-17
journal logging: write to /dev/null instead of -1

haircommander:
We should never flush the buffer because we have control of what we're putting in, but calls to writev() require a valid fd. This isn't a problem in runtime, but static code checks flag it. Fix it by writing to /dev/null

fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1780892

### DIFF
--- a/src/ctr_logging.c
+++ b/src/ctr_logging.c
@@ -167,6 +167,13 @@ bool write_to_logs(stdpipe_t pipe, char *buf, ssize_t num_read)
  */
 int write_journald(int pipe, char *buf, ssize_t buflen)
 {
+	/* When using writev_buffer_append_segment, we should never approach the number of
+	 * entries necessary to flush the buffer. Therefore, the fd passed in is for /dev/null
+	 */
+	_cleanup_close_ int dev_null = open("/dev/null", O_WRONLY | O_CLOEXEC);
+	if (dev_null < 0)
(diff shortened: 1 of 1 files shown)


rhatdan:
@mrunalp @giuseppe @vrothberg PTAL

src/ctr_logging.c
```diff
@@ -167,6 +167,13 @@ bool write_to_logs(stdpipe_t pipe, char *buf, ssize_t num_read)
  */
 int write_journald(int pipe, char *buf, ssize_t buflen)
 {
+	/* When using writev_buffer_append_segment, we should never approach the number of
+	 * entries necessary to flush the buffer. Therefore, the fd passed in is for /dev/null
+	 */
+	_cleanup_close_ int dev_null = open("/dev/null", O_WRONLY | O_CLOEXEC);
```
giuseppe: now we open the the file each time `write_journald` is used.  Could we do it just once?

haircommander: we could, but that introduces the possibility of leaks, as it'd have to be a global (or we'd have to significantly rework the api structure). Do you think those options are worth it?

giuseppe: opening it once and leaking it should be fine I think.  We could leak it in the function without affecting the api.

PR #107, opened 2019-12-20
Fixes format string to limit the size of the string to 10 characters

ret2libc:
%10s just ensures that when printing, at least 10 characters are
written. However, in this case we probably want *at most* 10 characters
to be written.


haircommander:
thanks for the PR @ret2libc LGTM

rhatdan:
LGTM

PR #109, opened 2019-12-23
split conmon.c into different files

haircommander:
conmon as it stands is pretty difficult to work with. over the years, conmon's main function has grown very large, and the conmon.c was about 1500 lines. This discourages contributors and adds maintenance burden.

instead, let's refactor conmon.c into many different files. The general structure either describes conmon lifecycle (cli process, runtime arg setup, container attach functionality, exit functions) or serves as a library (cgroup management, handling conmon oom).

I am marking a wip because I should document each file's use, and I think there are more ways main() can be cut down. mostly putting this out there to get it through CI. Nevertheless, PTAL at the general approach and let me know if you have any better grouping of functions


Makefile
```diff
@@ -78,7 +78,27 @@ src/utils.o: src/utils.c src/utils.h
 
 src/ctr_logging.o: src/ctr_logging.c src/ctr_logging.h src/utils.h
 
-src/conmon.o: src/conmon.c src/cmsg.h src/config.h src/utils.h src/ctr_logging.h
+src/conmon.o: src/conmon.c src/cmsg.h src/config.h src/utils.h src/ctr_logging.h src/globals.h src/cli.h src/oom.h src/cgroup.h src/conn_sock.h src/ctrl.h src/ctr_stdio.h
```
haircommander: I am woefully not knowledgeable about Makefiles. I am sure there's a better way to do this, but I don't know about it.

haircommander:
`cc  -o conmon 'conmon@exe/src_conmon.c.o' 'conmon@exe/src_cmsg.c.o' 'conmon@exe/src_ctr_logging.c.o' 'conmon@exe/src_utils.c.o' -Wl,--no-undefined -Wl,--as-needed -Wl,--start-group /usr/lib/x86_64-linux-gnu/libglib-2.0.so -Wl,--end-group  `
@giuseppe how do I update LGTM.com?

giuseppe:
>  @giuseppe how do I update LGTM.com?

there is not really any configuration there.  I think it detects that conmon is using  meson (there is a meson.build file) and tries to build using it

giuseppe:
I think you just need to update the list of files in `meson.build`

haircommander:
this is ready, more can come later

rhatdan:
Does not pass tests.

haircommander:
> Does not pass tests.

tests are busted because of an accounting error. They should be fixed soon

haircommander:
this is **actually** ready now. PTAL @giuseppe @rhatdan @saschagrunert @mheon @mrunalp

rhatdan:
LGTM

PR #117, opened 2020-01-14
create some unit tests

haircommander:
conmon has historically been tested against the tools that rely on it. This has caused difficulties with breaking changes, as well as some hacky testing procedures. Instead, we should test conmon on its own.

What better way to test conmon than in the way its used most frequently: exec'd from go.

This PR begins a test suite for conmon. This includes:
	create a ginkgo suite
	start creating some runner/conmon code to cleanly tap into conmon from go
	update makefile to have some go specific things
	add a cirrus task to run the tests on PRs


haircommander:
@cevich @giuseppe @rhatdan @mheon @saschagrunert PTAL

.gitignore
```diff
@@ -53,3 +53,5 @@ Mkfile.old
 dkms.conf
 
 bin/
+
+vendor/
```
mheon: You sure? We might need this if more Go code gets into the project

haircommander: Why would we need it, besides saving time on CI runs (and new downloads) for the initial `make vendor` ?

mheon: If we ever grow Go bindings (there was some talk of moving some of the OCI code in here, I recall?) we'll probably be vendoring for that

haircommander: yeah and this also starts that work, but the vendoring can be deterministically created from just `go.mod` and `go.sum`. There doesn't seem like a need to pull the whole vendor directory into it if it can just be generated

haircommander: IMO

mheon: Up to you - certainly keeps the repo cleaner

cevich:
Whoa! Whoa! Whoa! :+1: Can we drop the old broken Openshit 3.7 thing?

/me looks at the actual code....

haircommander:
> Whoa! Whoa! Whoa! +1 Can we drop the old broken Openshit 3.7 thing?
> 
> /me looks at the actual code....

we should probably still test against downstream projects until this has more complete coverage. but hopefully soon :tm:

cevich:
Yay!  The `.cirrus.yml` and `contrib/cirrus/setup_unit_tests.sh` stuff all looks fine. There's a much better version of 'req_env_var' in-use by libpod but it's not a biggie.  Still looking...

runner/conmon_test/conmon_test.go
```diff
@@ -0,0 +1,201 @@
...
+	})
+	Describe("no container UUID", func() {
+		It("should fail", func() {
+			_, err := getConmonOutputGivenOptions(
+				conmon.WithPath(conmonPath),
+				conmon.WithContainerID(ctrID),
+			)
+			Expect(err).To(ContainSubstring("Container UUID not provided. Use --cuuid"))
+		})
+	})
```
cevich: I'm really happy to see unit-tests period, so I'm hesitant about suggesting you avoid (and/or remove) the negative cases.

My main reason against negative unit-tests is they can be deceptively complex (as proving a "negative" often is).  They also tend to be the most burdensome in terms of maintenance of test and subject code.  It's okay IMHO if unit tests are only positive and interface-validating.  We can rely on integration tests to catch the negative cases worth investigating.

cevich:
...okay done.  I think this is a fine and ***MUCH*** needed addition, so everything above is just a suggestion for pondering.  I didn't see anything that can't easily be fixed/improved in future iterations.

Thanks for writing these @haircommander

runner/conmon/options.go
```diff
@@ -0,0 +1,68 @@
...
+		return nil
+	}
+}
+
+func WithRuntimePath(path string) ConmonOption {
+	return func(ci *ConmonInstance) error {
+		ci.args = append(ci.args, "--runtime", path)
+		return nil
+	}
+}
```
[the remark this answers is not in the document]
giuseppe: The advantage though is that each of them can be specialized.  For example the `WithLogDriver` seems very reasonable.

runner/conmon_test/conmon_test.go
```diff
@@ -0,0 +1,201 @@
...
+	. "github.com/onsi/gomega"
+)
+
+var (
+	conmonPath   = "/usr/bin/conmon"
+	ctrID        = "abcdefghijklm"
+	validPath    = "/tmp"
+	validLogPath = "/tmp/log"
+	invalidPath  = "/not/a/path"
+)
```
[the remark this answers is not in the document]
giuseppe: even with multiple tests in parallel, these paths would still be valid no?

TomSweeneyRedHat:
LGTM.  Unless @cevich has strong objections, I'd merge this now and then address stuff further if needed.  Great start @haircommander !

runner/conmon_test/conmon_test.go
haircommander: For the first set of options (the ones I've written tests for so far), conmon fails without them in the inital pid. These tests are all negative because they guarentee conmon requires them. The tests will get more complex from here. I don't mind the simple string matching as evidence of failure for now, and if proving "negative" gets overly burdensome in the future, I'll avoid them :smile:

runner/conmon_test/conmon_test.go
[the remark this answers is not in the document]
haircommander: ah validLogPath is shouldn't be used now, gotta fix that.

invalidPath will always be invalid, and any path that requires a valid path that will be created (--log-path /valid/path) currently is uniquely created per test

runner/conmon/options.go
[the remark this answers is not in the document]
haircommander: agreed, there is a lot of duplication in the operation of "add this to ci.args"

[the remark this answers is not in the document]
haircommander: there will definitely be more With... functions, I plan on having one for each option conmon supports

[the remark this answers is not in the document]
haircommander: added a helper function :+1:

haircommander:
@saschagrunert changed the test name to integration, and also changed the old integration task to cri-o_integration. addressed all other nits :smile:

runner/conmon_test/conmon_test.go
cevich: sounds good.

PR #122, opened 2020-02-25
oom v1: verify cgroup hasn't been cleaned up

haircommander:
This PR is a replacement for https://github.com/containers/conmon/commit/4100fb20bc156fdd602f8684affa27e6b139d0ad, in which we verify an OOM occured when we get triggered on the eventfd.
When we check for an oom, there are three cases we must worry about:
    1. the cgroup OOM'd
    2. the cgroup was destroyed
    3. the cgroup OOM'd and was destroyed
    
We have always handled case 1 (when we get an oom, the eventfd gets an events, we write the oom file)

For case two, we need to check to see if the cgroup.memory_control file exists. We are pretty sure if this file doesn't exist, and this is the first event we've received on the fd, the cgroup was cleaned up instead of an oom happening. This also matches what containerd does.

the third case is when we're not sure about the second case. If both an oom and a destroy event happen, we need to make sure we properly set an oom. We can't use the handling code for the second case here, because the files will be gone, but an oom will have happened. Thus, before we check if the files exist, we need to check if there are still events in the eventfd. If there are, that means both oom and destroy happened, and we should write the oom files

hopefully closes: https://bugzilla.redhat.com/show_bug.cgi?id=1782601 and https://bugzilla.redhat.com/show_bug.cgi?id=1801568

Special thanks to @rh-ulrich-o for finding the issues and essentially telling me what to do here

### DIFF
--- a/src/conmon.c
+++ b/src/conmon.c
@@ -534,52 +534,63 @@ static int write_oom_files()
 	return oom_fd >= 0 ? 0 : -1;
 }
 
-// user_data is expected to be the container's pid, used to verify the cgroup hasn't been cleaned up
+/* user_data is expected to be the container's cgroup.event_control file,
+ * used to verify the cgroup hasn't been cleaned up */
 static gboolean oom_cb_cgroup_v1(int fd, GIOCondition condition, gpointer user_data)
 {
-	int *pid = (int *)user_data;
-	uint64_t oom_event;
+	char *cgroup_event_control_path = (char *)user_data;
+	uint64_t event_count;
 	ssize_t num_read = 0;
-	_cleanup_free_ char *memory_cgroup_path = NULL;
-	_cleanup_free_ char *cgroup_event_control_path = NULL;
 
-	if ((condition & G_IO_IN) != 0) {
-		num_read = read(fd, &oom_event, sizeof(uint64_t));
-		if (num_read < 0) {
-			nwarn("Failed to read oom event from eventfd");
-			return G_SOURCE_CONTINUE;
-		}
-
-		if (num_read > 0) {
-			if (num_read != sizeof(uint64_t))
-				nwarn("Failed to read full oom event from eventfd");
-
-			/* attempt to read the container's cgroup path.
-			 * if we can't, the cgroup has probably been cleaned up.
-			 * In all likelihood, this means we recieved an event on the eventfd
-			 * because the memory.oom_control file was removed, not because of an OOM
-			 */
-			memory_cgroup_path = process_cgroup_subsystem_path(*pid, false, "memory");
-			if (!memory_cgroup_path) {
-				return G_SOURCE_CONTINUE;
-			}
-
-			/* it's unclear how the above condition could pass and this wouldn't, but let's just be safe */
-			cgroup_event_control_path = g_build_filename(memory_cgroup_path, "cgroup.event_control", NULL);
-			if (access(cgroup_event_control_path, F_OK) < 0) {
-				return G_SOURCE_CONTINUE;
-			}
-
-			write_oom_files();
-
-			return G_SOURCE_CONTINUE;
-		}
+	if ((condition & G_IO_IN) == 0) {
+		/* End of input */
+		close(fd);
+		oom_event_fd = -1;
+		free(cgroup_event_control_path);
+		return G_SOURCE_REMOVE;
 	}
 
-	/* End of input */
-	close(fd);
-	oom_event_fd = -1;
-	return G_SOURCE_REMOVE;
+	num_read = read(fd, &event_count, sizeof(uint64_t));
+	if (num_read < 0) {
+		nwarn("Failed to read oom event from eventfd");
+		return G_SOURCE_CONTINUE;
+	}
+
+	if (num_read == 0) {
+		close(fd);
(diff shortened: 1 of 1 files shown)


haircommander:
/me wishes we had a real style guide for conmon. The formatting is all over the place

haircommander:
PTAL @giuseppe

src/conmon.c
```diff
@@ -534,10 +534,14 @@ static int write_oom_files()
 	return oom_fd >= 0 ? 0 : -1;
 }
 
-static gboolean oom_cb_cgroup_v1(int fd, GIOCondition condition, G_GNUC_UNUSED gpointer user_data)
+// user_data is expected to be the container's pid, used to verify the cgroup hasn't been cleaned up
+static gboolean oom_cb_cgroup_v1(int fd, GIOCondition condition, gpointer user_data)
 {
+	int *pid = (int *)user_data;
```
haircommander: should there be a null check here? I considered it but wasn't sure what to do with a null value...

haircommander:
> Could we parse `/sys/fs/cgroup/memory/$CGROUP/memory.oom_control` and see if oom_kill is set? That will work similarly to what we are already doing on cgroup v2.

that's what we did here https://github.com/containers/conmon/commit/4100fb20bc156fdd602f8684affa27e6b139d0ad and it created false positives

giuseppe:
> that's what we did here [4100fb2](https://github.com/containers/conmon/commit/4100fb20bc156fdd602f8684affa27e6b139d0ad) and it created false positives

I am fine with the current version, but I am afraid whatever we try to do in conmon can race with systemd `CollectMode`.  If systemd is faster at reading the counter and cleaning the cgroup, then we miss the event.

I am still not sure what could cause the race in 4100fb20bc156fdd602f8684affa27e6b139d0ad, it seems unlikely it could be in the kernel and reading the `oom_kill` counter from `oom_control`  is similar to what systemd does; and it is still vulnerable to the race with `CollectMode`

haircommander:
PTAL @mrunalp @mheon @saschagrunert @rhatdan @TomSweeneyRedHat

mheon:
LGTM

haircommander:
hold this, I'm going to do some stress tests to see how systemd's CollectMode interacts with it

haircommander:
I ran an OOM test about 1300 times the past two days with CRI-O. Not once was there a false negative. This is ready, PTAL

haircommander:
haha jk another case has been brought up that needs consideration

haircommander:
hm, the time to catch the oom has dramatically raised (real 0:19 -> 5:10 on average)... I don't know if I'll find the cause today

rhatdan:
@haircommander Should we merge this or do you want to continue to attempt to improve it?

haircommander:
I'm still hitting performance issues I want to at least figure out the source of. stay tuned...

haircommander:
Okay, this is ready @rhatdan @giuseppe @mheon

src/conmon.c
```diff
@@ -534,31 +534,69 @@ static int write_oom_files()
...
+		return G_SOURCE_CONTINUE;
+	}
+	/* attempt to read the container's cgroup path.
+	 * if we can't, the cgroup has probably been cleaned up.
+	 * In all likelihood, this means we recieved an event on the eventfd
+	 * because the memory.oom_control file was removed, not because of an OOM
+	 */
+	memory_cgroup_path = process_cgroup_subsystem_path(*pid, false, "memory");
+	/* if event_count == 1, we know the only event triggered was a cgroup removal
+	 * if it was greater than 1, we know the cgroup oomed, and then was cleaned up.
```
giuseppe: can't there be a race if the cgroup is removed between the read of `event_count` and the time `process_cgroup_subsystem_path` fails?

src/conmon.c
```diff
@@ -534,31 +534,69 @@ static int write_oom_files()
...
-				nwarn("Failed to read full oom event from eventfd");
-			write_oom_files();
+	ndebugf("Memory cgroup event count: %ld", (long)event_count);
+	if (event_count == 0) {
+		nwarn("Unexpected event count (zero)");
+		return G_SOURCE_CONTINUE;
+	}
+	/* attempt to read the container's cgroup path.
+	 * if we can't, the cgroup has probably been cleaned up.
+	 * In all likelihood, this means we recieved an event on the eventfd
```
giuseppe: s/recieved/received/

src/conmon.c
haircommander: hm, you're right. I changed it so now we get the cgroup.memory_control path directly, so there's less chance of a race ( i don't know if there's anything we can do to entirely mitigate the race, we have to read the fd and check the file and they can't be atomic)

src/conmon.c
haircommander: fixed, thanks!

src/conmon.c
giuseppe: yeah this is tricky :/ even reading again from the eventfd won't be enough

giuseppe:
LGTM

haircommander:
PTAL @mrunalp @rhatdan @saschagrunert 

I ran the cri-o oom test quite a few times to make sure we're ship shape. it's now back to the performance we had before!

PR #124, opened 2020-02-29
Ignore timeout if container process is already dead.

### LINKED ISSUE #123: Race condition in timeout leads to killing all processes on system
We hit this on conmon 2.0.10.

## Symptoms

We've been experiencing an issue on a new CRI-O based Kubernetes cluster where, occasionally, a node would have _every_ user-land process (except init) killed with SIGKILL. This was most likely to happen during a rolling restart as we deployed new code. Our first thought was some sort of memory exhaustion issue, but this was happening on nodes with lots of free memory, and the kernel logs didn't indicate any OOM-killing activity (or anything else suspicious). The only heavily-contended resource was CPU (rolling restarts use a lot of CPU).

Our next best guess was a bug in some process in the root PID namespace accidentally killing everything. (I noted to a coworker that the behavior was a lot like running `kill -9 -1` as root.) We don't run much on the host besides the CRI and kubelet, and we don't have any pods with `hostPID` set, so we guessed it must be one of those two.

Given that _every_ process gets killed whenever the bug occurs, this made nailing the cause down a little challenging. Eventually we shipped a SystemTap module to watch for SIGKILLs (using detached "flight recorder" mode to avoid a killable user-space component) and were able to catch conmon red-handed:
  
Script:
```
probe signal.send {
  if (sig == 9 || pid_name == "auditd") {
    printf("[%ld] %s(%d) sent %s to %s(%d)\n", gettimeofday_ms(), execname(), pid(), sig_name, pid_name, sig_pid)
  }
}
```

Output:
```
[1582945630705] conmon(13331) sent SIGKILL to kthreadd(2)
[1582945630705] conmon(13331) sent SIGKILL to kworker/0:0H(4)
[1582945630705] conmon(13331) sent SIGKILL to mm_percpu_wq(6)
[1582945630705] conmon(13331) sent SIGKILL to ksoftirqd/0(7)
[1582945630705] conmon(13331) sent SIGKILL to rcu_sched(8)
... 250 omitted ...
[1582945630710] conmon(13331) sent SIGKILL to kubelet(13496)
[1582945630710] conmon(13331) sent SIGKILL to conntrack(13513)
[1582945630710] conmon(13331) sent SIGKILL to iptables-legacy(13514)
[1582945630710] conmon(13331) sent SIGKILL to kube-proxy(13515)
```

In the logs we can see that one conmon process sent (or tried to send) a SIGKILL to every 
process and kernel thread in numerical order in the span of about 5ms.

## Proposed mechanism

I wasn't able to find any code in conmon that loops over PIDs trying to kill everything, so I gathered that this must have been the result of an errant `kill(-1, SIGKILL)`. Per the manpage:

>If `pid` equals -1, then sig is sent to every process for which the calling process has permission to send signals, except for process 1 (`init`)

When I went looking, I was dismayed to discover that `-1` is the value conmon uses for invalid PIDs:

https://github.com/containers/conmon/blob/7a830be343876ac381c965c7429a7fb9b3d7a609/src/conmon.c#L923

So any call to `kill(container_pid, SIGKILL)` has the potential to kill a lot more than it bargained for (0 might be a better choice, so the target is `conmon` itself).

Most callsites of `kill` explicitly guard against a negative pid value. There's two exceptions:

https://github.com/containers/conmon/blob/7a830be343876ac381c965c7429a7fb9b3d7a609/src/conmon.c#L1746-L1754

There's two ways to get `kill(-1, SIGKILL)`: if `process_group` is 1, then the first branch will do it, or if `container_pid` is `-1`, then the second branch will (`getpgid(-1)` returns -1). The former seems pretty unlikely to me (you'd need a pretty weird bug in the container runtime), so I'll focus on the latter. (Also, all pidfiles had reasonable PIDs when I checked on the host post-mortem.)

To trigger that case, you need `timed_out` to be true and `container_pid` to be -1. `timed_out` means `timeout_cb` has to fire, which only happens in the glib event loop. There's only one call to the event loop after the timer is set, so it has to happen here:

https://github.com/containers/conmon/blob/7a830be343876ac381c965c7429a7fb9b3d7a609/src/conmon.c#L1724-L1725

We already established that the container runtime isn't messing with us by writing `-1` to the pidfile, so the only place `container_pid` can get set to -1 is in `container_exit_cb`:

https://github.com/containers/conmon/blob/7a830be343876ac381c965c7429a7fb9b3d7a609/src/conmon.c#L923

That in turn, can only be called from `check_child_processes`, which can be called from either the event loop SIGUSR1 (fake SIGCHLD) handler, or *right* before the event loop is entered: 

https://github.com/containers/conmon/blob/7a830be343876ac381c965c7429a7fb9b3d7a609/src/conmon.c#L1710

`container_exit_cb` and `timeout_cb` both exit the event loop; we know from above that `timeout_cb` had to happen in an event loop call, so that out-of-event-loop call must be the place where `container_pid` gets set to -1 (i.e. the container process already exited before this point).

The dead container process means there should be a pending SIGUSR1 going into the event loop. However, if the timeout is *also* pending going into the event loop, I believe the order of callbacks is undefined. (Reading the source, the most-recently-added event source is checked first, which is the timeout.)

In our case, it looks like the offending call was a periodic Kubernetes exec health check with a 1 second timeout. If that process (which is normally fast) completes before the initial `check_child_process` call, and conmon is CPU starved for long enough that the timeout fires before entering `g_main_loop_run`, then it appears we can get to line 1746 with both `container_pid == -1` and `timed_out` true, at which point conmon kills everything else on the host.

I haven't added enough debug logging to 100% confirm this is what's happening, but after carefully reading the code I'm pretty sure this is a plausible explanation, and is consistent with the state of the hosts.

PR forthcoming!

drakedevel:
As discussed in #123, there's a race condition where it's possible to get to this point with `timed_out` true and the process already dead (and `container_pid` already cleared). That has rather negative consequences (`kill(-1, SIGKILL)` as root).

Since the container exited on its own, it seems like we should just ignore the timeout and treat it like a normal container exit.

Fixes #123

Signed-off-by: Andrew Drake <andrew@frontapp.com>

### DIFF
--- a/src/conmon.c
+++ b/src/conmon.c
@@ -1753,7 +1753,11 @@ int main(int argc, char *argv[])
 	int exit_status = -1;
 	const char *exit_message = NULL;
 
-	if (timed_out) {
+	/*
+	 * If timed_out is TRUE but container_pid is -1, the process must have died before
+	 * the timer elapsed. Ignore the timeout and treat it like a normal container exit.
+	 */
+	if (timed_out && container_pid > 0) {
 		pid_t process_group = getpgid(container_pid);
 
 		if (process_group > 0)


haircommander:
Thanks for the investigation and PR! LGTM

rhatdan:
LGTM

PR #127, opened 2020-03-09
oom: fix potential race between verification steps

haircommander:
There exists an opportunity for a race between checking the eventfd for events, and checking that the cgroup exists. Specifically:

if we read the eventfd before the cgroup is destroyed, but check that the cgroup exists after its destroyed, we will miss a real oom (because the first event triggered was an oom)

To mitigate this, we need to check the existance of the file before we read the event. Since the kernel first removes the files, then triggers an event, we resolve the race here

### DIFF
--- a/src/conmon.c
+++ b/src/conmon.c
@@ -541,6 +541,7 @@ static gboolean oom_cb_cgroup_v1(int fd, GIOCondition condition, gpointer user_d
 	char *cgroup_event_control_path = (char *)user_data;
 	uint64_t event_count;
 	ssize_t num_read = 0;
+	gboolean cgroup_removed = FALSE;
 
 	if ((condition & G_IO_IN) == 0) {
 		/* End of input */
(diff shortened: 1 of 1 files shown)


haircommander:
@giuseppe PTAL

src/conmon.c
```diff
@@ -570,21 +570,30 @@ static gboolean oom_cb_cgroup_v1(int fd, GIOCondition condition, gpointer user_d
...
-		if (event_count == 1)
-			return G_SOURCE_CONTINUE;
+		if (event_count == 1) {
+			/* Before we decide the only event was the cgroup deletion, let's double-check for another
+			 * event in the event_fd. If we find one, the cgroup could have been deleted after we read the
+			 * first time, but before we checked for the event_control file.
+			 */
+			if (read(fd, &event_count, sizeof(uint64_t)) < 0)
+				return G_SOURCE_CONTINUE;
+
```
giuseppe: it looks correct to me, but should we ensure the fd is set to not blocking?

haircommander: whoops, I thought it was but it wasn't. good catch

haircommander:
PR simplified. We can check for the cgroup existing first, then read from the fd. Since a removal happens before the event is triggered, we still avoid the race, and this version cleans the code up a bit.

PTAL @giuseppe @rhatdan @mrunalp @saschagrunert @mheon

src/conmon.c
haircommander: With the new version, we don't need a non-block fd anymore

mheon:
LGTM

rhatdan:
@giuseppe Please review and merge.

PR #128, opened 2020-03-12
stdio: sometimes quit main loop after io is done

haircommander:
in the event of a non-terminal, api v1 exec session, we need to run the main_loop a second time to make sure we get all output from the i/o pipes connected to the container.

However, occasionally, the container process will finish before this main loop is called. We catch it when we check the child processes immediately preceeding the g_main_loop_run, but that gives no way to exit the main loop (as we've already removed container_exit_cb as a source.

The solution here is to exit the main loop if we get to stdio_cb, the first i/o pipe has been closed (either stderr or stdout, doesn't matter the order), and we know the container_status. this allows us to read all the i/o, but still properly exit conmon

fixes: https://github.com/containers/libpod/issues/5339

assuming that https://github.com/containers/libpod/pull/5481 passes

### DIFF
--- a/src/conmon.c
+++ b/src/conmon.c
@@ -1345,6 +1345,8 @@ int main(int argc, char *argv[])
 		close(oom_score_fd);
 	}
 
+	/* ignoring SIGPIPE prevents conmon from being spuriously killed */
+	signal(SIGPIPE, SIG_IGN);
 
 	main_loop = g_main_loop_new(NULL, FALSE);
 
--- a/src/conmon.c
+++ b/src/conmon.c
@@ -489,12 +489,28 @@ static gboolean stdio_cb(int fd, GIOCondition condition, gpointer user_data)
 		return G_SOURCE_REMOVE;
 	}
 
+	/* End of input */
 	if (read_eof || (has_hup && !has_input)) {
-		/* End of input */
-		if (pipe == STDOUT_PIPE)
+		/* There exists a case that the process has already exited
+		 * and we know about it (because we checked our child processes)
+		 * but we needed to run the main_loop to catch all the rest of the output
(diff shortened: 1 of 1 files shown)


baude:
lgtm

haircommander:
hold this, I wanna make sure this doesn't regress and also make totally sure it fixes the problem.

src/conmon.c
```diff
@@ -664,29 +680,17 @@ static gboolean conn_sock_cb(int fd, GIOCondition condition, gpointer user_data)
 	ssize_t num_read = 0;
 
 	if ((condition & G_IO_IN) != 0) {
-		num_read = splice(fd, NULL, masterfd_stdin, NULL, 1 << 20, 0);
-		if (num_read > 0)
+		char buf[CONN_SOCK_BUF_SIZE];
```
haircommander: @giuseppe I saw the call to splice hang long enough to have podman timeout. It's probably an i/o scheduler problem but this fixes it. I tried using NONBLOCK for the splice and that didn't work either. Can you think of anything that would prevent the hang under i/o pressure?

giuseppe: unfortunately this can hang indefinitely.

If the data written to `fd` is not read, then conmon will hang on the `write_all`.

I've used the splice to address: https://github.com/containers/conmon/issues/99.

Would lower the `len` argument to splice fix the issue you've seen?  It is currently set to 1M but I think it is fine if we get down to 4096 bytes

haircommander: okay I'll try this, thanks!

rhatdan:
@giuseppe PTAL

haircommander:
this is ready PTAL @saschagrunert @rhatdan @mheon @mrunalp

PR #130, opened 2020-03-17
Bump v2.0.13

haircommander:
as title says


haircommander:
> LGTM, please keep in mind that we have to upload the static binary to the release page (from the cirrus build artifacts).

ah thanks for the reminder

PR #131, opened 2020-03-18
conmon: drop usage of splice(2)

giuseppe:
splice(2) doesn't seem to handle correctly sockets having a type of SOCK_SEQPACKET.

When using a SOCK_SEQPACKET, each read is limited to the current packet.  The data that doesn't fit in a single read(2) call will be discarded.  This can be a problem when the fd_out doesn't accept all the data, so on a short write we lose the data that could not be written by the splice(2) call.

Replace splice(2) with splitting the read/write loop.

### DIFF
--- a/.dir-locals.el
+++ b/.dir-locals.el
@@ -1,0 +1,3 @@
+((c-mode . ((fill-column . 109)
+            (c-basic-offset . 8)
+            (indent-tabs-mode t))))
--- a/src/conmon.c
+++ b/src/conmon.c
@@ -689,29 +689,15 @@ static gboolean conn_sock_cb(int fd, GIOCondition condition, gpointer user_data)
 	ssize_t num_read = 0;
 
 	if ((condition & G_IO_IN) != 0) {
-		num_read = splice(fd, NULL, masterfd_stdin, NULL, 1 << 20, 0);
-		if (num_read > 0)
+		num_read = read(fd, buf, CONN_SOCK_BUF_SIZE);
+		if (num_read < 0)
 			return G_SOURCE_CONTINUE;
 
-		if (num_read < 0) {
-			if (errno != ESPIPE && errno != EINVAL) {
+		if (num_read > 0 && masterfd_stdin >= 0) {
+			if (write_all(masterfd_stdin, buf, num_read) < 0) {
 				nwarn("Failed to write to container stdin");
-			} else {
-				/* Fallback to read-write.  This may lock if the consumer
-				   doesn't read all the data.  */
-				char buf[CONN_SOCK_BUF_SIZE];
-
-				num_read = read(fd, buf, CONN_SOCK_BUF_SIZE);
-				if (num_read < 0)
-					return G_SOURCE_CONTINUE;
-
-				if (num_read > 0 && masterfd_stdin >= 0) {
-					if (write_all(masterfd_stdin, buf, num_read) < 0) {
-						nwarn("Failed to write to container stdin");
-					}
-					return G_SOURCE_CONTINUE;
-				}
 			}
+			return G_SOURCE_CONTINUE;
 		}
 	}
 
--- a/src/conmon.c
+++ b/src/conmon.c
@@ -284,8 +284,12 @@ static int masterfd_stderr = -1;
 /* Used for attach */
 struct conn_sock_s {
 	int fd;
+	gboolean data_ready;
 	gboolean readable;
 	gboolean writable;
+	size_t remaining;
+	size_t off;
+	char buf[CONN_SOCK_BUF_SIZE];
 };
 GPtrArray *conn_socks = NULL;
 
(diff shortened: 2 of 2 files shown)


.dir-locals.el
```diff
@@ -0,0 +1,3 @@
+((c-mode . ((fill-column . 109)
```
haircommander: what's this?

src/conmon.c
```diff
@@ -728,6 +701,91 @@ static gboolean conn_sock_cb(int fd, GIOCondition condition, gpointer user_data)
...
 }
 
+static gboolean conn_sock_cb(G_GNUC_UNUSED int fd, GIOCondition condition, gpointer user_data);
+
+static void sock_try_write_to_masterfd_stdin(struct conn_sock_s *sock)
+{
+	if (sock->remaining) {
+		ssize_t w;
+
+		w = write(masterfd_stdin, sock->buf + sock->off, sock->remaining);
```
haircommander: I would just compress this (I am looking to move away from declaring variables before defining them, c99 style)

src/conmon.c
```diff
@@ -728,6 +701,91 @@ static gboolean conn_sock_cb(int fd, GIOCondition condition, gpointer user_data)
 	return G_SOURCE_REMOVE;
 }
 
+static gboolean conn_sock_cb(G_GNUC_UNUSED int fd, GIOCondition condition, gpointer user_data);
+
+static void sock_try_write_to_masterfd_stdin(struct conn_sock_s *sock)
+{
+	if (sock->remaining) {
```
haircommander: can you do if !sock->remainiing { return;} to not have so many nested blocks?

src/conmon.c
```diff
@@ -728,6 +701,91 @@ static gboolean conn_sock_cb(int fd, GIOCondition condition, gpointer user_data)
...
+	if (num_read > 0) {
+		sock->remaining = num_read;
+		sock->off = 0;
+
+		sock_try_write_to_masterfd_stdin(sock);
+
+		/* Not everything was written to stdin, let's wait the fd is ready.  */
+		if (sock->remaining)
+			g_unix_fd_add(masterfd_stdin, G_IO_OUT, masterfd_write_cb, NULL);
+	} else {
```
haircommander: you could do `if (num_read == 0) {return ...}` above the `num_read > 0` block, which could then be indented in

haircommander:
good find!

.dir-locals.el
giuseppe: to set correctly the formatting mode we use in Emacs :-)

haircommander: just use vim instead :wink:

giuseppe:
fixed the comments and pushed a new version

src/conmon.c
```diff
@@ -728,6 +701,90 @@ static gboolean conn_sock_cb(int fd, GIOCondition condition, gpointer user_data)
...
+
+static void sock_try_write_to_masterfd_stdin(struct conn_sock_s *sock)
+{
+	if (!sock->remaining)
+		return;
+
+	ssize_t w = write(masterfd_stdin, sock->buf + sock->off, sock->remaining);
+	if (w < 0) {
+		nwarn("Failed to write to container stdin");
+	} else {
```
haircommander: stylistically, this is inconsistent with line 727. I don't mind which, but we should probably choose one for the project :smile:

src/conmon.c
```diff
@@ -728,6 +701,90 @@ static gboolean conn_sock_cb(int fd, GIOCondition condition, gpointer user_data)
...
+
+static void write_to_masterfd_stdin(gpointer data, gpointer user_data)
+{
+	struct conn_sock_s *sock = (struct conn_sock_s *)data;
+	bool *has_data = user_data;
+
+	sock_try_write_to_masterfd_stdin(sock);
+
+	if (sock->remaining)
+		*has_data = TRUE;
```
haircommander: why `TRUE` here when it's cast as a bool? either work, but you use `false` below

src/conmon.c
```diff
@@ -728,6 +701,90 @@ static gboolean conn_sock_cb(int fd, GIOCondition condition, gpointer user_data)
...
+	return G_SOURCE_REMOVE;
+}
+
+static gboolean read_conn_sock(struct conn_sock_s *sock)
+{
+	ssize_t num_read;
+
+	/* There is still data in the buffer.  */
+	if (sock->remaining) {
+		sock->data_ready = true;
```
haircommander: same here. Above has_data is capital, but here is lower case. I vote we choose one and stick to it

haircommander:
sorry, a couple more nits, then LGTM

haircommander:
uh @cevich 
```
Failed to start: Error getting access token for service account: 400 Bad Request {"error":"invalid_grant","error_description":"Invalid grant: account not found"}
Failed to stop: Error getting access token for service account: 400 Bad Request {"error":"invalid_grant","error_description":"Invalid grant: account not found"}
``` 
PTAL

src/conmon.c
giuseppe: thanks for noticing.

 We should use true\false for `bool` but TRUE/FALSE for `gboolean`.

mheon:
Are we thinking this is the cause of our exec bug? If so, it's actually amazing that it only happens with exec - this definitely looks like a case that run would hit as well. Regardless, great catch @giuseppe

haircommander:
@giuseppe is this still a draft?

giuseppe:
> Are we thinking this is the cause of our exec bug? If so, it's actually amazing that it only happens with exec - this definitely looks like a case that run would hit as well. Regardless, great catch @giuseppe

I could easily reproduce the issue you've reported locally and this patch solves it for me.  Do we use a different buffer size in Podman for `run`?  The 32kb buffer size used for exec is relevant to the reproducibility of the issue.

giuseppe:
> @giuseppe is this still a draft?

marked ready for review

haircommander:
LGTM

haircommander:
you may need to repush to trigger the tests

TomSweeneyRedHat:
Other than @haircommander's nits and unhappy tests, LGTM

cevich:
FYI: Cirrus-CI testing in GCP should un-break shortly (hours).

giuseppe:
tests are green

rhatdan:
Need to cut a release?

PR #135, opened 2020-03-27
WIP: add --output-file

haircommander:
a major drawback of conmon's double forking model is the parent loses stderr after the fork. This makes debugging challenging.

Add --output-file, allowing the caller to specify a file to write conmon's logs to. Eventually, all container engines should use this option instead of the default `stderr`

### DIFF
--- a/src/utils.c
+++ b/src/utils.c
@@ -5,16 +5,25 @@
 log_level_t log_level = WARN_LEVEL;
 char *log_cid = NULL;
 gboolean use_syslog = FALSE;
+FILE *output_file = NULL;
 
 /* Set the log level for this call. log level defaults to warning.
    parse the string value of level_name to the appropriate log_level_t enum value
 */
-void set_conmon_logs(char *level_name, char *cid_, gboolean syslog_, char *tag)
+void set_conmon_logs(char *level_name, char *cid_, gboolean syslog_, char *tag, char *opt_output_file_, FILE **conmon_output_file_)
 {
 	if (tag == NULL)
 		log_cid = cid_;
 	else
 		log_cid = g_strdup_printf("%s: %s", cid_, tag);
+
+	output_file = stderr;
+	if (opt_output_file_ != NULL) {
+		*conmon_output_file_ = fopen(opt_output_file_, "a");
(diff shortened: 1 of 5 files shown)


src/utils.h
```diff
@@ -33,49 +33,50 @@ typedef enum {
 extern log_level_t log_level;
 extern char *log_cid;
 extern gboolean use_syslog;
+extern FILE *output_file;
 
 #define _pexit(s) \
 	do { \
-		fprintf(stderr, "[conmon:e]: %s %s\n", s, strerror(errno)); \
+		fprintf(output_file, "[conmon:e]: %s %s\n", s, strerror(errno)); \
```
saschagrunert: I think we should log a timestamp here as well. I could follow up on that, WDYT?

src/utils.c
```diff
@@ -5,16 +5,25 @@
...
+void set_conmon_logs(char *level_name, char *cid_, gboolean syslog_, char *tag, char *opt_output_file_, FILE **conmon_output_file_)
 {
 	if (tag == NULL)
 		log_cid = cid_;
 	else
 		log_cid = g_strdup_printf("%s: %s", cid_, tag);
+
+	output_file = stderr;
+	if (opt_output_file_ != NULL) {
+		*conmon_output_file_ = fopen(opt_output_file_, "w");
```
saschagrunert: What about:
```suggestion
		*conmon_output_file_ = fopen(opt_output_file_, "a");
```

src/utils.h
haircommander: agreed, though I think that should be optional behind a flag to not add the overhead if a user doesn't want it

src/utils.c
haircommander: ah good idea, done

mheon:
Really looking forward to having this in Podman!

rhatdan:
LGTM

haircommander:
hold this, need to make sure it's actually printing past the fork lol

TomSweeneyRedHat:
LGTM and happy green test buttons

haircommander:
uh this definitely doesn't work, don't use this option yet

PR #139, opened 2020-03-31
conmon: store status while waiting for pid

giuseppe:
store each termination status reported to us while waiting for the container PID file to be created.

Once we know the container PID we check in the cache whether we already know the status.

### DIFF
--- a/src/conmon.c
+++ b/src/conmon.c
@@ -281,7 +281,11 @@ int main(int argc, char *argv[])
 	 * Glib does not support SIGCHLD so use SIGUSR1 with the same semantic.  We will
 	 * catch SIGCHLD and raise(SIGUSR1) in the signal handler.
 	 */
-	g_unix_signal_add(SIGUSR1, on_sigusr1_cb, pid_to_handler);
+	struct pid_check_data data = {
+		.pid_to_handler = pid_to_handler,
+		.exit_status_cache = NULL,
+	};
+	g_unix_signal_add(SIGUSR1, on_sigusr1_cb, &data);
 
 	if (signal(SIGCHLD, on_sigchld) == SIG_ERR)
 		pexit("Failed to set handler for SIGCHLD");
@@ -302,9 +306,12 @@ int main(int argc, char *argv[])
 	if (csname != NULL) {
 		g_unix_fd_add(console_socket_fd, G_IO_IN, terminal_accept_cb, csname);
 		/* Process any SIGCHLD we may have missed before the signal handler was in place.  */
-		check_child_processes(pid_to_handler);
-		if (!opt_exec || !opt_terminal || container_status < 0)
+		if (!opt_exec || !opt_terminal || container_status < 0) {
+			GHashTable *exit_status_cache = g_hash_table_new_full(g_int_hash, g_int_equal, g_free, g_free);
+			data.exit_status_cache = exit_status_cache;
+			g_idle_add(check_child_processes_cb, &data);
 			g_main_loop_run(main_loop);
+		}
 	} else {
 		int ret;
 		/* Wait for our create child to exit with the return code. */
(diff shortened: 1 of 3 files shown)


haircommander:
LGTM

haircommander:
hm hold this, it seems to have a regression with podman exec

```
Summarizing 2 Failures:

[Fail] Podman exec [It] podman exec pseudo-terminal sanity check 
/home/pehunt/go/src/github.com/containers/libpod/test/utils/utils.go:319

[Fail] Podman exec [It] podman exec with user only in container 
/home/pehunt/go/src/github.com/containers/libpod/test/e2e/exec_test.go:155
```

giuseppe:
> hm hold this, it seems to have a regression with podman exec
> 
> ```
> Summarizing 2 Failures:
> 
> [Fail] Podman exec [It] podman exec pseudo-terminal sanity check 
> /home/pehunt/go/src/github.com/containers/libpod/test/utils/utils.go:319
> 
> [Fail] Podman exec [It] podman exec with user only in container 
> /home/pehunt/go/src/github.com/containers/libpod/test/e2e/exec_test.go:155
> ```

thanks, that helped me to find another issue.  I've pushed a completely different patch

src/conmon.c
```diff
@@ -280,7 +280,11 @@ int main(int argc, char *argv[])
 	 * Glib does not support SIGCHLD so use SIGUSR1 with the same semantic.  We will
 	 * catch SIGCHLD and raise(SIGUSR1) in the signal handler.
 	 */
-	g_unix_signal_add(SIGUSR1, on_sigusr1_cb, pid_to_handler);
+	struct pid_check_data data = {
```
haircommander: can we use a more informative variable name?

src/conmon.c
```diff
@@ -301,9 +305,12 @@ int main(int argc, char *argv[])
 	if (csname != NULL) {
 		g_unix_fd_add(console_socket_fd, G_IO_IN, terminal_accept_cb, csname);
 		/* Process any SIGCHLD we may have missed before the signal handler was in place.  */
-		check_child_processes(pid_to_handler);
-		if (!opt_exec || !opt_terminal || container_status < 0)
+		if (!opt_exec || !opt_terminal || container_status < 0) {
+			GHashTable *cache = g_hash_table_new_full(g_int_hash, g_int_equal, g_free, g_free);
+			data.cache = cache;
+			g_idle_add(check_child_processes_cb, &data);
```
haircommander: can't we unconditionally add this? it looks like we add it the same way before both loops, we just update the cache in this case. However, since we're dealing with the address of data, it shouldn't matter we update cache here

haircommander: unconditionally as in adding it after defining data, not here as well as line 419

src/conmon.c
```diff
@@ -280,7 +280,11 @@ int main(int argc, char *argv[])
 	 * Glib does not support SIGCHLD so use SIGUSR1 with the same semantic.  We will
 	 * catch SIGCHLD and raise(SIGUSR1) in the signal handler.
 	 */
-	g_unix_signal_add(SIGUSR1, on_sigusr1_cb, pid_to_handler);
+	struct pid_check_data data = {
+		.pid_to_handler = pid_to_handler,
+		.cache = NULL,
```
mrunalp: nit: `exit_status_cache` ?

mrunalp:
LGTM

src/conmon.c
giuseppe: the `check_child_processes_cb` callback runs only once (it returns `G_SOURCE_REMOVE`), so we need to make sure it is registered for both loops.

src/conmon.c
giuseppe: I'll use `exit_status_cache` as @mrunalp suggested

PR #140, opened 2020-03-31
some minor cleanups

giuseppe:
two tweaks I've noticed while working on a diffent issue.

### DIFF
--- a/src/ctr_logging.c
+++ b/src/ctr_logging.c
@@ -54,6 +54,11 @@ static char *container_name = NULL;
 static char *container_tag = NULL;
 static size_t container_tag_len;
 
+typedef struct {
+	int iovcnt;
+	struct iovec iov[WRITEV_BUFFER_N_IOV];
+} writev_buffer_t;
+
 static void parse_log_path(char *log_config);
 static const char *stdpipe_name(stdpipe_t pipe);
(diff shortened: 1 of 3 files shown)


src/ctr_logging.c
```diff
@@ -54,6 +54,11 @@ static char *container_name = NULL;
 static char *container_tag = NULL;
 static size_t container_tag_len;
 
+typedef struct {
+	int iovcnt;
+	struct iovec iov[WRITEV_BUFFER_N_IOV];
+} writev_buffer_t;
+
```
TomSweeneyRedHat: Just curious about this move?  I know writev_buffer_t is only used currently here in this c file and probably will only ever be used here.  But did you see a compilation or other error with it in the .h file?  The move is fine, just wondering about the motivation.

TomSweeneyRedHat:
LGTM, I do have one question to satisfy my curiosity.

TomSweeneyRedHat:
and all happy green test buttons.

src/ctr_logging.c
giuseppe: no no compilation warning.  It is just to make explicit it is used only in this file and prevent in the future to have functions using write_buffer_t in other files

TomSweeneyRedHat: Thanks for the 411 @giuseppe.

PR #148, opened 2020-05-07
integration: test with runtime

haircommander:
conmon has had a history of not being well tested. the best we had previously done is test conmon with downstream users: podman and cri-o. This creates an awkward cyclical dependency change if conmon ever has to make a breaking change (which shouldn't happen, but could). Further, podman and cri-o use conmon pretty specifically, and this cases their test cases to miss things.

Instead, we should test conmon's behavior when directly called.

This PR adds an additional integration test to the conmon_tests, which actually sets it up to call runc and checks that it did. It's still very minimal, but provides a starting point of further testing


TomSweeneyRedHat:
Awesome addtion!

TomSweeneyRedHat:
@haircommander the go.sum change was much large than I anticipated.  Do you think it's OK?
@edsantiago if you've a spare cycle, I'd appreciate an eyeball on this.
Good start @haircommander !

runner/conmon_test/ctr_logs_test.go
```diff
@@ -0,0 +1,129 @@
...
+			conmon.WithLogDriver("k8s-file", tmpLogPath),
+			conmon.WithLogDriver("journald", ""),
+		)
+		Expect(stderr).To(BeEmpty())
+
+		_, err := os.Stat(tmpLogPath)
+		Expect(err).To(BeNil())
+	})
+	It("multiple log drivers with one invalid should fail", func() {
+		invalidLogDriver := "invalid"
```
edsantiago, line 118: I basically made it here on Friday and had to give up. I've had all weekend to try to find a safe way to express my opinion, and am not sure I've reached that, but I need to try.

First: this is not an indictment on you, it's a rant on Go. Go seems to make it difficult to write clean code in some cases. My purpose in venting here is to ask for help from the more seasoned Go experts here.

OK. This whole file offends me down to the deepest of my core. Again, not your fault, just... it's horrible. It's unreadable and unmaintainable. The syntactic sugar -- many many lines to express what could/should be said in a few characters -- is depressing. The duplication, even more so. I had to go into my editor and find ways to confirm that, with the exception of `WithLogDriver`, every test in this file is identical. It is impossible for human eyes to confirm that! Couldn't there be a way to table-drive it? Something like defining a helper `_getConmonOutputGivenLogDriver(.....) which only takes `LogDriver` options and refactors all the other identical stuff?

Also, if you're making the effort (thank you) to refactor `invalidLogDriver`, it might be nice to use it in lines 99 and 125.

Again, this is not directed at you - it's a gripe about the system, the Go infrastructure that makes it so easy to be guided into writing unmaintainable code. I've given up on the podman test cases, but am commenting here because I think this is still small enough that it can be cured. Can we try?

haircommander: thanks for your well thought out analysis and suggestions. I appreciate the review. I have updated the file as suggested :smile:

edsantiago: Thank you. That part is much better, and I appreciate that. The need for complicated `WithThisOrThat()` is still depressing, but I don't know enough Go to offer fixes. We may have to live with that. Anyhow, other than that, LGTM and nice work.

haircommander: fyi, WithThisOrThat() is to prevent every caller from needing to specify each configuration value. Right now, each of them are needed. One day, I would like every value that needs configuring to be put into a constructor, and every optional parameter be added with a With() function, but this test suite is still incomplete enough for me to have figured out which I'd deem necessary

rhatdan:
@haircommander Needs a rebase and is pretty old. What do you want to do with this one?

haircommander:
let's see how it does!

haircommander:
this is ready PTAL @rhatdan @saschagrunert @giuseppe

giuseppe (approved):
LGTM

rhatdan:
LGTM

PR #149, opened 2020-05-09
Add Security Policy

TomSweeneyRedHat:
As the title says.

Signed-off-by: TomSweeneyRedHat <tsweeney@redhat.com>


TomSweeneyRedHat:
@rhatdan PTAL

dougsland:
Looks good to me. I am just wondering if there is any kind of generic security email for containers. I see the doc pointed to podman@lists.podman.io.

TomSweeneyRedHat:
@dougsland we'd talked about that when putting together the full policy file in containers/common.  The thinking was there would probably be more people in the podman.io list than a security email list that we might set up.  The nice thing is if we decide to change that in the future, we only have one place to do it at rather than all 30+ container/* projects.

PR #150, opened 2020-05-12
tty: flush pending data when fd is ready

giuseppe:
if there was any data already received, make sure it is flushed to the tty as soon as it is opened.

Closes: https://github.com/containers/libpod/issues/4397

### DIFF
--- a/src/ctrl.c
+++ b/src/ctrl.c
@@ -5,6 +5,7 @@
 #include "globals.h"
 #include "config.h"
 #include "ctr_logging.h"
+#include "conn_sock.h"
 #include "cmsg.h"
 #include "cli.h" // opt_bundle_path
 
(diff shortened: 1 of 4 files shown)


src/ctrl.c
```diff
@@ -59,6 +60,8 @@ gboolean terminal_accept_cb(int fd, G_GNUC_UNUSED GIOCondition condition, G_GNUC
 	masterfd_stdin = console.fd;
 	masterfd_stdout = console.fd;
 
+	schedule_master_stdin_write();
```
haircommander, line 63: could you add a comment motivating this change? just so future developers know why we need to wait till here to add the fd

haircommander:
integration tests are not necessarily the fault of this PR. I will look into fixing them today to get this in

src/ctrl.c
giuseppe: sure, added a comment

mheon:
+1 from me - confirmed that this does resolve my issues!

haircommander:
if https://github.com/cri-o/cri-o/pull/3753 passes I'll merge this and cut 2.0.16

PR #152, opened 2020-05-12
WIP: Cirrus: Migrate to VM images shared with libpod

cevich:
This brings conmon in-line with the setup deployed for all other
container projects, and updates to F31/32 VMs for testing.

Signed-off-by: Chris Evich <cevich@redhat.com>


.cirrus.yml
```diff
@@ -162,80 +121,89 @@ fmt_task:
...
-        image_name: "${IMAGE_BUILDER_CACHE_IMAGE_NAME}"
-        cpu: 4
-        memory: "4Gb"
-        disk: 200
-        # Additional permissions for building GCE images, within a GCE VM
-        scopes:
-            - 'compute'
-            - 'devstorage.full_control'
+    binaries_artifacts:
+        path: "bin/conmon"
```
TomSweeneyRedHat, line 136: Do we need this conmon stuff?

TomSweeneyRedHat: Ooops!  Never mind.  Too many reviews, I thought I was looking at "common", not "conmon"....  <FacePalm>

TomSweeneyRedHat:
@edsantiago if you have a moment to eyeball this, that would be great.

TomSweeneyRedHat:
@cevich static binary test is failing and not leaving many bread crumbs

.cirrus.yml
```diff
@@ -5,135 +5,96 @@ env:
...
+        CIRRUS_CLONE_DEPTH: 1  # source not used
 
-    setup_environment_script: '$SCRIPT_BASE/setup_environment.sh'
-    integration_test_script: '$SCRIPT_BASE/cri-o_test.sh'
+    # Source not used
+    clone_script: mkdir -p '${CIRRUS_WORKING_DIR}'
 
-# this task runs the conmon integration tests
-integration_task:
+    script: '/usr/local/bin/entrypoint.sh |& ${_TIMESTAMP}'
```
edsantiago, line 89: This cirrus.yml does not have `_TIMESTAMP`

edsantiago:
This is the error I see on the static job:
```
Failed to start: Required 'compute.images.useReadOnly' permission for 'projects/libpod-218412/global/images/fedora-29-conmon-75ea13be'
```
I will assume that @cevich can easily understand, diagnose, and fix that.

.cirrus.yml
cevich: not strictly speaking, no.  It simply archives a copy of the binary, in case it's needed for some hind-sight debugging.  I can remove it it's getting in the way.

.cirrus.yml
cevich: oh thanks!

cevich:
> I will assume that @cevich can easily understand, diagnose, and fix that.

Ya, images aren't ready yet.  That's happening in https://github.com/containers/libpod/pull/6193 sorry I forgot to mention that.

.cirrus.yml
cevich: it's gone.

cevich:
I'm still expecting this to fail, just making sure `.cirrus.yml` is mostly right.

note to me: Update `contrib/cirrus/setup_environment.sh` to install downloaded/cached packages in `$PACKAGE_DOWNLOAD_DIR`

cevich:
@haircommander okay, I think this is about as close to "mostly working" as I'm able to get.  

The next thing to solve (maybe easy) is a file conflict.  Take a look into possibly calling the function `build_and_replace_conmon()` (from `lib.sh`), before the RPM command in `setup_environment.sh`.  That will yank out the conflicting file, and since the tests are going to be building conmon from source anyway, should be safe to do.

```
file /usr/libexec/crio/conmon from install of cri-o-2:1.15.4-1.module_f32+8822+4e56faf0.x86_64 conflicts with file from package conmon-2:2.0.15-1.fc32.x86_64
```

I would only go after a packaging-fix as a last resort (implies rebuilding VM images) as it could be a very slippery-slope.  Maybe reach out to @edsantiago for help fixing `setup_environment.sh` and/or it's call to `build_and_replace_conmon()`.  I believe he will recognize this function, as he's seen an analogous operation we do in libpod and buildah CI (also in `lib.sh`).

If all else-fails, IMHO repurposing the cri-o tests in this repo., is a bad design anyway (it was a "short term" solution).  I would seriously consider killing that task off, and relying on the dedicated integration tests you wrote.  Let CRI-O test conmon the CRI-O-way in it's environment.  Just as podman does podman in it's environment.  Anyway, that's MHO.

cevich:
@haircommander I made a bunch of progress here (after merging in the changes from your PR).  I've got to the point where the ginkgo tests _want_ to build, but I think it [fails because it's not installed](https://cirrus-ci.com/task/4683188192673792?command=integration_test#L9-L5).  Are you aware of how we did this before and/or how I may have broken it (installing ginkgo/gomega)?

haircommander:
you'll want to do a `make vendor` to download the dependencies

cevich:
Oooohh Thanks!  That's much easier than what I had in mind, so glad I asked :smile:

cevich:
hrmmm....something seems broken (somewhere) WRT `$GOPATH`.  Ginkgo/gomega and lots of other modules are installed, but 'make test' isn't seeing them somehow.

@haircommander are you sure using `go test...` is the right way, b/c in libpod the Makefile calls ginkgo directly (IIRC)?  

I tried doing it with and without `GO11MODULE=on`, and it fails differently both ways.

cevich:
force-pushed: trying to use ginkgo directly, like how libpod does...

haircommander:
oops sorry, just seeing this. there's no particular reason to use `go test` over `ginkgo`

cevich:
okay, and it seems there is actually a problem with the system not thinking ginkgo is installed.  Maybe my `$PATH` doesn't include `$GO/bin`.  Of course all this runs fine locally :confounded:

haircommander:
@cevich I think the attached patch should fix the vendor problems
[apply.txt](https://github.com/containers/conmon/files/4804823/apply.txt)

cevich:
> @cevich I think the attached patch should fix the vendor problems

Oh thanks...my bad for not noticing your comment.  Okay, applied and `--force` pushed...

cevich:
sigh...no love for CI today :face_with_head_bandage:

cevich:
@haircommander help please :smile:

cevich:
Rebased and force pushed with updated F32 `go.sum`.

cevich:
@haircommander I'm working on this more directly using `hack/get_ci_vm.sh` and more and more coming over to your side...it's very much FUBAR.  There's so much technical-debt here, it's about at the point where I'm feeling a restart is in order.  I'm willing to stick it out here a bit longer though, since it seems I am slowly making progress.

haircommander:
a restart would be fine with me if this setup is worth abandoning

cevich:
Where I'm seeing the most trouble ATM is with trying to graft the CRI-O tests in.  It's blowing up on me with additional go-mod / vendor related errors I have 0% chance of debugging :grinning: But in general, yes, I'm thinking a fresh start might be faster than applying band-aids on top of band-aids.

PR #155, opened 2020-05-20
Conmon image update

haircommander:
carries: https://github.com/containers/conmon/pull/152/files
plus a patch to force the installation of cri-o to address https://github.com/containers/conmon/pull/152#issuecomment-629477014


rhatdan:
LGTM

contrib/cirrus/setup_environment.sh
```diff
@@ -54,31 +55,44 @@ then
...
+			# to build_and_replace_conmon below
+            ooe.sh rpm --force -ivh $PACKAGE_DOWNLOAD_DIR/*.rpm
+
             install_testing_deps
             build_and_replace_conmon
+            build_and_replace_bats
 
             cd "$CRIO_SRC"  # cri-o source
             echo "Building binaries required for testing"
+			ooe.sh make vendor
```
haircommander, line 86: nit, whitespace

rhatdan:
@haircommander Still working on this one?

haircommander:
> @haircommander Still working on this one?

it is on the list yes. I may need more guidance from @cevich

cevich:
Okay, I'm back now.  @haircommander in which direction would you like me to point my finger or eyeballs?

haircommander:
welcome back! Back to the world of fires for you :smiley: 

It seems the bats suite isn't actually running. do you know how to get access to these vms to debug? I haven't figured out why they're not running

cevich:
In the `integration` output I see:
```
...
install  -D -m 755 bin/conmon /usr/bin/conmon
...
```

The Makefile in this repo is missing the definition.  In libpod, it's done like this:

```
SELINUXOPT ?= $(shell test -x /usr/sbin/selinuxenabled && selinuxenabled && echo -Z)
```

The error however, is coming from the `integration_test.sh` call `GOCACHE=/tmp/go-build make vendor`.  Is this make target known to be functional otherwise?

haircommander:
> In the `integration` output I see:
> 
> ```
> ...
> install  -D -m 755 bin/conmon /usr/bin/conmon
> ...
> ```
> 
> The Makefile in this repo is missing the definition. In libpod, it's done like this:
> 
> ```
> SELINUXOPT ?= $(shell test -x /usr/sbin/selinuxenabled && selinuxenabled && echo -Z)
> ```


the selinux enabled bool seems to be a build flag for podman itself. conmon does not use it, so it shouldn't be necessary

> The error however, is coming from the `integration_test.sh` call `GOCACHE=/tmp/go-build make vendor`. Is this make target known to be functional otherwise?

what's the point of using GOCACHE here? the target is functional otherwise

cevich:
> the selinux enabled bool seems to be a build flag for podman itself. conmon does not use it, so it shouldn't be necessary

Naw, it's an `install` (command) flag.  It makes sure the correct context is set on the file after it's put in place.

> what's the point of using GOCACHE here? the target is functional otherwise

I was questioning that as well.  I think it was added as a workaround for some obscure go tooling bug.  I'm also wondering about the wisdom of even using `make vendor` at this level.  I'm going to try splitting that out into a separate task...

cevich:
I've rebased my PR from this one, and am continuing to poke at the `integration` task failures...

rhatdan:
@cevich Any progress?

rhatdan:
This is so old, I am closing. I have updated the cirrus image in my man page PR.

PR #157, opened 2020-05-24
[WIP] Add nix derivation for static builds

### LINKED ISSUE #156: Static Build with nix Package Manager
From CRI-O toolchain point of view following packages already coming with "nix Package Manager" (https://nixos.org/nix/) for static build support:
- crun (https://github.com/containers/crun/pull/366)
- Podman (https://github.com/containers/libpod/pull/5566)
- CRI-O (https://github.com/cri-o/cri-o/pull/2270)

As long as conmon already coming with static build support in Podman style (https://github.com/containers/conmon/blob/master/Makefile#L57-L65), we should also able to simplify it with nix-based solution, in order to get rid of the "chicken or the egg" issue.

This could also unify the static build process with nix-based, at least from crun >> conmon >> crio; afterward we may also promote it for skopeo >> buildah >> podman, which is also partly ready.

Shall we also go for a similar solution?

hswong3i:
Switching from the current static build to a nix based derivation.

Fixes #156

### DIFF
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -1,0 +1,43 @@
+{ system ? builtins.currentSystem }:
+let
+  pkgs = (import ./nixpkgs.nix {
+    overlays = [ (final: prev: {
+      pcre = prev.pcre.overrideAttrs (x: {
+        configureFlags = x.configureFlags ++ [ "--enable-static" ];
+      });
+      p11-kit = prev.p11-kit.overrideAttrs(x: { doCheck = false; });
+      e2fsprogs = prev.e2fsprogs.overrideAttrs(x: { doCheck = false; });
+      gnutls = prev.gnutls.overrideAttrs(x: { doCheck = false; });
+    })];
+    config = {
+      packageOverrides = pkg: {
+        systemd = pkg.systemd.overrideAttrs(x: {
+          mesonFlags = x.mesonFlags ++ [ "-Dstatic-libsystemd=true" ];
+          postFixup = ''
+            ${x.postFixup}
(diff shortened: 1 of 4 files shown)


nix/default.nix
```diff
@@ -0,0 +1,43 @@
+{ system ? builtins.currentSystem }:
+let
+  pkgs = (import ./nixpkgs.nix {
+    overlays = [ (final: prev: {
+      pcre = prev.pcre.overrideAttrs (x: {
+        configureFlags = x.configureFlags ++ [ "--enable-static" ];
+      });
```
saschagrunert, line 7: To clarify if someone is wondering why we use an `overlay` vs a `config` (down below) here: `pcre` is part of `stdenv`, which does not allow us to change it without re-building all dependencies. The major reason why we need to do that is the presence of `glib2` in conmon, which relies on `pcre`.

I vote for removing `glib2` form conmon since it seems to bloat-up the whole binary. I'm happy to contribute that change if nobody has strong opinions against it.

hswong3i:
The pinned rev not working, I am now trying again with https://github.com/NixOS/nixpkgs/releases/tag/20.03

hswong3i:
@saschagrunert i tried for multiple pins for nix but still not able to compile it locally; moreover I tried to learn more about nix but couldn't pick it up quickly.

Shall I close this PR so don't duplicate your contribution effort?

saschagrunert:
Are you running nix as root? This should usually not necessary and may be the reason why the libuv test-suite fails.

hswong3i:
> Are you running nix as root? This should usually not necessary and may be the reason why the libuv test-suite fails.

Ohhh... Yes running nix as  root... That's why I should be and DevOps but a Devs :-S

PR #158, opened 2020-05-25
Add option to delay execution of exit command

mheon:
This is inspired by a need to periodically reap exec sessions within Podman. When launched from the HTTP API, exec sessions must persist for some amount of time after they exit (Docker runs a reaper every 5 minutes, marking inactive sessions and then removing sessions that are already marked on a subsequent run).

Rather than doing this in Podman, which does not have a daemon, it seems logical to leverage Conmon. We're already using Conmon's exit command infrastructure to clean up detached exec sessions, so it's not much of a stretch to use it for this. The change itself is quite simple (new argument to set a delay, and then a sleep before the exit command is invoked).


rhatdan:
LGTM
@haircommander @giuseppe PTAL

haircommander:
LGTM

PR #161, opened 2020-05-27
Add nix derivation for static builds

### LINKED ISSUE #156: Static Build with nix Package Manager

hswong3i:
Switching from the current static build to a nix based derivation.

Similar PR will go for crun/conmon/libpod/etc, too.

Also see
- ~~https://github.com/containers/crun/pull/372~~
- https://github.com/containers/conmon/pull/161
- https://github.com/containers/libpod/pull/6402

Fixes #156

P.S. conmon-v2.0.17 with nix-based binary could be found from https://github.com/alvistack/conmon/releases/tag/v2.0.17

### DIFF
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -1,0 +1,48 @@
+{ system ? builtins.currentSystem }:
+let
+  pkgs = (import ./nixpkgs.nix {
+    overlays = [ (final: prev: {
+      pcre = prev.pcre.overrideAttrs (x: {
+        configureFlags = x.configureFlags ++ [ "--enable-static" ];
+      });
+    })];
+    config = {
+      packageOverrides = pkg: {
+        glib = pkg.glib.overrideAttrs(x: {
+          outputs = [ "bin" "out" "dev" ];
+          mesonFlags = [
+            "-Ddefault_library=static"
+            "-Ddevbindir=${placeholder ''dev''}/bin"
+            "-Dgtk_doc=false"
+            "-Dnls=disabled"
+          ];
+        });
+        systemd = pkg.systemd.overrideAttrs(x: {
+          mesonFlags = x.mesonFlags ++ [ "-Dstatic-libsystemd=true" ];
+          postFixup = ''
+            ${x.postFixup}
+            sed -ri "s;$out/(.*);$nukedRef/\1;g" $lib/lib/libsystemd.a
+          '';
+        });
+      };
+    };
+  });
+
+  self = with pkgs; {
+    conmon-static = (conmon.overrideAttrs(x: {
+      name = "conmon-static";
+      src = ./..;
+      doCheck = false;
+      buildInputs = [
+        glib
+        glibc
+        glibc.static
+        pcre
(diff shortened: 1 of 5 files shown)


nix/default.nix
```diff
@@ -0,0 +1,48 @@
...
+      name = "conmon-static";
+      src = ./..;
+      doCheck = false;
+      buildInputs = [
+        glib
+        glibc
+        glibc.static
+        pcre
+        systemd
+      ];
```
saschagrunert, line 42:
```suggestion
      buildInputs = x.buildInputs ++ [
        glibc.static
        pcre
      ];
```

nix/default.nix
```diff
@@ -0,0 +1,48 @@
...
+        });
+      };
+    };
+  });
+
+  self = with pkgs; {
+    conmon-static = conmon.overrideAttrs(x: {
+      name = "conmon-static";
+      src = ./..;
+      doCheck = false;
```
saschagrunert, line 35: I think we can remove that:
```suggestion
```

nix/default.nix
hswong3i: I try to override the original `buildInputs` but not extend it, because I found that may be not precise enough for nix derivation + static build.

nix/default.nix
hswong3i: Seem this is not useful? https://nixos.org/nixpkgs/manual/#ssec-check-phase

saschagrunert: We do not execute checks on conmon build, so the flag seems not necessary to me, right?

nix/default.nix
saschagrunert: Sounds good :+1:

nix/default.nix
hswong3i: removed ;-)

hswong3i: oic why did I add it... in case of crun `doCheck = false;` is required...

For simple stupid and sync, I just copy-and-paste `doCheck = false;` for all PR ;-)

haircommander:
nix is a mystery to me, but I trust you both and static_binary job passing :smiley: 

LGTM

hswong3i:
@haircommander nix also looks mystery for me so I give up once; without @saschagrunert helps and verify this won’t get happened ;-)

PR #166, opened 2020-05-29
WIP: conn_sock: close stdout side when we don't close stdin

haircommander:
the main consumer of leave-stdin-open, cri-o, needs stdin to stay open, but also wants to consume all data from stdout.
It does not know when the output from stdout has ended unless conmon closes its side of the pipe.

the problem is, if we don't close masterfd_stdin, we also don't close masterfd_stdout, as they point to the same side pipe, which results in cri-o sitting and waiting for input forever.

instead, we need to close the stdout side of the pipe if we don't close masterfd_stdin

### DIFF
--- a/src/conn_sock.c
+++ b/src/conn_sock.c
@@ -192,7 +192,14 @@ static gboolean terminate_conn_sock(struct conn_sock_s *sock)
 			close(masterfd_stdin);
 			masterfd_stdin = -1;
 		} else {
-			ninfo("Not closing input");
+			/*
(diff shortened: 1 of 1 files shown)


haircommander:
waiting on https://github.com/cri-o/cri-o/pull/3787 to verify no breaking changes here (the test passed but I didn't test any others)

mrunalp:
https://github.com/cri-o/cri-o/pull/3812 fixes the tests. We don't need this.

PR #168, opened 2020-05-30
[master] Use `stdenv.mkDerivation` instead of `conmon.overrideAttrs`

hswong3i:
Using `conmon.overrideAttrs` may result as broken due to nixpkgs upstream changes; replace with native `stdenv.mkDerivation` with complete definition could simply avoid this happening.

Changes based on https://github.com/NixOS/nixpkgs/blob/e469b77/pkgs/applications/virtualization/conmon/default.nix

### DIFF
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -28,21 +28,17 @@ let
     };
   });
 
-  self = with pkgs; {
-    conmon-static = (conmon.overrideAttrs(x: {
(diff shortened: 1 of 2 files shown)


hswong3i:
@rhatdan  @saschagrunert sorry I rename the branch and so reopen PR as #169 with fixes ;-)

PR #169, opened 2020-05-31
[nix] Use `stdenv.mkDerivation` instead of `conmon.overrideAttrs`

hswong3i:
Using `conmon.overrideAttrs` may result as broken due to nixpkgs upstream changes; replace with native `stdenv.mkDerivation` with complete definition could simply avoid this happening.

Changes based on https://github.com/NixOS/nixpkgs/blob/e469b77/pkgs/applications/virtualization/conmon/default.nix; also fix `conmon --version` with correct version and commit, e.g.

```
conmon version 2.0.17
commit: 6be2eb8bb24fc403183ad5186a381ae61151b372

```

Fixes #168

### DIFF
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -28,21 +28,18 @@ let
     };
   });
 
-  self = with pkgs; {
-    conmon-static = (conmon.overrideAttrs(x: {
-      name = "conmon-static";
-      src = ./..;
-      doCheck = false;
-      buildInputs = [
-        glib
-        glibc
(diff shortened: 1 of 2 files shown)


hswong3i:
> LGTM

@saschagrunert core logic should be good enough, but need some input for making CI/CD happy, hope you may share some hints for me ;-)

haircommander:
CI/CD failures are known. The test this really cares about is `static` which has passed

PR #175, opened 2020-06-13
Update log-path parsing and unit-tests to handle a lot more edge cases.

goochjj:
When using strtok, you cannot tell if the original string had a : or not, because with

with "journald" you get driver = "journald" and path = NULL
with "journald:" you get driver = "journald" and path = NULL

That means invalid: writes to a file invalid, instead of throwing an error.
Similarly, when passing none or none: we make sure no file is written to disk.
Add debugging output for logdriver diagnostics and use in unit tests.
And "" and : just segfault.

All fixed now.

### DIFF
--- a/runner/conmon/options.go
+++ b/runner/conmon/options.go
@@ -52,6 +52,12 @@ func WithRuntimePath(path string) ConmonOption {
 	}
 }
 
+func WithLogLevel(path string) ConmonOption {
+	return func(ci *ConmonInstance) error {
+		return ci.addArgs("--log-level", path)
+	}
+}
+
 func WithLogDriver(driver, path string) ConmonOption {
 	return func(ci *ConmonInstance) error {
 		fullDriver := path
@@ -59,6 +65,12 @@ func WithLogDriver(driver, path string) ConmonOption {
 			fullDriver = fmt.Sprintf("%s:%s", driver, path)
 		}
 		return ci.addArgs("--log-path", fullDriver)
+	}
+}
+
+func WithLogPath(path string) ConmonOption {
+	return func(ci *ConmonInstance) error {
+		return ci.addArgs("--log-path", path)
 	}
 }
 
--- a/runner/conmon_test/conmon_test.go
+++ b/runner/conmon_test/conmon_test.go
@@ -83,14 +83,19 @@ var _ = Describe("conmon", func() {
 	Describe("ctr logs", func() {
 		var tmpDir string
 		var tmpLogPath string
+		var origCwd string
 		BeforeEach(func() {
 			d, err := ioutil.TempDir(os.TempDir(), "conmon-")
 			Expect(err).To(BeNil())
 			tmpDir = d
 			tmpLogPath = filepath.Join(tmpDir, "log")
+			origCwd, err = os.Getwd()
+			Expect(err).To(BeNil())
 		})
 		AfterEach(func() {
 			Expect(os.RemoveAll(tmpDir)).To(BeNil())
+			err := os.Chdir(origCwd)
+			Expect(err).To(BeNil())
 		})
 		It("no log driver should fail", func() {
 			_, stderr := getConmonOutputGivenOptions(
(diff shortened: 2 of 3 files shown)


runner/conmon/options.go
```diff
@@ -62,6 +68,13 @@ func WithLogDriver(driver, path string) ConmonOption {
 	}
 }
 
+func WithLogDriverRaw(path string) ConmonOption {
+	return func(ci *ConmonInstance) error {
+		fullDriver := path
+		return ci.addArgs("--log-path", fullDriver)
```
haircommander, line 74: Why not just use `path` here? Or name the variable `fullDriver` above?

goochjj: Cuz I copied it and removed the if.  I could consolidate.

Honestly I'm not sure why the withlogdriver function needs driver and path - I'd think the unit tests would be closer to real life without having this intermediary deciding when we need a : and where it should go.  Unless the runner/conmon/ is used for something other than just unit testing.

haircommander: One day, I'd like it to replace the conmon-calling code in cri-o and podman, and give a nice go interface instead of the ad-hoc os/exec code that exists now.

changing it so that we do less validation on `runner/conmon` side is fine with me. I am not married to the implementation as is

goochjj: My main concern was I looked at the implementation and saw 
```
		if driver != "" {
			fullDriver = fmt.Sprintf("%s:%s", driver, path)
		}
```

So there's currently no way to test for ":filename".. Which, fine, it should default to k8s-file in that case.  So from a usage perspective, withLogDriver makes sense as driver,path...

goochjj: See changes below - that probably makes more sense short and long run.

runner/conmon_test/conmon_test.go
```diff
@@ -101,58 +101,170 @@ var _ = Describe("conmon", func() {
...
 				conmon.WithPath(conmonPath),
 				conmon.WithContainerID(ctrID),
 				conmon.WithContainerUUID(ctrID),
 				conmon.WithRuntimePath(validPath),
 				conmon.WithLogDriver("", tmpLogPath),
+				conmon.WithLogLevel("debug"),
 			)
-			Expect(stderr).To(BeEmpty())
+			Expect(stderr).To(ContainSubstring("LogDriver K8S TRUE k8s_file "+tmpLogPath))
+			Expect(stderr).To(ContainSubstring("LogDriver JOURNALD FALSE"))
```
haircommander, line 154: I'm not super convinced I like checking against this. I would rather check against what's actually in the file/journal like we do [here](https://github.com/containers/conmon/pull/148)

runner/conmon_test/conmon_test.go
```diff
@@ -101,58 +101,170 @@ var _ = Describe("conmon", func() {
...
 			)
-			Expect(stderr).To(BeEmpty())
+			Expect(stderr).To(ContainSubstring("LogDriver K8S FALSE k8s_file (null)"))
+			Expect(stderr).To(ContainSubstring("LogDriver JOURNALD FALSE"))
+
+			_, err := os.Stat("none")
+			Expect(err).NotTo(BeNil())
 		})
 		It("log driver as off should pass", func() {
+			os.Remove("off")
```
haircommander, line 207: this should be in a file in a tmpdir for idempotency

haircommander: also can we put "off" in a variable so we can change it in one place :)

haircommander: hmmm nevermind I see what's happening here....

haircommander: what happens if we just don't remove this file? I'm fine hitting a flake if this file happens to exist

haircommander:
sorry, just got to a more complete review

haircommander:
also you've got fmt problems :smiley:

runner/conmon_test/conmon_test.go
haircommander: if you want, we can just drop checks to this now, and later implement a more complete check that the journal/file is actually being written to

goochjj: My intention here was that the debug logging indicates that it will actually do what I requested it to do.  (Hence, given the arguments I provided and the expected behavior, conmon's configured state matches that)

I did it on multiple log lines so additional log drivers could be added without breaking the check.

Checking that the file or journal is actually written to - well, I'm not against that, I just only ever saw empty files from the unit test and wasn't sure how to get conmon to log something :-D i.e. Mocking a "runc" that would instead echo something specific.

runner/conmon_test/conmon_test.go
goochjj: Well, now that the test passes, that's fine.  It'll just flake.  Before when the test was failing (i.e. conmon rc17), --log-path off writes to a file called off, and I'd have to manually rm the file as I was developing the code to fix that.

Since conmon shouldn't actually write those files anymore, I can get behind removing the rm

runner/conmon_test/conmon_test.go
goochjj: Dun

runner/conmon_test/conmon_test.go
goochjj: Removed the os.Remove, and chdir to the tmpdir so any incidentally created files are contained (and the folder is automatically cleaned before/after each test)

runner/conmon_test/conmon_test.go
```diff
@@ -83,14 +83,19 @@ var _ = Describe("conmon", func() {
 	Describe("ctr logs", func() {
 		var tmpDir string
 		var tmpLogPath string
+		var origCwd string
 		BeforeEach(func() {
 			d, err := ioutil.TempDir(os.TempDir(), "conmon-")
 			Expect(err).To(BeNil())
 			tmpDir = d
 			tmpLogPath = filepath.Join(tmpDir, "log")
+			origCwd, err = os.Getwd()
```
haircommander, line 92: I don't see us ever Chdir to the tmpDir, am I missing something?

haircommander: ah, only some tests do, nvm

haircommander:
LGTM, can you squash your commits to a couple? (one would work too)

goochjj:
AYW

squashed and signed off

haircommander:
LGTM

@giuseppe @mheon @rhatdan PTAL

PR #178, opened 2020-06-15
conmon: drop fsync calls

giuseppe:
on a crash, CRI-O completely wipes out the entire storage so there is
no need on syncing the log files.

### DIFF
--- a/src/conmon.c
+++ b/src/conmon.c
@@ -426,8 +426,6 @@ int main(int argc, char *argv[])
 	if (!timed_out)
 		drain_stdio();
 
-	sync_logs();
-
 	int exit_status = -1;
 	const char *exit_message = NULL;
 
--- a/src/ctr_logging.c
+++ b/src/ctr_logging.c
@@ -489,11 +489,6 @@ static void reopen_k8s_file(void)
 
 	_cleanup_free_ char *k8s_log_path_tmp = g_strdup_printf("%s.tmp", k8s_log_path);
 
-	/* Sync the logs to disk */
-	if (fsync(k8s_log_fd) < 0) {
-		pwarn("Failed to sync log file on reopen");
-	}
-
 	/* Close the current k8s_log_fd */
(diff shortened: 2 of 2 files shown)


rhatdan:
What about non CRI-O use cases?

giuseppe:
> What about non CRI-O use cases?

I think the fsync was added to conmon trying to address a different issue with the Kubelet, that is now solved in Kubernetes: https://github.com/kubernetes/kubernetes/pull/73041

I think these fsync are superfluous also for other non CRI-O use cases.  The fsync is performed only at container shutdown, so we would still lose logs if the system crashes while the container is running.  The only case where it makes a difference is when the container is stopped and the system crashes immediately after.

giuseppe:
let's wait for @mrunalp and @haircommander before merging

haircommander:
LGTM waiting for @mrunalp +1

mrunalp:
One case to think about is the kubelet accessing logs of previously exited container. By default one previous instances is kept around for logs.

haircommander:
checking for regressions here https://github.com/cri-o/cri-o/pull/3876

giuseppe:
> One case to think about is the kubelet accessing logs of previously exited container. By default one previous instances is kept around for logs.

does that happen always without a reboot between the container exited and thge kubelet reading the logs?

rhatdan:
@mrunalp Please answer @giuseppe question?

giuseppe:
the only time the fsync could make a difference is on a crash and the file wasn't written to disk before it happened.  AFAIK, with CRI-O we drop completely the storage on such cases so it should not make any difference.  The fsync doesn't make any other difference while the system is running, i.e. another process sees immediately all the changes performed to a file without any need for the fsync call.

mrunalp:
I think we can get this in and we will have enough soak time before 4.6 to revert if needed.

mrunalp:
LGTM

PR #181, opened 2020-06-22
Reap PIDs before running exit command

mheon:
This was first noticed when running Podman when using the exit command delay functionality. When the OCI runtime exits with an error, there are some cases where Conmon will exit without reaping it; thus, we were ending up with a zombie that persisted until the exit command completed and exited. We didn't notice this before because the exit command only took a second or two so the zombie's lifetime was very short, since it would reparent and be reaped once Conmon died. With exit-delay, though, we could be leaving a zombie around for hundreds of seconds, and this means we cannot kill the container during this period.

### DIFF
--- a/src/ctr_exit.c
+++ b/src/ctr_exit.c
@@ -142,6 +142,11 @@ void do_exit_command()
 		_pexit("Failed to reset signal for SIGCHLD");
 	}
 
+	/* We need to reap any zombies (from an OCI runtime that errored) before
+	   running it. */
+	while (waitpid(-1, NULL, WNOHANG) > 0)
+		;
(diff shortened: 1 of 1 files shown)


mheon:
@giuseppe @haircommander PTAL. I made the reaping unconditional, for now, because it shouldn't be that expensive and I don't think it's dangerous.

haircommander:
ah good catch LGTM (though fmt isn't happy)

rhatdan:
LGTM

vrothberg:
Just want to highlight how much I appreciate @mheon's commit messages :+1:

PR #182, opened 2020-06-24
Refactor I/O and add SD_NOTIFY proxy support

goochjj:
Refactored all the conn_sock functionality to be more generic. It can deal
with different types of sockets, stream vs dgram, and reuses all the same
callbacks, shutdown and async functionality.

Conmon creates a notify socket which podman bind-mounts into the container,
and passes in via the spec's environment variables.  Conmon relays the
READY=1 signal.  This is similar to what runc and crun do, but doing it in
conmon and NOT passing NOTIFY_SOCKET to the OCI runtime allows us to start
up properly without runc and crun blocking on the "start" command.

It would also be trivial to add more proxied sockets, i.e. the /dev/log
proof of concept I did would now be super easy, if we wanted to revisit that.

Signed-off-by: Joseph Gooch <mrwizard@dok.org>

### DIFF
--- a/src/conmon.c
+++ b/src/conmon.c
@@ -89,6 +89,11 @@ int main(int argc, char *argv[])
 	/* before we fork, ensure our children will be reaped */
 	atexit(reap_children);
 
+	/* If we were passed a sd-notify socket to use, set it up now */
+	if (opt_sdnotify_socket) {
+		setup_notify_socket(opt_sdnotify_socket);
+	}
+
 	/* Environment variables */
 	sync_pipe_fd = get_pipe_fd_from_env("_OCI_SYNCPIPE");
 
--- a/src/conn_sock.c
+++ b/src/conn_sock.c
@@ -13,14 +13,64 @@
 #include <sys/un.h>
 #include <sys/stat.h>
 
-static gboolean attach_cb(int fd, G_GNUC_UNUSED GIOCondition condition, G_GNUC_UNUSED gpointer user_data);
-static gboolean conn_sock_cb(int fd, GIOCondition condition, gpointer user_data);
-static gboolean read_conn_sock(struct conn_sock_s *sock);
-static gboolean terminate_conn_sock(struct conn_sock_s *sock);
-void conn_sock_shutdown(struct conn_sock_s *sock, int how);
-static void sock_try_write_to_mainfd_stdin(struct conn_sock_s *sock);
-static gboolean mainfd_write_cb(G_GNUC_UNUSED int fd, G_GNUC_UNUSED GIOCondition condition, G_GNUC_UNUSED gpointer user_data);
+static gboolean attach_cb(int fd, G_GNUC_UNUSED GIOCondition condition, gpointer user_data);
+static gboolean remote_sock_cb(int fd, GIOCondition condition, gpointer user_data);
+static void init_remote_sock(struct remote_sock_s *sock, struct remote_sock_s *src);
+static gboolean read_remote_sock(struct remote_sock_s *sock);
+static gboolean terminate_remote_sock(struct remote_sock_s *sock);
+static void remote_sock_shutdown(struct remote_sock_s *sock, int how);
+static void schedule_local_sock_write(struct local_sock_s *local_sock);
+static void sock_try_write_to_local_sock(struct remote_sock_s *sock);
+static gboolean local_sock_write_cb(G_GNUC_UNUSED int fd, G_GNUC_UNUSED GIOCondition condition, G_GNUC_UNUSED gpointer user_data);
+static char *bind_unix_socket(char *socket_relative_name, int sock_type, mode_t perms, struct remote_sock_s *remote_sock,
+			      gboolean do_chdir);
+/*
+  Since our socket handling is abstract now, handling is based on sock_type, so we can pass around a structure
+  that contains everything we need to handle I/O.  Callbacks used to handle IO, for example, and whether this
+  can be read from or written to or both, and the buffers used for the communication.
+*/
 
+/*
+  This defines the Container STDIN, attaches it to the correct FD and sets the flags for handling I/O.
+  setup_attach_socket() is responsible for setting the correct remote FD and pushing it onto the queue.
+*/
+static struct local_sock_s local_mainfd_stdin = {&mainfd_stdin, true, NULL, "container stdin", NULL};
+struct remote_sock_s remote_attach_sock = {
+	SOCK_TYPE_CONSOLE,   /* sock_type */
+	-1,		     /* fd */
+	&local_mainfd_stdin, /* dest */
+	true,		     /* listening */
+	false,		     /* data_ready */
+	true,		     /* readable */
+	true,		     /* writable */
+	0,		     /* remaining */
+	0,		     /* off */
+	{0}		     /* buf */
+};
+/*
+  This defines the Container SDNotify socket, attaches it to the correct FD and sets the flags for handling I/O.
+  setup_notify_socket() is responsible for initializing the unix sockets and pushing it onto the queue.
+
+  If the local_notify_host_fd stays -1 (i.e. we have not requested SD-NOTIFY) then setup was never run and
+  this has no effect.
+*/
+static int local_notify_host_fd = -1;
+static struct sockaddr_un local_notify_host_addr = {0};
+static struct local_sock_s local_notify_host = {&local_notify_host_fd, false, NULL, "host notify socket", &local_notify_host_addr};
+struct remote_sock_s remote_notify_sock = {
+	SOCK_TYPE_NOTIFY,   /* sock_type */
+	-1,		    /* fd */
+	&local_notify_host, /* dest */
+	false,		    /* listening */
+	false,		    /* data_ready */
+	true,		    /* readable */
+	false,		    /* writable */
+	0,		    /* remaining */
+	0,		    /* off */
+	{0}		    /* buf */
+};
+
+/* External */
 char *setup_console_socket(void)
 {
 	struct sockaddr_un addr = {0};
@@ -60,16 +110,48 @@ char *setup_console_socket(void)
 
 char *setup_attach_socket(void)
 {
-	struct sockaddr_un attach_addr = {0};
-	attach_addr.sun_family = AF_UNIX;
+	char *symlink_dir_path = bind_unix_socket("attach", SOCK_SEQPACKET | SOCK_NONBLOCK | SOCK_CLOEXEC, 0700, &remote_attach_sock, 0);
+
+	if (listen(remote_attach_sock.fd, 10) == -1)
+		pexitf("Failed to listen on attach socket: %s/%s", symlink_dir_path, "attach");
+
+	g_unix_fd_add(remote_attach_sock.fd, G_IO_IN, attach_cb, &remote_attach_sock);
+
+	return symlink_dir_path;
+}
+
+void setup_notify_socket(char *socket_path)
+{
+	/* Connect to Host socket */
+	if (local_notify_host_fd < 0) {
+		local_notify_host_fd = socket(AF_UNIX, SOCK_DGRAM | SOCK_NONBLOCK | SOCK_CLOEXEC, 0);
+		if (local_notify_host_fd == -1) {
+			pexit("Failed to create notify socket");
+		}
+		local_notify_host_addr.sun_family = AF_UNIX;
+		strncpy(local_notify_host_addr.sun_path, socket_path, sizeof(local_notify_host_addr.sun_path) - 1);
+	}
+
+	_cleanup_free_ char *symlink_dir_path =
+		bind_unix_socket("notify/notify.sock", SOCK_DGRAM | SOCK_NONBLOCK | SOCK_CLOEXEC, 0777, &remote_notify_sock, 1);
+	g_unix_fd_add(remote_notify_sock.fd, G_IO_IN | G_IO_HUP | G_IO_ERR, remote_sock_cb, &remote_notify_sock);
+}
+
+/* REMEMBER to g_free() the return value! */
+static char *bind_unix_socket(char *socket_relative_name, int sock_type, mode_t perms, struct remote_sock_s *remote_sock, gboolean do_chdir)
+{
+	int socket_fd = -1;
+	struct sockaddr_un socket_addr = {0};
+	socket_addr.sun_family = AF_UNIX;
+	_cleanup_free_ char *cwd = NULL;
 
 	/*
 	 * Create a symlink so we don't exceed unix domain socket
 	 * path length limit.
+	 *
+	 * We do NOT free this because it's returned to the parent, who is responsible for freeing it!
 	 */
-	char *attach_symlink_dir_path = g_build_filename(opt_socket_path, opt_cuuid, NULL);
-	if (unlink(attach_symlink_dir_path) == -1 && errno != ENOENT)
-		pexit("Failed to remove existing symlink for attach socket directory");
+	char *base_path = g_build_filename(opt_socket_path, opt_cuuid, NULL);
 
 	/*
 	 * This is to address a corner case where the symlink path length can end up being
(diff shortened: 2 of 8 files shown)


goochjj:
@haircommander PTAL

This is another take on what I had done w/ the syslog proxy. You had requested the console sockets and udp sockets share more code - I've done that by moving some flags and state into the structures used to manage the sockets.  In addition this implements a SD-NOTIFY proxy using the same constructs.

SD-NOTIFY conversation from here https://github.com/containers/libpod/issues/6688

src/conmon.c
```diff
@@ -84,6 +84,15 @@ int main(int argc, char *argv[])
 		exit(0);
 	}
 
+	if (getenv("NOTIFY_SOCKET") != NULL) {
```
haircommander, line 87: idk what's involved in a getenv call, but maybe we should save the value here and use it below?

rhatdan:
I am a little worried about SELinux here, making sure that the container processes are still allowed to write the SD_NOTIFY socket.  The path needs to be writeable from the container, and the  listener has to be ok.
Most likely this will work, but we should test this on an SELinux enforced system, with the container writing to the socket.

goochjj:
> I am a little worried about SELinux here, making sure that the container processes are still allowed to write the SD_NOTIFY socket. The path needs to be writeable from the container, and the listener has to be ok.
> Most likely this will work, but we should test this on an SELinux enforced system, with the container writing to the socket.

Agreed, I need to revisit podman doing the bind mount - specifically setting the mount label appropriately.

Also even though I'm creating the socket as 777, it's getting reset to 750.  Haven't figured that out yet.

goochjj:
> I am a little worried about SELinux here, making sure that the container processes are still allowed to write the SD_NOTIFY socket. The path needs to be writeable from the container, and the listener has to be ok.
> Most likely this will work, but we should test this on an SELinux enforced system, with the container writing to the socket.

libpod now relabels /run/notify

goochjj:
> > I am a little worried about SELinux here, making sure that the container processes are still allowed to write the SD_NOTIFY socket. The path needs to be writeable from the container, and the listener has to be ok.
> > Most likely this will work, but we should test this on an SELinux enforced system, with the container writing to the socket.
> 
> libpod now relabels /run/notify

And I get this 
```
[589620.624683] audit: type=1400 audit(1593097271.023:4559): avc:  denied  { sendto } for  pid=1644255 comm="systemd-notify" path="/var/lib/containers/storage/overlay-containers/7393144d563f1d3acb7a33003cbc43077d62748484963a5d9184406c4a7589e9/userdata/notify/notify.sock" scontext=system_u:system_r:container_init_t:s0:c355,c461 tcontext=unconfined_u:system_r:container_runtime_t:s0 tclass=unix_dgram_socket permissive=0
```
Do I need to change the tcontext somehow? Suggestions?

src/conn_sock.c
```diff
@@ -13,14 +13,46 @@
...
-static gboolean attach_cb(int fd, G_GNUC_UNUSED GIOCondition condition, G_GNUC_UNUSED gpointer user_data);
-static gboolean conn_sock_cb(int fd, GIOCondition condition, gpointer user_data);
-static gboolean read_conn_sock(struct conn_sock_s *sock);
-static gboolean terminate_conn_sock(struct conn_sock_s *sock);
-void conn_sock_shutdown(struct conn_sock_s *sock, int how);
-static void sock_try_write_to_masterfd_stdin(struct conn_sock_s *sock);
-static gboolean masterfd_write_cb(G_GNUC_UNUSED int fd, G_GNUC_UNUSED GIOCondition condition, G_GNUC_UNUSED gpointer user_data);
-
+static gboolean attach_cb(int fd, G_GNUC_UNUSED GIOCondition condition, gpointer user_data);
+static gboolean rmt_sock_cb(int fd, GIOCondition condition, gpointer user_data);
```
haircommander, line 17: what is rmt and lcl?

goochjj: see conn_sock.h

goochjj:
Note @rhatdan this does work because in podman I'm setting an appropriate label on the bind mount.  Tested with and without selinux, root and rootless.

I did hardcode in the NOTIFY_SOCKET inside the container, if we want that to be optional - maybe an additional env variable of something like NOTIFY_SOCKET_INSIDE or something can indicate the destination?

src/conmon.c
```diff
@@ -84,6 +84,16 @@ int main(int argc, char *argv[])
 		exit(0);
 	}
 
+	char *notify_socket_path = getenv("NOTIFY_SOCKET");
```
haircommander, line 87: is there a reason to expose this via env vs cli?

src/conn_sock.c
```diff
@@ -13,14 +13,46 @@
...
+static void init_rmt_sock(struct rmt_sock_s *sock, struct rmt_sock_s *src);
+static gboolean read_rmt_sock(struct rmt_sock_s *sock);
+static gboolean terminate_rmt_sock(struct rmt_sock_s *sock);
+static void rmt_sock_shutdown(struct rmt_sock_s *sock, int how);
+static void schedule_lcl_sock_write(struct lcl_sock_s *lcl_sock);
+static void sock_try_write_to_lcl_sock(struct rmt_sock_s *sock);
+static gboolean lcl_sock_write_cb(G_GNUC_UNUSED int fd, G_GNUC_UNUSED GIOCondition condition, G_GNUC_UNUSED gpointer user_data);
+
+static struct lcl_sock_s lcl_masterfd_stdin = {&masterfd_stdin, true, NULL, "container stdin", NULL};
+struct rmt_sock_s rmt_attach_sock = {
```
haircommander, line 27: can we have more explaination here on the reasoning behind setting these values and what these structures represent?

src/conn_sock.c
```diff
@@ -13,14 +13,46 @@
...
+	true,		     /* readable */
+	true,		     /* writable */
+	0,		     /* remaining */
+	0,		     /* off */
+	{0}		     /* buf */
+};
+static int lcl_notify_host_fd = -1;
+static struct sockaddr_un lcl_notify_host_addr = {0};
+static struct lcl_sock_s lcl_notify_host = {&lcl_notify_host_fd, false, NULL, "host notify socket", &lcl_notify_host_addr};
+struct rmt_sock_s rmt_notify_sock = {
```
haircommander, line 42: same here

src/conn_sock.c
```diff
@@ -111,51 +145,150 @@ char *setup_attach_socket(void)
...
+	 * the same as the socket.  When it happens, the symlink prevents the socket from being
+	 * be created.  This could still be a problem with other containers, but it is safe
+	 * to assume the CUUIDs don't change length in the same directory.  As a workaround,
+	 *  in such case, make the symlink one char shorter.
+	 */
+	if (strlen(notify_symlink_dir_path) == (sizeof(notify_addr.sun_path) - 1))
+		notify_symlink_dir_path[sizeof(notify_addr.sun_path) - 2] = '\0';
+
+	if (symlink(opt_bundle_path, notify_symlink_dir_path) == -1)
+		pexit("Failed to create symlink for notify socket");
```
haircommander, line 189: we could create a helper function with this, as I'm pretty sure it's duplicated with setup_console_socket

src/conn_sock.c
```diff
@@ -111,51 +145,150 @@ char *setup_attach_socket(void)
...
+		pexitf("Failed to remove existing notify socket: %s", notify_sock_fullpath);
+
+	if (bind(notify_socket_fd, (struct sockaddr *)&notify_addr, sizeof(struct sockaddr_un)) == -1)
+		pexitf("Failed to bind notify socket: %s", notify_sock_fullpath);
+
+	if (chmod(notify_sock_fullpath, 0777))
+		pexit("Failed to change notify socket permissions");
+
+	if (chdir(cwd) == -1)
+		pexitf("Could not chdir to %s", cwd);
```
haircommander, line 228: same here, I am pretty sure there's a decent amount of duplication here

src/conn_sock.c
haircommander: I think I'd prefer `local` and `remote` here. the number of chars isn't that much more

src/conn_sock.h
```diff
@@ -4,9 +4,33 @@
...
+   In most cases in conn_sock.c, this struct is "INPUT" and
+   the next one is "OUTPUT".  Really it's this struct is "one"
+   and the next is "many", but I don't want the same fd in
+   two different sockets.
+
+   rmt seems to indicate "Remote User" i.e. attached console, or
+       "container's /dev/log" or "container's /run/notify"
+   lcl seems to incidate "A socket we own, locally" i.e. masterfd_stdin
+      or "host /dev/log" or "host /run/systemd/notify"
+  */
```
haircommander, line 28: I think you've chosen the right nomenclature here, can you update the comment to be more sure

src/globals.c
```diff
@@ -7,8 +7,6 @@ int masterfd_stdin = -1;
 int masterfd_stdout = -1;
 int masterfd_stderr = -1;
 
-GPtrArray *conn_socks = NULL;
```
haircommander, line 10: :tada: very happy about this change

haircommander:
sorry it took so long to finally review. all together it looks great. thank you very much for taking this on.

I'd also like a review from @giuseppe

src/conn_sock.c
goochjj: Agreed. I think actually between attach and notify right now, the symlink is being removed, created, socket bound.. then the next one removes it, creates the symlink, binds the socket... So it's def duplicated.

I'd like your read on the approach here - I found with rootless that using "notify/notify.sock" as the socket name exceeded the unix socket path.  The nice thing is, since I hardcoded it to always be notify/notify.sock, I know exactly how long it's going to be ... The downside being that it will ALWAYS be notify.sock in the container... we could have some play with the target directory if we implement an ENV or a cmdline option for that.  But regardless in rootless it was exceeding the length of sun_path, so I chdir'd to the user_socket_directory so I could be sure the relative path from there was < sizeof(sun_path)

If that approach is solid, it should probably be done that way in both attach and notify  (console for some reason takes a different approach of just creating a socket in /tmp?)... For that matter symlinks could be removed entirely and we could use the bundledata path if we just use ./attach and ./notify/notify.sock instead of putting the whole path in there.

src/conn_sock.h
goochjj: Heh sure

src/conmon.c
goochjj: Just that it's how the notify socket is usually passed from systemd et al - but podman is always our parent, so I'm fine with whatever.

Maybe --notify-socket-host and --notify-socket-destdir or something - that can get rid of the hardcoded path of /run/notify/notify.sock in the container, and instead have it be (notify-socket-destdir)/notify.sock in the container.

goochjj:
Also FYI, even if the notify socket cmdline/env are passed.. right now podman mkdir's the notify folder because I have it set the mount label, and pass the mount label as a bind mount to the spec.  If the calling process doesn't create that socket, but a NOTIFY_SOCKET is passed, it'll error with a cryptic error.

It was just an order of operations thing - that the runtime spec and mounts get done before conmon is called.

We could, instead, throw a nice error , like "notify socket passed but no folder created" or something.

src/conn_sock.c
haircommander: ah I didn't notice the chdir....

I am unsure of the approach to be honest. If we were going to run the conmon in a specific dir, we should make clear in documentation that is where.

does just `notify.sock` exceed the length? why have the specific dir? The container directory passed should be mostly touched by conmon and its caller, so it doesn't necessarily have to be very pretty.

I think I'd prefer to not have the chdir and instead go through the linking shenannigans we currently do

src/conmon.c
haircommander: yes I like that idea. then it can be exposed via cli on podman's side if the user wants to change it.

giuseppe: could we include the commit `52546c919be261e74847981b4db1227537c47345` into the commit that first introduced the change?

Also, I think we need a way to disable the notify socket to be handled by conmon, there are cases when it is helpful to use the handling in the OCI runtime.  For example when the user needs `$RUNTIME start $CTR` to wait for the container to be ready.

goochjj:
> thanks for the patch. I think it is a great idea but I've not managed to get it to work:
> 
> ```
> $ NOTIFY_SOCKET=/run/user/1000/systemd/notify podman --conmon ~/src/conmon/bin/conmon  run --rm -ti fedora sh -c 'printenv NOTIFY_SOCKET; systemd-notify --ready'
> Error: container create failed (no logs from conmon): EOF
> ```

@giuseppe you need https://github.com/containers/conmon/pull/182

goochjj:
@giuseppe re: commit, yes, I would squash this whole PR before merging.

goochjj:
@giuseppe Re: letting OCI runtime do Notify - I'm ok with that, but I think it basically means extending the --sdnotify option in podman.

from https://github.com/containers/libpod/issues/6693:

--sdnotify conmon-only|container|none

With "conmon-only", we send the MAINPID, and clear the NOTIFY_SOCKET so the OCI
runtime doesn't pass it into the container. We also advertise "ready" when the
OCI runtime finishes to advertise the service as ready.

With "container", we send the MAINPID, and leave the NOTIFY_SOCKET so the OCI
runtime passes it into the container for initialization, and let the container advertise further metadata.

The "none" option does what it's always done in the past - passes NOTIFY_SOCKET
to conmon's process and the OCI runtime processes, and does no manipulation or "help".

This removes the need for hardcoded CID and PID files in the command line, and
the PIDFile directive, as the pid is advertised directly through sd-notify.


Perhaps "container" should be split.  I'm not sure into what and what.  Maybe "ociruntime" and "podman".  Then we'd have the groundwork to tweak the options passed to conmon.

The advantage of having conmon do the proxy is exactly to prevent CRUN/RUNC from blocking on start, to fix https://github.com/containers/libpod/issues/6688.

src/conmon.c
```diff
@@ -84,6 +84,16 @@ int main(int argc, char *argv[])
...
+	if (notify_socket_path != NULL) {
+		setup_notify_socket(notify_socket_path);
+		int r = unsetenv("NOTIFY_SOCKET");
```
giuseppe, line 90: how would `--sdnotify none` work if we always unset `NOTIFY_SOCKET`?

And who will set the bind mount inside the container when we won't pass `NOTIFY_SOCKET` to the OCI runtime?

goochjj: Podman does.

We unset this because we DON'T want the OCI runtime to have it.  Podman has already set a mount label and added a bind mount to the OCI spec. 

Essentially instead of setting NOTIFY_SOCKET in the env, podman sets a NOTIFY_SOCKET in the Environment part of the runtime spec, and adds a bind mount to the Mounts area.

That's part of this PR
https://github.com/goochjj/libpod/pull/1
(hence why you need it)

If you looked at it look again, I just rebased it on the most recent libpod-sd-notify-cmdline (which it's dependent on)

goochjj: If we turn this into a cmdline option, we just wouldn't set the NOTIFY_SOCKET in podman, and this wouldn't be necessary.

mheon:
@goochjj Can you rebase?

Let's see about getting this moved forward and merged, I'd love to have this in Podman 2.1.x

goochjj:
@haircommander @mheon What's the deal with the tests...

haircommander:
f28 and f29 are EOL, and the images haven't been updated. It is on my list this week to update them. sorry for the inconvenience

goochjj:
Heh no worries, as long as it's not a blocker :-D

I did run `make fmt` locally

rhatdan:
@haircommander @goochjj Other then the static build fail above is this ready to go in?  Would like to get this in to close a Podman  issue.

goochjj:
Yep, should be.

haircommander:
I think there are some unresolved review comments from @giuseppe

goochjj:
> I think there are some unresolved review comments from @giuseppe

If so you'll need to tell me what needs to happen.  Last I can see, I responded to @giuseppe and he didn't respond back.

src/conn_sock.c
```diff
@@ -111,51 +145,150 @@ char *setup_attach_socket(void)
...
 	return attach_symlink_dir_path;
 }
 
-static gboolean attach_cb(int fd, G_GNUC_UNUSED GIOCondition condition, G_GNUC_UNUSED gpointer user_data)
+void setup_notify_socket(char *socket_path)
+{
+	int notify_socket_fd = -1;
+	struct sockaddr_un notify_addr = {0};
+	notify_addr.sun_family = AF_UNIX;
+	char cwd[1024];
```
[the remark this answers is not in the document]
haircommander: was talking about these nits @goochjj

src/conn_sock.c
```diff
@@ -111,51 +145,150 @@ char *setup_attach_socket(void)
...
+
+	strncpy(notify_addr.sun_path, notify_sock_path, sizeof(notify_addr.sun_path) - 1);
+	ninfof("addr{sun_family=AF_UNIX, sun_path=%s}", notify_addr.sun_path);
+
+	/*
+	 * We make the socket non-blocking to avoid a race where client aborts connection
+	 * before the server gets a chance to call accept. In that scenario, the server
+	 * accept blocks till a new client connection comes in.
+	 */
+	if (getcwd(cwd, sizeof(cwd)) == NULL)
```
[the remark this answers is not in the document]
haircommander: and this one

src/conn_sock.c
haircommander: this is still unresolved, I am hesitant of the chdir

haircommander:
I'm sorry about the nitpicking @goochjj if you are done with carrying this PR, we can have someone else take over it. A lot relies on conmon and the testing is really subpar, so please excuse my anxiety

goochjj:
Let me go through and see if I can verify the resolved things

src/conn_sock.c
goochjj: Why are you hesitant? It chdirs back to CWD at the end...

src/conn_sock.c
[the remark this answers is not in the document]
goochjj: Then I don't have a buffer, and have to run a malloc.  It's function local, why is yours more preferable?

src/conn_sock.c
goochjj: haircommander:cri-o:  10:25 AM
I've never used it in a c program and I am worried about suddenly relying on relative paths. also, if we exceed PATH_MAX, won't that fail in other ways and places other than conmon?
I guess ultimately it feels hacky (despite the other work around being similarly hacky)

Joseph Gooch  10:26 AM
It IS hacky, we’re working around unix socket path length issues :smile:

haircommander:cri-o:  10:27 AM
oh right lol

Joseph Gooch  10:27 AM
I chdir, and then chdir back at the end of the function
if we exceed PATH_MAX, I imagine chdir will fail (which is fatal)

haircommander:cri-o:  10:28 AM
are you opposed to just doing notify.sock without the notify directory?

Joseph Gooch  10:28 AM
Can’t
Podman has to bindmount a directory
To set selinux labels
Plus the bind mount happens at a different time the unix socket is created
So Podman exposes the folder, conmon owns the socket within the folder
Otherwise bindmounting the socket itself gets… problematic
And we wouldn’t want to bindmount our entire /var or /run podman state folder

haircommander:cri-o:  10:31 AM
okay I'm convinced. not super stoked about it, but convinced :slightly_smiling_face:

src/conn_sock.c
goochjj: Slightly different, since we chdir - Let me look closer.

goochjj:
Resolved just about everything except the open "Should we use ENV or a command-line argument"

src/conn_sock.c
haircommander: I'd be in favor of pulling out the common bits to reduce duplication

goochjj: Dun

goochjj:
@haircommander As far as I know I've completed all I need to - let me know if there's something I need to look at, or you want to further discuss ENV vs cmdline.  I can squash the PR whenever you want.

src/conmon.c
goochjj: Done

src/conmon.c
```diff
@@ -89,6 +89,11 @@ int main(int argc, char *argv[])
 	/* before we fork, ensure our children will be reaped */
 	atexit(reap_children);
 
+	/* If we were passed a sd-notify socket to use, set it up now */
+	if (opt_sdnotify_socket != NULL) {
```
haircommander, line 93: can we just do `if (opt_sdnotify_socket)`

goochjj: AYW

haircommander:
test for cri-o regressions: https://github.com/cri-o/cri-o/pull/4187

haircommander:
hold this, there are some failures I need to investigate

rhatdan:
@haircommander Whats up with this one?

src/conn_sock.c
```diff
@@ -165,40 +285,63 @@ static gboolean read_conn_sock(struct conn_sock_s *sock)
...
+				if (!opt_leave_stdin_open) {
+					close(*(sock->dest->fd));
+					*(sock->dest->fd) = -1;
+				} else {
+					ninfo("Not closing input");
+				}
+			}
 		} else {
-			ninfo("Not closing input");
+			ninfo("Not closing input - still have open sockets");
```
haircommander, line 338: what's the motivation behind this change? I believe it may be causing crio tests to fail https://storage.googleapis.com/origin-federated-results/pr-logs/pull/cri-o_cri-o/4187/test_pull_request_crio_critest_fedora/19310/artifacts/journals/crio.service

goochjj: I think the idea was to distinguish between the two branches. (line 334 vs line 338). So I could tell during testing which branch it ended up on.

!opt_leave_stdin_open isn't an error condition.
Technically line 338 it's left open because there are still readers.

If you want to change it back or your tests are looking for specific words, obvi it won't break anything.

goochjj: Is your question regarding the verbiage, or that a message is generated at all?

goochjj: applied test commit

haircommander:
LGTM assuming https://github.com/cri-o/cri-o/pull/4187 passes
@giuseppe PTAL

giuseppe (approved):
LGTM

PR #183, opened 2020-06-24
add options to handle children differently

haircommander:
This includes:
unconditionally reap children as it doesn't hurt, and we've seen cases where the children are not cleaned up after a clean exit (causing excess load for systemd)
add --sync to allow conmon to only fork once, allowing the caller to have fewer children
also, make sure we don't call atexit callbacks for failed children after fork() by replacing calls to pexit() with _pexit()

### DIFF
--- a/src/conmon.c
+++ b/src/conmon.c
@@ -66,23 +66,28 @@ int main(int argc, char *argv[])
 	if (dev_null_w < 0)
 		pexit("Failed to open /dev/null");
 
-	/* In the create-container case we double-fork in
-	   order to disconnect from the parent, as we want to
-	   continue in a daemon-like way */
-	pid_t main_pid = fork();
-	if (main_pid < 0) {
-		pexit("Failed to fork the create command");
-	} else if (main_pid != 0) {
-		if (opt_conmon_pid_file) {
-			char content[12];
-			sprintf(content, "%i", main_pid);
+	/* In the non-sync case, we double-fork in
+	 * order to disconnect from the parent, as we want to
+	 * continue in a daemon-like way */
+	if (!opt_sync) {
+		pid_t main_pid = fork();
+		if (main_pid < 0) {
+			pexit("Failed to fork the create command");
+		} else if (main_pid != 0) {
+			if (opt_conmon_pid_file) {
+				char content[12];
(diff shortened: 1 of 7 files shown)


src/ctr_exit.c
```diff
@@ -131,6 +131,14 @@ void container_exit_cb(G_GNUC_UNUSED GPid pid, int status, G_GNUC_UNUSED gpointe
 	g_main_loop_quit(main_loop);
 }
 
+void reap_children()
+{
+	/* We need to reap any zombies (from an OCI runtime that errored) before
+	   running it. */
```
mheon, line 137: Suggest changing `running it` to `exiting`

src/conmon.c
```diff
@@ -289,6 +289,7 @@ int main(int argc, char *argv[])
 	if (signal(SIGCHLD, on_sigchld) == SIG_ERR)
 		pexit("Failed to set handler for SIGCHLD");
 
+	atexit(reap_children);
```
mheon, line 292: From looking at the documentation, it looks like multiple `atexit` functions are executed in reverse order, so this will be executed after the exit delay, thus reintroducing the Podman issue - maybe reorder this to after exit command?

haircommander: nice good catch. unfortunately, we need to run it earlier, so there'll just be multiple calls :woman_shrugging:

haircommander:
> I don't really understand `conmon` but I'm very happy to see a `--sync` argument just because not checking return values is usually a source for bugs.
> 
> However, README.md is very particular about double-fork. Would you document the suggested execution mode there?

I think the wording of the PR was not quite right. this isn't handling it "better" but allowing a different option. The default mode should still be double-fork: that way, the container processes can not be direct children of their caller.

this option is most useful when we want the entire`$runtime` process to run before execution of the calling process ends, like in the case where CRI-O gets a ContainerExecSync call.

I will make a docs update to illuminate the option, though.

haircommander:
docs updated

haircommander:
conflicts have been fixed, but github is lagging today...

mrunalp:
needs rebase

mrunalp:
LGTM

mrunalp:
@giuseppe ptal

PR #194, opened 2020-08-03
Use Podman for building

rhdojun:
I liked the idea from the podman repo to build stuff inside a container. It keeps the host clean. So tried to reproduce it for this repo as well. Have a look and let me know if you like it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -105,3 +105,13 @@ fmt:
 	find . '(' -name '*.h' -o -name '*.c' ! -path './vendor/*' ')'  -exec clang-format -i {} \+
 	find . -name '*.go' ! -path './vendor/*' -exec gofmt -s -w {} \+
 	git diff --exit-code
+
+
+.PHONY: dbuild
+dbuild:
+	-mkdir -p bin
+	-podman rm conmon-devenv
+	podman build -t conmon-devenv:latest .
+	podman create --name conmon-devenv conmon-devenv:latest
+	podman cp conmon-devenv:/devenv/bin/conmon bin/conmon
+	@echo "for installation move conmon file to /usr/local/libexec/podman/"
--- a/Containerfile
+++ b/Containerfile
@@ -1,0 +1,20 @@
+FROM registry.fedoraproject.org/fedora:latest
+
+RUN sudo dnf install -y make automake gcc gcc-c++ kernel-devel glib2-devel && \
+    sudo dnf clean all && \
+    rm -rf /var/cache/dnf
+
+RUN sudo dnf update -y && \
+    sudo dnf clean all && \
+    rm -rf /var/cache/dnf
+
+# replaces the mktemp from the tutorial as everything is temporary in a
+# container unless bind mounted out
+RUN mkdir -p /tmp/gocache
+ENV GOCACHE=/tmp/gocache
+
+RUN mkdir -p /devenv
+ADD . /devenv
+WORKDIR /devenv
+
+RUN make


Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM registry.fedoraproject.org/fedora:latest
```
TomSweeneyRedHat, line 1: If you're using Podman, I'd suggest naming this file `Containerfile` and not `Dockerfile`

Makefile
```diff
@@ -105,3 +105,13 @@ fmt:
 	find . '(' -name '*.h' -o -name '*.c' ! -path './vendor/*' ')'  -exec clang-format -i {} \+
 	find . -name '*.go' ! -path './vendor/*' -exec gofmt -s -w {} \+
 	git diff --exit-code
+
+
+.PHONY: dbuild
+dbuild:
+	-mkdir -p bin
+	-podman rm conmon-devenv
```
TomSweeneyRedHat, line 113: Is the starting dash (`-`) correct in this line and the one prior?

TomSweeneyRedHat:
Looks like problems pulling Fedora:29?

Dockerfile
rhdojun: I did not know this is an option. Will do!

Makefile
rhdojun: It depends.

With the `mkdir` the `-p` might be enough. I'll remove it if you consider it better. But I also don't see how it could hurt.

With the `podman rm` I think it's good because:

```
14:57:54:conmon🍔  podman rm i-do-not-exist
Error: no container with name or ID i-do-not-exist found: no such container
14:58:00:conmon🍔  echo $?
1                                                       <- it would error
```

haircommander:
thanks for the PR @rhdojun !

CI failures are not the fault of this PR, the CI is currently busted

Makefile
```diff
@@ -105,3 +105,13 @@ fmt:
 	find . '(' -name '*.h' -o -name '*.c' ! -path './vendor/*' ')'  -exec clang-format -i {} \+
 	find . -name '*.go' ! -path './vendor/*' -exec gofmt -s -w {} \+
 	git diff --exit-code
+
+
+.PHONY: dbuild
+dbuild:
```
haircommander, line 111: is dbuild short for docker build? We may want a different name here,. unless this is standardized across projects

Makefile
TomSweeneyRedHat: @rhdojun I'm fine with or without it.  I was just questioning as  I don't recall seeing one before and wasn't sure if it was meant to be there, and if so, what it's purpose was.

Makefile
rhdojun: Well, it's mostly a [copy&paste from your partner repo](https://github.com/containers/podman/blob/master/Makefile#L268-L278) and I would like to keep the stuff in sync as I mostly build it together.

rhdojun:
I'll check the build results once I get a round tuit. What surprises me is that the image pull may not work. Do I have a special setup with my RHEL CSB that authorizes there? Because for me this worked. If nobody has a specific problem in the back of their head I suggest to retrigger the build.

rhdojun:
> CI failures are not the fault of this PR, the CI is currently busted

Ah sorry, this I only see now.

Makefile
rhdojun: Ah I see. Usually make exits with an error code when a command fails. With the dash before the command you can ensure that an error won't result in an exit. So it's a great help for cleanup and prepare steps that can fail if nothing is to be cleaned or prepared. The pattern is also reused in other tools, like [for instance systemd](https://github.com/rhdojun/simpleserver/blob/master/simpleserver.service#L8-L12).

TomSweeneyRedHat: @rhdojun tyvm for the education!  :clinking_glasses:

haircommander:
LGTM, force merging because there are no code changes

PR #200, opened 2020-08-26
Remove go.sum file and add go.mod

saschagrunert:
We excluded the go.mod via the .gitignore file, which should not be
correct. Beside that, we do not need a go.sum file because we do not
have external dependencies yet.


haircommander:
ah that makes sense I didn't notice go.mod wasn't included
while we're at it, can you remove the
```
# the || true is to make this call idempotent, in case we already have initialized the module
go mod init || true
```
from contrib/cirrus/integration_test.sh ? that should fix the config task

saschagrunert:
> ah that makes sense I didn't notice go.mod wasn't included
> while we're at it, can you remove the
> 
> ```
> # the || true is to make this call idempotent, in case we already have initialized the module
> go mod init || true
> ```
> 
> from contrib/cirrus/integration_test.sh ? that should fix the config task

Hah, yeah sure that makes sense. Fixed that, too. :+1:

haircommander:
oop, sorry, there's also an instance in .cirrus.yml

saschagrunert:
> oop, sorry, there's also an instance in .cirrus.yml

No worries, fixed that too :+1:

rhatdan:
Any idea why the static builds are failing?

saschagrunert:
> Any idea why the static builds are failing?

The master rebase should fit the issue.

haircommander:
LGTM

PR #201, opened 2020-08-26
Fail if conmon config could not be written

saschagrunert:
`fmt.Errorf` returns a result which has not being used in
conmon-config.go before this patch. We now `log.Fatal` the error to
actually report to the user that config generation has been failed.


haircommander:
oops :open_mouth: LGTM

PR #203, opened 2020-09-02
protect against kill(-1)

haircommander:
if kill is passed a process that is less than -1, it kills every pid in the process group.
However, if it is passed -1, then conmon attempts to kill everything it can

protect against the case that getpgid returns a 1

### DIFF
--- a/src/conmon.c
+++ b/src/conmon.c
@@ -443,8 +443,9 @@ int main(int argc, char *argv[])
 	 */
 	if (timed_out && container_pid > 0) {
 		pid_t process_group = getpgid(container_pid);
-
-		if (process_group > 0)
+		/* if process_group is 1, we will end up calling
+		 *  kill(-1), which kills everything conmon is allowed to. */
+		if (process_group > 1)
 			kill(-process_group, SIGKILL);
 		else
 			kill(container_pid, SIGKILL);


src/conmon.c
```diff
@@ -444,7 +444,7 @@ int main(int argc, char *argv[])
 	if (timed_out && container_pid > 0) {
 		pid_t process_group = getpgid(container_pid);
 
-		if (process_group > 0)
+		if (process_group > 1)
```
mrunalp, line 447: Can we add a comment?

haircommander: done!

haircommander:
ptal @giuseppe (if you happen to be checking your email, which you should not be :palm_tree: )

giuseppe:
How does it work in the sync case?  Do we have a new process group?

haircommander:
what do you mean by the sync case?

giuseppe:
When conmon doesn't double fork

haircommander:
in that case it seems the process group is the same as the container pid

giuseppe:
Well I am fine with the fix, I am just wondering why we need to go through the process group at all but we can change that later

rhatdan:
LGTM

PR #207, opened 2020-09-11
Update nix pin with `make nixpkgs`

### LINKED ISSUE #206: Static binary not uploaded in cirrus
The run for the static binary seems to succeed, but we have no binary being uploaded:
https://cirrus-ci.com/task/6040168039710720

cc @hswong3i

hswong3i:
Backport changes from https://github.com/cri-o/cri-o/pull/4065/files#diff-1d37e48f9ceff6d8030570cd36286a61R189-R197

Also update systemd 246 as per https://github.com/NixOS/nixpkgs/pull/94354

Fix #206

### DIFF
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -11,6 +11,8 @@ let
     config = {
       packageOverrides = pkg: {
         autogen = (static pkg.autogen);
+        e2fsprogs = (static pkg.e2fsprogs);
+        libuv = (static pkg.libuv);
         glib = (static pkg.glib).overrideAttrs(x: {
           outputs = [ "bin" "out" "dev" ];
           mesonFlags = [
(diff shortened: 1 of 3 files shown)


nix/default.nix
```diff
@@ -30,18 +32,10 @@ let
           ];
         });
         systemd = (static pkg.systemd).overrideAttrs(x: {
+          outputs = [ "bin" "out" "lib" ];
           mesonFlags = x.mesonFlags ++ [
             "-Dstatic-libsystemd=true"
           ];
```
hswong3i, line 38: @saschagrunert seems like this changes for systemd 246 working with crun (https://github.com/containers/crun/pull/485/files#diff-990d93ae620a3c4824fd6544548bf806L29-L37) but not here.

I think most likely related to our static pcre overlays. May I have some hints for this?

saschagrunert: This looks good, doesn’t it? 👀

saschagrunert: I think we needed the PCRE overlay because of building glib statically, right?

Hmmm, this build error looks indeed strange:
```
PermissionError: [Errno 13] Permission denied: '//02qcpld1y6xhs5gz9bchpxaw0xdhmsp5dv88lh25r2ss44kh8dxz'
```

hswong3i: Fixed, only `outputs = [ "out" "lib" ];` could specify :-S

saschagrunert (approved):
/lgtm
Nice!

haircommander:
thanks @hswong3i

PR #219, opened 2020-12-01
conn_sock: Explicitly free a heap-allocated string

tbaederr:
GCC does not complain about the variable here being unused, but clang
does:

../src/conn_sock.c:134:23: error: unused variable 'symlink_dir_path' [-Werror,-Wunused-variable]
        _cleanup_free_ char *symlink_dir_path =
                             ^
1 error generated.


haircommander:
thanks for the PR @tbaederr ! can you add a comment describing why we have it in two steps? ideally, we'd skip the clang warning to follow the rest of the code base, but I am okay with this also.

tbaederr:
Sure, I added a comment. Thanks.

haircommander:
LGTM

rhatdan:
LGTM

PR #222, opened 2020-12-18
conmon: store open FDs and close only them

giuseppe:
now that we use a delay to call the cleanup program, we might end up in a race where the event fd used by glib is close'd and it causes the glib event handler to keep polling the closed file descriptor in a tight loop.
    
To avoid closing files that are handled by glib, store what FDs are
opened when conmon first started and close only them.
    
Closes: https://github.com/containers/conmon/issues/221
    

src/conmon.c
```diff
@@ -22,6 +22,53 @@
 #include <sys/prctl.h>
 #include <sys/stat.h>
 
+static int open_files_max_fd;
```
haircommander, line 25: can these additions live in a different file?

src/conmon.c
```diff
@@ -464,24 +512,14 @@ int main(int argc, char *argv[])
...
-		closedir(fdsdir);
-	}
+	int fd;
+
+	for (fd = 3; fd < open_files_max_fd; fd++)
+		if (open_files_set == NULL || FD_ISSET (fd % FD_SETSIZE, &(open_files_set[fd / FD_SETSIZE])))
+			if (fd == sync_pipe_fd || fd == attach_pipe_fd || fd == dev_null_r || fd == dev_null_w
+			    || fd == oom_cgroup_fd || fd == oom_event_fd)
+				close (fd);
+	close_all_readers();
```
haircommander, line 522: and this could all also live in that new file

src/conmon.c
giuseppe: sure, moved to a new file

src/conmon.c
giuseppe: same

haircommander:
LGTM

rhatdan:
LGTM
@giuseppe Cleanup format and then we can get this merged and released.

giuseppe:
fixed clang-format errors

Issue #204, opened 2020-08-07
pasting large amount of text into vim running inside of podman hangs container


wdouglascampbell:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

When pasting a large amount of text into a container environment (usually while editing a file in vim but also to a /bin/bash and /bin/sh shell prompt)  the session hangs and there is no way to cancel, quit, exit, etc.

**Steps to reproduce the issue:**

1.  run a busybox container
```
podman run --rm -it busybox /bin/sh
```

2.  edit test file
```
vi test
```

3.  enter insert mode and paste large amount of text

**Describe the results you received:**

Terminal hangs.  I have to open new session and stop the podman container and sometimes that doesn't stop the process so I have to search for it and kill by the process id.

**Describe the results you expected:**

All content pasted is now in file being edited in vim.

**Additional information you deem important (e.g. issue happens only occasionally):**

This happens when running PuTTY 0.74.  It doesn't happen if running SSH from Windows 10 command line.

It also doesn't happen if I run the container using ```docker run --rm -it busybox /bin/sh``` instead of using podman.

**Output of `podman version`:**

```
Version:            1.9.3
RemoteAPI Version:  1
Go Version:         go1.14.2
OS/Arch:            linux/amd64
```

**Output of `podman info --debug`:**

```
debug:
  compiler: gc
  gitCommit: ""
  goVersion: go1.14.2
  podmanVersion: 1.9.3
host:
  arch: amd64
  buildahVersion: 1.14.9
  cgroupVersion: v1
  conmon:
    package: conmon-2.0.18-1.fc32.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.18, commit: 6e8799f576f11f902cd8a8d8b45b2b2caf636a85'
  cpus: 12
  distribution:
    distribution: fedora
    version: "32"
  eventLogger: file
  hostname: coreos.
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  kernel: 5.7.8-200.fc32.x86_64
  memFree: 6526885888
  memTotal: 67493203968
  ociRuntime:
    name: runc
    package: runc-1.0.0-144.dev.gite6555cc.fc32.x86_64
[...]

PR #223, opened 2020-12-18
conn_sock: make dest fd non blocking

giuseppe:
we already have code in place to handle partial writes, just make sure
the fd is not blocking so it doesn't hang on writes.

Closes: https://github.com/containers/conmon/issues/204


haircommander:
LGTM, thanks @giuseppe

rhatdan:
LGTM

Issue #225, opened 2021-01-04
Release v2.0.22 fails to run under Ubuntu 18.04


ksdd:
Release v2.0.22 fails to run under Ubuntu 18.04 with the following error message

> Error: failed to connect to container's attach socket: /run/user/205013812/libpod/tmp/socket/7c19123d148494b91f6689e0882d9f2b1b7b7bb8036128bbd4c363c11bbe3044/atta: dial unixpacket /run/user/205013812/libpod/tmp/socket/7c19123d148494b91f6689e0882d9f2b1b7b7bb8036128bbd4c363c11bbe3044/atta: connect: no such file or directory

I tested the binary provided by the [recommended repo](https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_18.04) as well as the [official release binary](https://github.com/containers/conmon/releases/download/v2.0.22/conmon-2.0.22.zip). Both fail to run with the same error.

BTW: When running the [official release binary](https://github.com/containers/conmon/releases/download/v2.0.22/conmon-2.0.22.zip) with --version option, I get:

> $ /usr/libexec/podman/conmon --version
> conmon version 2.0.23-dev
> commit: 715c680ed732f1e92166f03a7a94ff2f4c4f2f3e-dirty

Obviously the binary has been build from a dirty working copy and is **NOT** based on the commit 9c34a8663b85e479e0c083801e89a2b2835228ed  tagged as v2.0.22!

On the other hand the binary from the [recommended repo](https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_18.04) seems to be build based on v2.0.22 according to its version output.

haircommander:
how are you running conmon? is this with podman or CRI-O?

ksdd:
I run it via podman.

haircommander:
what was the podman command you used? and what was the podman version?

ksdd:
> $ podman --version
> podman version 2.2.1

> $ podman run --interactive --tty --rm alpine
> Error: failed to connect to container's attach socket: /run/user/205013812/libpod/tmp/socket/aa6712c41c455cd6c136c0af15d2345956ab9425e8690e4bd36845a59c384c83/atta: dial unixpacket /run/user/205013812/libpod/tmp/socket/aa6712c41c455cd6c136c0af15d2345956ab9425e8690e4bd36845a59c384c83/atta: connect: no such file or directory

haircommander:
[...]

Issue #213, opened 2020-10-22
Failed to remove existing attach socket


egreen77:
Firstly, I noticed that this section of `conmon` was recently refactored in 59c2817b70746e78d5957d3c275e6385697224d1. I haven't tried to replicate this issue with that version but suspect that those changes would also fix this error. I'm submitting this issue primarily as an advisory and to help anyone else who might run into this problem.

Environment:
```
$ cat /etc/redhat-release 
Red Hat Enterprise Linux Server release 7.8 (Maipo)
$ podman version
Version:            1.6.4
RemoteAPI Version:  1
Go Version:         go1.12.12
OS/Arch:            linux/amd64
$ conmon --version
conmon version 2.0.15
commit: 372b4a12f1c2df4f70c280d41173b60acd3f1260
```

One of our users has been reporting that attempting to run anything with podman fails with the following error: `Error: container create failed (no logs from conmon): EOF`. All other users on the system have had no issues. Debugging suggested that the error was triggered by a combination of the length of the user's UID and temp directory path.

 We were able to reproduce the issue by creating a new test user with the same length of UID and temp directory path:
```
username=test
uid=12345
TMPDIR=/var/tmp/test
```

Here is the strace output leading up to the error:
```
open("/var/tmp/test/conmon-term.2U29S0", O_RDWR|O_CREAT|O_EXCL, 0600) = 7
close(7)                                = 0
write(2, "[conmon:i]: addr{sun_family=AF_UNIX, sun_path=/var/tmp/test/conmon-term.2U29S0}\n", 80) = 80
socket(AF_UNIX, SOCK_STREAM|SOCK_CLOEXEC, 0) = 7
fchmod(7, 0700)                         = 0
unlink("/var/tmp/test/conmon-term.2U29S0") = 0
bind(7, {sa_family=AF_UNIX, sun_path="/var/tmp/test/conmon-term.2U29S0"}, 110) = 0
listen(7, 128)                          = 0
pipe2([8, 9], O_CLOEXEC)                = 0
unlink("/var/tmp/test/run-12345/libpod/tmp/socket/4dd4e6a3714e25e5fc0ceb68a599cb38db7c0a49063813f631929bf2a4545414") = 0
[...]

PR #226, opened 2021-01-11
conn_sock: add chdir-attach option

haircommander:
unfortunately, older podman versions rely on conmon truncating the unix socket length when setting up the attach socket.

Introduce a flag to gate the changed behavior, allowing newer podman versions to be able to handle longer UID in rootless mode,
but also allowing older podman versions to work.

### DIFF
--- a/src/conn_sock.c
+++ b/src/conn_sock.c
@@ -22,7 +22,9 @@ static void remote_sock_shutdown(struct remote_sock_s *sock, int how);
 static void schedule_local_sock_write(struct local_sock_s *local_sock);
 static void sock_try_write_to_local_sock(struct remote_sock_s *sock);
 static gboolean local_sock_write_cb(G_GNUC_UNUSED int fd, G_GNUC_UNUSED GIOCondition condition, G_GNUC_UNUSED gpointer user_data);
-static char *bind_unix_socket(char *socket_relative_name, int sock_type, mode_t perms, struct remote_sock_s *remote_sock);
+static char *bind_unix_socket(char *socket_relative_name, int sock_type, mode_t perms, struct remote_sock_s *remote_sock,
+			      gboolean use_full_attach_path);
+static char *get_sock_path(gboolean use_full_attach_path, char *base_path, size_t desired_len);
 /*
   Since our socket handling is abstract now, handling is based on sock_type, so we can pass around a structure
   that contains everything we need to handle I/O.  Callbacks used to handle IO, for example, and whether this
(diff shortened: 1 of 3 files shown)


haircommander:
fixes https://github.com/containers/conmon/issues/225

rhatdan:
go fmt is not happy.

rhatdan:
LGTM

haircommander:
ptal @giuseppe @saschagrunert @mheon

src/conn_sock.c
```diff
@@ -188,7 +202,7 @@ static char *bind_unix_socket(char *socket_relative_name, int sock_type, mode_t
 	if (fchmod(socket_fd, perms))
 		pexitf("Failed to change socket permissions %s", sock_fullpath);
 
-	if (chdir(opt_socket_path) == -1)
+	if (do_chdir && chdir(opt_socket_path) == -1)
```
giuseppe, line 205: Could we use the same trick with O_PATH I've used for Podman so we don't need a new flag and avoid using chdir?

haircommander: I think that's a good trick for the `do_chdir` case, but we still need this PR to maintain backwards compatibility with podman (that truncates the path once it gets long enough)

haircommander: acutally, that's a good trick for both cases, but we still will need to truncate the path sometimes, thus still need this flag

giuseppe: don't we need to teach podman how to use the new flag though?  How will it work with older versions?

haircommander: yeah new podman will use this flag to be able to fix https://github.com/containers/conmon/issues/213
#213 will not be able to be fixed with older versions of podman

haircommander: basically, there's no backwards compatible way to fix #213, so we need to gate the fix with a new flag

giuseppe: I still don't understand why we need a new flag.  Can't we fix the unlink with?

```diff
- 	if (unlink(attach_addr.sun_path) == -1 && errno != ENOENT)
+ 	if (unlink(attach_sock_path) == -1 && errno != ENOENT)
```

giuseppe: nevermind my last comment.

For backward compatibility could we bump the API version and use the full path version only when the API version is 2?

haircommander: jw, is there a reason you're opposed to a flag here? we could use api-version here, but it seems like a very small change to bump the conmon major version. and if we did, we'd still need a change in podman to actually absorb the change, which is no different than a flag

saschagrunert (approved):
LGTM from my side

src/conn_sock.c
giuseppe: the main reason is that I'd prefer if possible to avoid `chdir` as it affects the entire process

haircommander: okay I dropped chdir but kept a flag. Now we gate the behavior of truncating the path on a flag, but regardless don't  chdir

giuseppe: thanks!

giuseppe (approved):
/lgtm

PR #228, opened 2021-01-14
conn_sock: fix path opening

haircommander:
I must admit I didn't test correctly, and there were two regressions with the previous commit on conn_sock.
1: sock_fullpath doesn't exist at the time that we need to copy its location into sun_path. As such, we need to use the proc entry of the parent dir, and use that as the parent path.
2: we do need the symlink to the bundle_path, as that's where the attach file actually lives.

### DIFF
--- a/src/conn_sock.c
+++ b/src/conn_sock.c
@@ -24,7 +24,7 @@ static void sock_try_write_to_local_sock(struct remote_sock_s *sock);
 static gboolean local_sock_write_cb(G_GNUC_UNUSED int fd, G_GNUC_UNUSED GIOCondition condition, G_GNUC_UNUSED gpointer user_data);
 static char *bind_unix_socket(char *socket_relative_name, int sock_type, mode_t perms, struct remote_sock_s *remote_sock,
 			      gboolean use_full_attach_path);
-static char *get_sock_path(gboolean use_full_attach_path, char *base_path, size_t desired_len);
+static char *socket_parent_dir(gboolean use_full_attach_path, size_t desired_len);
 /*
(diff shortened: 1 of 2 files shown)


src/conn_sock.c
```diff
@@ -161,21 +161,35 @@ static char *bind_unix_socket(char *socket_relative_name, int sock_type, mode_t
 		get_sock_path(use_full_attach_path, g_build_filename(opt_socket_path, opt_cuuid, NULL), sizeof(socket_addr.sun_path));
 
 	/*
-	 * Instead of creating a symlink, open the desired socket path and use the fd reference
-	 * from /proc/self/fd.
+	 * Create a symlink to match expectations of callers, even though we no longer
+	 * worry about exceeding the unix socket path limit.
```
giuseppe, line 165: could we create the symlink only when `!use_full_attach_path`?

haircommander: we need this symlink regardless, as podman references `socket_path/attach` but the attach file actually lives in `bundle_path/attach`

haircommander: unless you're referring to dropping the sock_path entirely with this PR, and then do so in podman when we use --full-attach

haircommander: done, now callers who use `--full-attach` should not rely on  the socket dir being used (looking at you, podman)

giuseppe (approved):
LGTM

haircommander:
@rhatdan @mheon @saschagrunert PTAL (I packaged a totally broken  conmon and it's on it's way to fedora now :upside_down_face: this PR fixes it)

rhatdan:
LGTM

PR #232, opened 2021-01-15
fix propagation of PREFIX variable to docs/Makefile

### LINKED ISSUE #231: PREFIX is not propagated to docs/Makefile
If i set PREFIX when building, it does not propagate to Makefile for docs, so it installs on default location.
By changing this simple thing, the problem is fixed:

https://github.com/containers/conmon/blob/0331601d5767a3506f5a9c1d03adb44f9b24b94b/docs/Makefile#L1

Change to:
```
PREFIX ?= /usr/local
```

chax:
fixes #231

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -1,4 +1,4 @@
-PREFIX := /usr/local
+PREFIX ?= /usr/local
 DATADIR := ${PREFIX}/share
 MANDIR := $(DATADIR)/man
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
@giuseppe @haircommander PTAL

saschagrunert (approved):
LGTM

PR #233, opened 2021-01-20
conn_sock: fix double free

haircommander:
as well as bump the conmon version

PR #234, opened 2021-01-20
src: avoid unuseful strdup

giuseppe:

### DIFF
--- a/src/conn_sock.c
+++ b/src/conn_sock.c
@@ -76,7 +76,7 @@ char *setup_console_socket(void)
 {
 	struct sockaddr_un addr = {0};
 	_cleanup_free_ const char *tmpdir = g_get_tmp_dir();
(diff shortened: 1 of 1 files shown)


src/conn_sock.c
```diff
@@ -106,7 +106,10 @@ char *setup_console_socket(void)
 	if (listen(console_socket_fd, 128) < 0)
 		pexit("Failed to listen on console-socket");
 
-	return g_strdup(csname);
+	/* Move the ownership out of csname.  */
+	char *tmp = csname;
+	csname = NULL;
+	return tmp;
```
haircommander, line 112: couldn't we also just drop _cleanup_free_ when declaring csname?

giuseppe: yes, good point.  I'll do that

haircommander:
LGTM, I'll merge this then https://github.com/containers/conmon/pull/233 after

PR #239, opened 2021-02-09
Add arm64 static build binary

saschagrunert:
This adds an arm64 derivation to the static binary build.

#### Manual test
```
> nix-build nix/arm64.nix
> file result/bin/conmon
result/bin/conmon: ELF 64-bit LSB executable, ARM aarch64, version 1 (SYSV), statically linked, for GNU/Linux 2.6.32, stripped
```
Test build: [conmon.tar.gz](https://github.com/containers/conmon/files/5951723/conmon.tar.gz)
Refers to https://github.com/cri-o/cri-o/pull/4552

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -141,28 +141,44 @@ static_binary_task:
         memory: 12
         disk: 200
 
-    init_script: |
-        set -ex
-        setenforce 0
-        growpart /dev/sda 1 || true
-        resize2fs /dev/sda1 || true
-        yum -y install podman
+    # Community-maintained task, may fail on occasion.  If so, uncomment
+    # the next line and file an issue with details about the failure.
+    # allow_failures: true
 
-    nix_cache:
-      folder: '.cache'
-      fingerprint_script: |
-        echo "nix-v1-$(sha1sum nix/nixpkgs.json | head -c 40)"
+    timeout_in: '20m'
+
+    env:
+        # Do not use 'latest', fixed-version tag for runtime stability.
+        CTR_FQIN: "docker.io/nixos/nix:2.3.6"
+        # Authentication token for pushing the build cache to cachix.
+        # This is critical, it helps to avoid a very lengthy process of
+        # statically building every dependency needed to build conmon.
+        # Assuming the pinned nix dependencies in nix/nixpkgs.json have not
+        # changed, this cache will ensure that only the static conmon binary is
+        # built.
+        CACHIX_AUTH_TOKEN: ENCRYPTED[4c3b8d82b0333abf048c56a71f2559ddb1c9ed38f0c28916eca13f79affa5904cf90c76a5bd8686680c89f41079ef341]
+
+    alias: static_binary
+
+    matrix:
+      - env:
+            TARGET: default.nix
+      - env:
+            TARGET: default-arm64.nix
 
     build_script: |
         set -ex
-        mkdir -p .cache
-        mv .cache /nix
-        if [[ -z $(ls -A /nix) ]]; then podman run --rm --privileged -ti -v /:/mnt nixos/nix cp -rfT /nix /mnt/nix; fi
-        podman run --rm --privileged -ti -v /nix:/nix -v ${PWD}:${PWD} -w ${PWD} nixos/nix nix --print-build-logs --option cores 8 --option max-jobs 8 build --file nix/
+        podman run -i --rm \
+          -e CACHIX_AUTH_TOKEN \
+          -v $PWD:$PWD:Z \
+          -w $PWD \
+          $CTR_FQIN \
+          sh -c \
+          "nix-env -iA cachix -f https://cachix.org/api/v1/install && \
+           cachix use conmon && \
+           nix-build nix/$TARGET && \
+           nix-store -qR --include-outputs \$(nix-instantiate nix/$TARGET) | grep -v conmon | cachix push conmon && \
+           cp -R result/bin ."
 
     binaries_artifacts:
-        path: "result/bin/conmon"
-
-    save_cache_script: |
-        mv /nix .cache
-        chown -Rf $(whoami) .cache
+        path: "bin/conmon"
--- a/nix/default-arm64.nix
+++ b/nix/default-arm64.nix
@@ -1,0 +1,95 @@
+let
+  pkgs = (import ./nixpkgs.nix {
+    crossSystem = {
+      config = "aarch64-unknown-linux-gnu";
+    };
+    overlays = [
+      (final: pkg: {
+        pcre = (static pkg.pcre).overrideAttrs (x: {
+          configureFlags = x.configureFlags ++ [
+            "--enable-static"
+          ];
+        });
+      })
+    ];
+    config = {
+      packageOverrides = pkg: {
+        autogen = (static pkg.autogen);
+        e2fsprogs = (static pkg.e2fsprogs);
+        libuv = (static pkg.libuv);
+        glib = (static pkg.glib).overrideAttrs (x: {
+          outputs = [ "bin" "out" "dev" ];
+          mesonFlags = [
+            "-Ddefault_library=static"
+            "-Ddevbindir=${placeholder ''dev''}/bin"
+            "-Dgtk_doc=false"
+            "-Dnls=disabled"
+          ];
+          postInstall = ''
+            moveToOutput "share/glib-2.0" "$dev"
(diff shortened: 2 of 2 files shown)


saschagrunert:
The cache is 2131Mb large and the build took ~2 hours. I think if the cache only invalidates when updating the pinned nixpkgs then we're probably fine with that change, right?

PTAL @cevich

cevich:
> cache only invalidates when updating the pinned nixpkgs then we're probably fine with that change, right?

I'm nearly always amazed by how deceptively complex caching can be.  A 2-hour build puts us *right* up there against a hard-coded Cirrus-CI max job runtime (no setting can override it).  Considering our experience on the podman repo, would it make sense to go straight to the cachix solution here as well?

.cirrus.yml
```diff
@@ -166,3 +166,41 @@ static_binary_task:
...
+# Build the static binary for arm64
+static_binary_arm64_task:
+    depends_on:
+        - 'config'
+        - 'fmt'
+
+    gce_instance:
+        image_name: "${FEDORA_CACHE_IMAGE_NAME}"
+        cpu: 8
+        memory: 12
```
cevich, line 179: Ouch, this could get expensive for us if it runs multiple times per PR and fails after 2-hours.  Would it make sense/be safer to limit this task to only branches?  Another idea could be to add a `timeout_in: 30m` attribute (or similar).  What do you think?

.cirrus.yml
```diff
@@ -166,3 +166,41 @@ static_binary_task:
     save_cache_script: |
         mv /nix .cache
         chown -Rf $(whoami) .cache
+
+# Build the static binary for arm64
+static_binary_arm64_task:
+    depends_on:
```
cevich, line 172: If this task fails, it will probably happen for some critical "must merge" PR, and while you or I aren't around. 
 Would you mind adding an "escape hatch" comment [similar to what we have in podman](https://github.com/containers/podman/blob/6a6e86829071d20a7f7fc30d7346c95a6fd3927a/.cirrus.yml#L330-L332)?

.cirrus.yml
```diff
@@ -166,3 +166,41 @@ static_binary_task:
...
+        set -ex
+        setenforce 0
+        growpart /dev/sda 1 || true
+        resize2fs /dev/sda1 || true
+        yum -y install podman
+
+    nix_arm64_cache:
+      folder: '.cache'
+      fingerprint_script: |
+        echo "nix-v1-$(sha1sum nix/nixpkgs.json | head -c 40)"
```
cevich, line 192: Would you mind moving this section to the end, right after `save_cache_script`?  I think it might be easier to read that way.

.cirrus.yml
```diff
@@ -166,3 +166,41 @@ static_binary_task:
...
+        cpu: 8
+        memory: 12
+        disk: 200
+
+    init_script: |
+        set -ex
+        setenforce 0
+        growpart /dev/sda 1 || true
+        resize2fs /dev/sda1 || true
+        yum -y install podman
```
cevich, line 187: I'm not in favor of run-time installs, they tend to cause ugly (hard to solve) automation-reliability problems.  In any case, the podman package should already be installed (along with buildah and skopeo) in every VM image.  That will never change, we have to install it to make sure all default dependencies are present.  Did you find this not to be the case?

.cirrus.yml
```diff
@@ -166,3 +166,41 @@ static_binary_task:
...
+      folder: '.cache'
+      fingerprint_script: |
+        echo "nix-v1-$(sha1sum nix/nixpkgs.json | head -c 40)"
+
+    build_script: |
+        set -ex
+        mkdir -p .cache
+        mv .cache /nix
+        if [[ -z $(ls -A /nix) ]]; then podman run --rm --privileged -ti -v /:/mnt nixos/nix cp -rfT /nix /mnt/nix; fi
+        podman run --rm --privileged -ti -v /nix:/nix -v ${PWD}:${PWD} -w ${PWD} nixos/nix nix --print-build-logs --option cores 8 --option max-jobs 8 build --file nix/default-arm64.nix
```
cevich, line 199: It just occurred to me, this will hit the docker-hub, and so could fail randomly (do we do this in podman?).  Would it make sense to point at (or configure) `mirror.gcr.io` proxy for the image?

cevich (commented):
Good first-draft, I made a few comments/questions for discussion.  Bonous points for getting it working on your own!

.cirrus.yml
saschagrunert: No I just took that part from the existing static build. I'll clean that up after this PR got merged. :+1:

.cirrus.yml
saschagrunert: I think we do not need a cache if we go via cachix. :)

.cirrus.yml
saschagrunert: Yes, I'll added this as suggested.

.cirrus.yml
saschagrunert: Added a 20 minute timeout. With using cachix this should be totally enough.

saschagrunert:
> Considering our experience on the podman repo, would it make sense to go straight to the cachix solution here as well?

Yeah I think using cachix is a good move forward here. I'll pre-populate the [conmon cache](https://app.cachix.org/cache/conmon) and come back to the CI if that is done. :+1:

.cirrus.yml
saschagrunert: Yes, I think this makes sense. Added it as suggested.

saschagrunert: I tried it but it looks like that the mirror does not cache the nix image: https://cirrus-ci.com/task/5561064597225472

saschagrunert:
The job is down to 3 minutes with the cache, and the binary artifacts looks fine, too:
```
> file conmon
conmon: ELF 64-bit LSB executable, ARM aarch64, version 1 (SYSV), statically linked, for GNU/Linux 2.6.32, stripped
```

.cirrus.yml
cevich: Hmmm, dang.  Somehow we've managed to not have problems (that I've observed) on podman, so it's probably okay for now.  Esp. on this low-traffic repo.

cevich:
Great!  Thanks for all your work on this.  Do you want to e-mail me so I can create an encrypted `$CACHIX_AUTH_TOKEN`?

(apologies if you already did and I didn't see it yet for some reason)

saschagrunert:
> (apologies if you already did and I didn't see it yet for some reason)

Yes, I already did it :)

cevich:
Well go figure, now that I go looking for it...there it is! :rofl: 

Use `ENCRYPTED[4c3b8d82b0333abf048c56a71f2559ddb1c9ed38f0c28916eca13f79affa5904cf90c76a5bd8686680c89f41079ef341]`

cevich:
It looks to me like there's a lot of duplication between `static_binary_task` and `static_binary_arm64_task`.  Just by eyeball, is the only difference `default.nix` vs `default-arm64.nix`?

Assuming so, we can easily simplify this to one task, using an env. var. and a matrix...

saschagrunert:
> It looks to me like there's a lot of duplication between `static_binary_task` and `static_binary_arm64_task`. Just by eyeball, is the only difference `default.nix` vs `default-arm64.nix`?
> 
> Assuming so, we can easily simplify this to one task, using an env. var. and a matrix...

Alright changed to a matrix, let's see how that works.

cevich:
Lol, I was just about to post a proposed diff but you posted almost the exact same change :smile: 

I think the only secret-sauce (it's not obvious) in the task is `alias: static_binary` (that will cause the `name:` value to work properly)

saschagrunert:
> Lol, I was just about to post a proposed diff but you posted almost the exact same change 
> 
> I think the only secret-sauce (it's not obvious) in the task is `alias: static_binary` (that will cause the `name:` value to work properly)

Ah thank you for the hint, are those names fine now?

TomSweeneyRedHat:
I'm not at all an expert in this arena, but the changes LGTM.  
@cevich and @saschagrunert thank you both for your great work here!

cevich:
> Ah thank you for the hint, are those names fine now?

Hmmm, something is still wrong...oh, you removed the 'name' attr. from matrix, that's why.  No matter, I think it's probably fine as-is...it's a small detail not worth fretting over.  Important thing is you got it all working :grinning: 

/lgtm

cevich:
Woops...no bot here.

Issue #240, opened 2021-02-23
podman restart with port bindings results in 'port in use' errors


baude:
Sometimes when restarting a container with podman restart, it results in a `port already in use` error.  It seems like you need to have an application in the container that takes a little longer to execute to trip it more reliably.

Reproducer

```
$ sudo podman run -dt --name foo -p 80:80 alpine_nginx
b7fc4ef1bc8a130ec0dec234c59bc11a7d6e88b24e48d04e8afede3b832db3db
$ sudo podman restart foo
Error: cannot listen on the TCP port: listen tcp4 :80: bind: address already in use
```

split from containers/podman#9465

PR #241, opened 2021-02-24
close_fds: fix close of external fds

giuseppe:
commit 0f092d5446fdadb761b8143b5c4b1d1ac9793f29 introduced the
regression.

Closes: https://github.com/containers/conmon/issues/240

### DIFF
--- a/src/close_fds.c
+++ b/src/close_fds.c
@@ -32,7 +32,6 @@ static void __attribute__((constructor)) init()
 	ssize_t size = 0;
 	DIR *d;
 
-	/* Store how many FDs were open before the Go runtime kicked in.  */
 	d = opendir("/proc/self/fd");
 	if (!d)
 		return;
@@ -74,10 +73,10 @@ void close_other_fds()
 {
 	int fd;
 
-	for (fd = 3; fd < open_files_max_fd; fd++) {
-		if (open_files_set == NULL || FD_ISSET(fd % FD_SETSIZE, &(open_files_set[fd / FD_SETSIZE])))
-			if (fd == sync_pipe_fd || fd == attach_pipe_fd || fd == dev_null_r || fd == dev_null_w || fd == oom_cgroup_fd
-			    || fd == oom_event_fd)
-				close(fd);
+	if (open_files_set == NULL)
+		return;
+	for (fd = 3; fd <= open_files_max_fd; fd++) {
+		if (FD_ISSET(fd % FD_SETSIZE, &(open_files_set[fd / FD_SETSIZE])))
+			close(fd);
 	}
 }


giuseppe:
CC @baude

src/close_fds.c
```diff
@@ -74,10 +73,8 @@ void close_other_fds()
 {
 	int fd;
 
-	for (fd = 3; fd < open_files_max_fd; fd++) {
+	for (fd = 3; fd <= open_files_max_fd; fd++) {
 		if (open_files_set == NULL || FD_ISSET(fd % FD_SETSIZE, &(open_files_set[fd / FD_SETSIZE])))
-			if (fd == sync_pipe_fd || fd == attach_pipe_fd || fd == dev_null_r || fd == dev_null_w || fd == oom_cgroup_fd
-			    || fd == oom_event_fd)
-				close(fd);
+			close(fd);
```
haircommander, line 78: do we change the name "close_other_fds" if we're closing all the fds?

giuseppe: we are closing only the fds that were open when the program started so it is still the "other fds" that weren't opened by conmon itself

haircommander: I am not sure I understand how this fixes the problem. I can see the `<=` being helpful (we're not missing one of the fds). but why do we no longer ignore all of the fds conmon itself opened. Those are closed in different sections of the code AFAIK, and we'd double close them if we close them here, right?

giuseppe: we were previously closing the fd if it was one of `sync_pipe_fd`, `attach_pipe_fd`, `dev_null_r`, `dev_null_w`, `oom_cgroup_fd`, `oom_event_fd` and that never triggered because these fds were open after the program started up.  The logic was broken since the beginning, although it was enough to fix the 100% cpu usage bug we had earlier

haircommander:
LGTM, @giuseppe did you test against podman unit/system tests? since we don't run them on CI it has to be manually done (for now..)

giuseppe:
> LGTM, @giuseppe did you test against podman unit/system tests? since we don't run them on CI it has to be manually done (for now..)

I've ran the e2e tests and they pass locally

src/close_fds.c
```diff
@@ -74,10 +73,8 @@ void close_other_fds()
 {
 	int fd;
 
-	for (fd = 3; fd < open_files_max_fd; fd++) {
+	for (fd = 3; fd <= open_files_max_fd; fd++) {
 		if (open_files_set == NULL || FD_ISSET(fd % FD_SETSIZE, &(open_files_set[fd / FD_SETSIZE])))
```
haircommander, line 77: sorry, one more quick nit, can we return from this function early if `open_files_set == NULL`?

giuseppe: good idea!

Pushed new version

haircommander:
thanks @giuseppe , I assume we want a release for this @baude ?

PR #246, opened 2021-03-08
exec: don't fail on EBADFD

haircommander:
in legacy exec, we send the PID back to the parent through the sync pipe fd
we can't close it before we do so.


haircommander:
@giuseppe PTAL

giuseppe (approved):
LGTM

PR #248, opened 2021-03-09
Add Podman integration test GitHub action

saschagrunert:
We now add the Podman integration tests side-by-side to the CRI-O run.
This way we can ensure that changes to conmon will not negatively impact
Podman.

Some tests have to be ecluded for now, we have to dig deeper to find out
why they do not run.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -5,7 +5,6 @@ on:
       - v*
     branches:
       - master
-      - ghactions # TODO: remove when it works
   pull_request:
 jobs:
   cri-o:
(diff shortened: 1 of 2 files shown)


vrothberg (commented):
I would prefer to exercise the system tests in `test/system`. They are used in the gating tests in all kinds of distributions _and_ they pass the eagle eyes of @edsantiago.

haircommander:
> I would prefer to exercise the system tests in `test/system`. They are used in the gating tests in all kinds of distributions _and_ they pass the eagle eyes of @edsantiago.

why not both :D

rhatdan:
LGTM

saschagrunert:
> > I would prefer to exercise the system tests in `test/system`. They are used in the gating tests in all kinds of distributions _and_ they pass the eagle eyes of @edsantiago.
> 
> why not both :D

We now have both working :)

vrothberg (commented):
LGTM, thank you!

rhatdan:
LGTM

PR #254, opened 2021-04-01
Use less resources

fkorotkov:
It seems these tasks are not using more then 1 CPU


fkorotkov:
See 
image

haircommander:
LGTM, @cevich PTAL

rhatdan:
LGTM

cevich:
Yep this is fine.

PR #263, opened 2021-05-27
Reset OOM score back to 0 for container runtime

mrunalp:
We don't want container runtime procesess to be unkillable
so we reset oom_score_adj back to 0 before execv
of the runtime process.

### DIFF
--- a/src/oom.c
+++ b/src/oom.c
@@ -5,16 +5,14 @@
 #include <string.h>
 #include <unistd.h>
 
-#define OOM_SCORE "-1000"
-
-void attempt_oom_adjust()
+void attempt_oom_adjust(const char *const oom_score)
 {
 	int oom_score_fd = open("/proc/self/oom_score_adj", O_WRONLY);
 	if (oom_score_fd < 0) {
 		ndebugf("failed to open /proc/self/oom_score_adj: %s\n", strerror(errno));
 		return;
 	}
-	if (write(oom_score_fd, OOM_SCORE, strlen(OOM_SCORE)) < 0) {
+	if (write(oom_score_fd, oom_score, strlen(oom_score)) < 0) {
 		ndebugf("failed to write to /proc/self/oom_score_adj: %s\n", strerror(errno));
 	}
 	close(oom_score_fd);
--- a/src/conmon.c
+++ b/src/conmon.c
@@ -41,7 +41,7 @@ int main(int argc, char *argv[])
 
 	process_cli();
 
-	attempt_oom_adjust();
+	attempt_oom_adjust("-1000");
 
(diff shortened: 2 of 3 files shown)


mrunalp:
@giuseppe @haircommander ptal

rhatdan:
LGTM

saschagrunert (approved):

giuseppe (approved):
LGTM

giuseppe:
the container runtime exits immediately.  In what cases does it turn to be an issue?

There are some formatting issues with `make fmt`

src/oom.c
```diff
@@ -5,16 +5,14 @@
 #include <string.h>
 #include <unistd.h>
 
-#define OOM_SCORE "-1000"
-
-void attempt_oom_adjust()
+void attempt_oom_adjust(char* oom_score)
```
haircommander, line 8: nit can we do `const char* const` here?

haircommander: I think there may be a line change needed in `src/oom.h` as well

mrunalp: done

kolyshkin:
> the container runtime exits immediately. In what cases does it turn to be an issue?

@giuseppe I believe this is for cases when `runc init` is waiting on the exec_fifo to continue. The only processes in cgroup are conmon and runc, both are unkillable (oom_score_adj == -1000), and the memory limit is hit, so the kernel would do an OOM kill but it can't (and there's not much memory in this cgroup to reclaim either). As a result of all this, the kernel hits code path(s) which is/are rarely hit and buggy, and bad things happen (now I am not sure what "bad things" are, perhaps this is high CPU usage by the kernel trying to find pages to reclaim, or something else entirely).

So basically this is a kludge against hitting that code path and thus those kernel bugs.

One other thing is, the whole container can inherit oom score of -1000 unless OomScoreAdj is set in its config.

giuseppe:
> @giuseppe I believe this is for cases when `runc init` is waiting on the exec_fifo to continue. The only processes in cgroup are conmon and runc, both are unkillable (oom_score_adj == -1000), and the memory limit is hit, so the kernel would do an OOM kill but it can't (and there's not much memory in this cgroup to reclaim either). As a result of all this, the kernel hits code path(s) which is/are rarely hit and buggy, and bad things happen (now I am not sure what "bad things" are, perhaps this is high CPU usage by the kernel trying to find pages to reclaim, or something else entirely).
> 
> So basically this is a kludge against hitting that code path and thus those kernel bugs.
> 
> One other thing is, the whole container can inherit oom score of -1000 unless OomScoreAdj is set in its config.

thanks for the explanation.  That makes sense.

LGTM

haircommander:
LGTM

Issue #262, opened 2021-05-27
An Analysis of Conmon Container Logging Behaviors with Recommendations


portante:
This is a detailed analysis of the logging behavior in conmon as of tag `v2.0.28`.

Some recommendations:

 * Match read size to actual `pipe(7)` size (64 KB by default on Linux)
 * Align read buffer on page boundary, don't use leading byte just for remote consoles
 * Consider reading from `pipe(7)` until `EWOULDBLOCK`
 * Raise IO vector count to 1,024 to reduce number of `writev(2)` system calls per buffer `read(2)`
   * OR implement https://github.com/cri-o/cri-o/pull/1605 to avoid pathological behaviors

```
An Analysis of Conmon Container Logging Behaviors
Peter Portante, May 2021
Based on the v2.0.28 (GitHub tag) code base

1. Conmon uses non-blocking pipes for reading from the stdout & stderr pipes
   of a container

   a. Pipes created at lines 165 [3] and 180 [4] of conmon.c using the pipe2()
      [5] system call, with the O_CLOEXEC flag

      i. Pipes are 64 KB in size, the Linux default

      ii. Does Go change the size of the pipes dynamically?

   b. Registers stdio_cb(), line 17 [1] of ctr_stdio.c for poll() events

2. In response to being invoked, the stdio_cb() in turn calls read_stdio()
   [2], which reads from the pipe file descriptor using STDIO_BUF_SIZE sized
   buffer (+2 as described at [7])

   a. Where STDIO_BUF_SIZE is 8 KB [6], for a total of 8194 bytes per read

   b. [Flag] This means that 8 read system calls are required to drain the
      pipe

      i. [Question] Does the Go integration change the pipe size to 8,192
         bytes?

      ii. The read_stdio() only makes one read(2) system call

          * [Question] Why is only one read(2) system call made?  Does that
            allow the polling loop to fairly round-robin between reading from
            stdout and stderr?

          * [Flag] We should consider having the read_stdio() continue reading
            until read(2) returns EWOULDBLOCK

      ii. [Flag] The STDIO_BUF_SIZE should be 64 KB to allow the most
          efficient use of system calls, and we should even consider
          increasing the size to the maximum allowed on the system further
          increase the efficiency if we can implement
          https://github.com/cri-o/cri-o/pull/1605.

   c. The read buffer is a stack local variable in read_stdio(), but the
      target address used in the read(2) system call is moved by one byte [7],
[...]

PR #264, opened 2021-05-31
Adjustments to container logging to address concerns from #262

portante:
There are two commits in this PR:

----

Move `WRITEV_BUFFER_N_IOV` to `src/ctr_logging.c`

The `WRITEV_BUFFER_N_IOV` constant is not part of the `ctr_logging` interface, and is only used internal to `src/ctr_logging.c`.

----

Match `read(2)` buffer size to pipe size.

Address a number of conditions flagged in issue #262.

The general thrust of this commit is to move closer to a one-to-one, `read(2)` system call to `writev(2)` system call ratio by creating a memory buffer for reading sized to match the stdio pipes.

The commit also allocates a maximum number of I/O vectors sized to the read buffer size, calculating 2 I/O vectors per line, defaulting to a target line size of 25 bytes (defined as the constant, `AVG_LOG_LINE_TARGET`).

The commit also now allocates the I/O vectors and read buffer memory using `mmap()` so that we don't cross page boundaries for I/O operations.

As a result, the commit removes the `STDIO_BUF_SIZE` constance, using `fcntl(F_GETPIPE_SZ)` calls to size all buffers for reading from pipes.

The `stdpipe_t` value is now passed to `write_to_remote_consoles()` so that we do not have to play games with pre-/post- byte allocations to buffers.

### DIFF
--- a/cmd/conmon-config/conmon-config.go
+++ b/cmd/conmon-config/conmon-config.go
@@ -9,12 +9,10 @@ import (
 )
 
 func main() {
-	output := `
-#if !defined(CONFIG_H)
+	output := `#if !defined(CONFIG_H)
 #define CONFIG_H
 
 #define BUF_SIZE %d
-#define STDIO_BUF_SIZE %d
 #define CONN_SOCK_BUF_SIZE %d
 #define DEFAULT_SOCKET_PATH "%s"
 #define WIN_RESIZE_EVENT %d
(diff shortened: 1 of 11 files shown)


haircommander:
I opened https://github.com/cri-o/cri-o/pull/5013 to test these changes with the kubernetes e2e and cri-o

rhatdan:
Seems to be some packaging issues:
/usr/bin/ld: src/ctr_logging.o: in function `writev_buffer_init':
ctr_logging.c:(.text+0xd5a): undefined reference to `ceilf'
/usr/bin/ld: ctr_logging.c:(.text+0xdd9): undefined reference to `ceilf'

Also fmt test is not happy.

PR #265, opened 2021-05-31
Implement proposed `k8s-stream-file` format

portante:
Instead of parsing the contents of the data read from the `stdout` and `stderr` pipes, this commit adds support for a "stream" format, named `k8s-stream-file`, which just records what is read from a pipe to disk.

It significantly saves on CPU spend processing the buffer read, uses only 2 I/O vectors, and never touches the memory read from the pipe.

This is an updated implementation of https://github.com/cri-o/cri-o/pull/1605.


portante:
See also #262.

src/ctr_logging.c
```diff
@@ -134,9 +137,10 @@ void configure_log_drivers(gchar **log_drivers, int64_t log_size_max_, char *cuu
  * parse_log_path branches on log driver type the user inputted.
  * log_config will either be a ':' delimited string containing:
  * <DRIVER_NAME>:<PATH_NAME> or <PATH_NAME>
- * in the case of no colon, the driver will be kubernetes-log-file,
+ * in the case the log driver is 'k8s-stream-file', the <PATH_NAME> must be present.
+ * in the case of no colon, the driver will be k8s-file,
  * in the case the log driver is 'journald', the <PATH_NAME> is ignored.
- * exits with error if <DRIVER_NAME> isn't 'journald' or 'kubernetes-log-file'
+ * exits with error if <DRIVER_NAME> isn't 'journald' or 'k8s-file'
```
haircommander, line 143: nit, we should probably add `k8s-stream-file` to this list

src/ctr_logging.c
```diff
@@ -165,6 +169,17 @@ static void parse_log_path(char *log_config)
 		return;
 	}
 
+	// Driver is k8s-file or empty
```
haircommander, line 172: nit, this comment doesn't seem correct

src/ctr_logging.c
```diff
@@ -540,3 +565,108 @@ void sync_logs(void)
...
 }
+
+
+/* strlen("1997-03-25T13:20:42.999999999+01:00 stdout 9999999999 999999999 ") + 1 */
+#define TSSTREAMBUFLEN 128
+
+/*
+ * PROPOSED: CRI Stream Format, variable length file format
+ */
+static int set_k8s_stream_timestamp(char *buf, ssize_t bufsiz, ssize_t *tsbuflen, const char *pipename, uint64_t offset, ssize_t buflen, ssize_t *btbw)
```
haircommander, line 576: there seem to be some similarities between this and set_k8s_timestamp. can we either make the shared functionality a function or document why that's not possible?

haircommander (commented):
Thanks for the implementation @portante ! I have some fiirst-pass review comments, but if this is just a POC feel free to ignore them until the concept is proven.

you'll need to run `make fmt` and commit the changes to have the majority of CI checks run.

I am betting this will perpetually fail kubernetes e2e tests but I am curious to see what actually ends up happening.

src/ctr_logging.c
portante: Fixed.

src/ctr_logging.c
portante: I did not want to disturb any existing code with this to avoid introducing un-intended bugs.

If we decide that it is worth merging this, then we should consider combing those two methods into one.

portante:
> you'll need to run make fmt and commit the changes to have the majority of CI checks run.

Fixed.

Thanks for the review!

rhatdan:
@giuseppe PTAL

src/ctr_logging.c
```diff
@@ -540,3 +564,110 @@ void sync_logs(void)
...
+	if (localtime_r(&ts.tv_sec, &current_tm) == NULL)
+		return err;
+
+	off = (int)current_tm.tm_gmtoff;
+	if (current_tm.tm_gmtoff < 0) {
+		off_sign = '-';
+		off = -off;
+	}
+
+	len = snprintf(buf, bufsiz, "%d-%02d-%02dT%02d:%02d:%02d.%09ld%c%02d:%02d %s %lud %ld ", current_tm.tm_year + 1900,
```
mtrmac, line 601: (A drive-by comment with little context:) 

If the goal is to offload the CPU processing to consumers, shouldn’t this just be a timestamp instead of all the timezone lookups and formatting?

Either way, UTC would be better than ambiguous local time — does this one need custom parser code to get a `struct timespec` back?

portante:
> @portante regarding this PR - maybe UTC format would be better as @mtrmac mentions?

I agree that UTC would be better.  Let me see about that.

src/ctr_logging.c
portante: The problem here is establishing a reference point for what the timestamp "means".

First, for efficiency, a simple monotonically incremental timestamp is all that is needed, along with a periodic mapping of that timestamp to a real clock.  This format could be modified to periodically emit (once a second?  once every 5 seconds?) an entry for that mapping: <monotonic stamp, realtime stamp>

That realtime stamp would be emitted using UTC instead of the local timezone.  All logs after would calculate the real timestamp from that that offset.

If that makes sense, we can implement this instead.

portante: I should add that the goal is to push the details of interpreting the log stream to the reader and not on `conmon` the writer.

Issue #271, opened 2021-06-06
[logging] allow custom message format


tobwen:
## preface
I'm using _conmon_  with _podman_. Since some containerized applications don't have good logging support, I'm using _journald_ to capture the logs from _conmon_. For longtime storage, I'm processing the logs using _rsyslog_.

## What's my issue?
Although _conmon_ exported the container name and tags into _journld_, these details aren't part of the _syslog_ export. It's impossible to filter out by container name. So _rsyslog_ has to use the _imjournal_ module, which makes all the tags available - but with way trouble in performance.

```
Jun  6 20:00:00 debian conmon[12975]: 2021-06-06 20:00:00.000 UTC [2] LOG:  database system is shut down
````

## What might be a possible solution?
Allow configuring the _syslog_ message.

```
Jun  6 20:00:00 debian conmon[12975]: $CONTAINER-NAME 2021-06-06 20:00:00.000 UTC [2] LOG:  database system is shut down
```

or 

```
Jun  6 20:00:00 debian conmon-$CONTAINER-NAME[12975]: 2021-06-06 20:00:00.000 UTC [2] LOG:  database system is shut down
```

etc.

**Comment by tobwen** [comment]
Easy patch, which might work for me:
https://github.com/containers/conmon/blob/24c73c2abb3f220ebd97acf498b57c560ca26c04/src/ctr_logging.c#L236-L238

``` C
         _cleanup_free_ char *message = g_strdup_printf("MESSAGE=%s %s", name, buf);
         if (writev_buffer_append_segment(dev_null, &bufv, message, line_len + name_len + 1 + MESSAGE_EQ_LEN) < 0)
                return -1;

```

portante:
In general, it will only create performance problems for `conmon` to spend time formatting log output.  See #262 for details on the impacts that happen when a subsystem is involved in interpreting the byte stream.

Instead, moving to a raw format, as proposed in #265, and moving to a log reader of that format that transforms it for a target log receiver would be the most sustainable approach.

PR #273, opened 2021-06-07
include name and tag of a container into journald message for syslog [WIP]

tobwen:
# summary
When writing logs from _conmon_ into _journald_ and post-process the journal with _rsyslog_ or others, there's no way to filter logs from plain _syslog_ messages. This PR prefixes the container's name and its tags (if set) to each message. See: https://github.com/containers/conmon/issues/271

# syntax
`container-name[container-tag]`

# examples
## container-name has been set, container-tag has been set
`Jun  7 00:00:00 debian conmon[666]: podman-nginx[test-buid-20210706] message`

## only container-name has been set
`Jun  7 00:00:00 debian conmon[667]: podman-nginx It works.`

## only container-tag has been set
`Jun  7 00:00:00 debian conmon[668]: [test-buid-20210706] It works.`

## neither ... nor ... has been set
`Jun  7 00:00:00 debian conmon[669]: It works.`

### DIFF
--- a/src/ctr_logging.c
+++ b/src/ctr_logging.c
@@ -47,6 +47,7 @@ static char *k8s_log_path = NULL;
 static char short_cuuid[TRUNC_ID_LEN + 1];
 static char *cuuid = NULL;
 static char *name = NULL;
+static char *msg_prefix = "";
 static size_t cuuid_len = 0;
 static size_t name_len = 0;
 static char *container_id_full = NULL;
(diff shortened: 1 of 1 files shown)


haircommander:
thanks for the PR @tobwen LGTM

haircommander:
one thing of note: I wonder if this break compatibilty with the docker journald log driver (on which this behavior was based)

tobwen:
> one thing of note: I wonder if this break compatibilty with the docker journald log driver (on which this behavior was based)

True. I can try to add a switch. But I'd also need to make changes to _podman_ then (`--log-opt`).

haircommander:
> > one thing of note: I wonder if this break compatibilty with the docker journald log driver (on which this behavior was based)
> 
> True. I can try to add a switch. But I'd also need to make changes to _podman_ then (`--log-opt`).

yeah I think that'd be best

tobwen:
Implementation is too complicated and the audience is too little. Closing.

PR #281, opened 2021-07-29
Update nixpkgs

saschagrunert:
Update to the latest packages and refactor some derivation parts.


saschagrunert:
Both static builds are green. PTAL @haircommander

PR #283, opened 2021-08-12
Remove unreachable code path

saschagrunert:
`winsz_fd_r` is always `-1` in this code path, because it will only set
in `setup_console_fifo()` which gets called a few lines below.

PTAL @haircommander

### DIFF
--- a/src/conmon.c
+++ b/src/conmon.c
@@ -169,12 +169,6 @@ int main(int argc, char *argv[])
 
 		mainfd_stdout = fds[0];
 		workerfd_stdout = fds[1];
-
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
@haircommander @mrunalp PTAL

haircommander:
it is consistently a wonder to me conmon works...

I am not sure we can just drop this call, I think we need to do https://github.com/containers/conmon/pull/284 instead

haircommander:
well, now I'm not so sure. I can't tell if we will actually need window resizing if we're !terminal....

giuseppe (approved):
LGTM

haircommander:
LGTM

PR #284, opened 2021-08-12
correct winsz registration

haircommander:
when a container is not terminal, the ctrl_winsz_cb was never correctly registered,
as there was an ordering problem with the registration function. Fix this

### DIFF
--- a/src/conmon.c
+++ b/src/conmon.c
@@ -169,12 +169,6 @@ int main(int argc, char *argv[])
 
 		mainfd_stdout = fds[0];
 		workerfd_stdout = fds[1];
-
-		/* now that we've set mainfd_stdout, we can register the ctrl_winsz_cb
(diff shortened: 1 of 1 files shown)


haircommander:
alternative to https://github.com/containers/conmon/pull/283

Though, it's not totally clear to me that we even need to register this. Will we have window resizing if we're not running with a terminal?

rhatdan:
@giuseppe @saschagrunert PTAL

giuseppe (approved):
LGTM

giuseppe:
code LGTM although I don't think we need this code if it is not a terminal.  I'd rather go with #283

saschagrunert:
> Will we have window resizing if we're not running with a terminal?

Logically not.

haircommander:
yeah let's go for that

PR #289, opened 2021-09-01
logging: new mode -l passthrough

giuseppe:
when passthrough is specified, the std stream files are passed directly to the container without conmon intercepting them.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -42,7 +42,8 @@ jobs:
       - name: Run conmon integration tests
         run: |
           make vendor
-          sudo make test
+          sudo mkdir -p /var/run/crio
+          sudo make test-binary
 
   cri-o:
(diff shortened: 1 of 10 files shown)


giuseppe:
@haircommander PTAL

rhatdan:
LGTM
 But it would be nice if more tests were passing.

giuseppe:
I see the following error in the tests: `<string>: [conmon:e]: No such log driver passthrough`.

Are we using the correct binary for the CI? :-)

giuseppe:
> I see the following error in the tests: `<string>: [conmon:e]: No such log driver passthrough`.
> 
> Are we using the correct binary for the CI? :-)

I'll take a look at what is going on

giuseppe:
rebased.

@haircommander does it look good to you?

haircommander:
LGTM, thanks!

Issue #290, opened 2021-04-12
podman-remote: rootless: exec /no/such/command makes container become unusable


edsantiago:
I have a feeling this is conmon or crun or something non-podman.

Setup: with `podman system service` running as rootless:
```console
$ podman-remote run --name foo -d quay.io/libpod/testimage:20210223 sh -c 'touch /foo;while [ -f /foo ]; do sleep 1;done'
51f48c694e4a2a8b580b6cb3ccef93ad47c1c79ef84722fde11fa7cfae2f5091
$ podman-remote exec foo /no/such/command
Error: executable file `/no/such/command` not found in $PATH: No such file or directory: OCI not found
ERRO[0592] error attaching to container 51f48c694e4a2a8b580b6cb3ccef93ad47c1c79ef84722fde11fa7cfae2f5091 exec session a47b26da81dce1122495741a3b4188230b908d64afb068be590c7eae20bc3e7c: executable file `/no/such/command` not found in $PATH: No such file or directory: OCI not found
$ podman-remote exec foo true
```
So far, so good. But now, trying to make the container exit:
```
$ podman-remote exec foo rm /foo
$ podman-remote ps
CONTAINER ID  IMAGE                              COMMAND               CREATED             STATUS                 PORTS   NAMES
51f48c694e4a  quay.io/libpod/testimage:20210223  sh -c touch /foo;...  About a minute ago  Up About a minute ago          foo
```
This should not happen. The container should be in Exited(0) state. Furthermore:
```
$ podman-remote exec foo true
Error: fail startup: OCI runtime error
ERRO[0609] error attaching to container 51f48c694e4a2a8b580b6cb3ccef93ad47c1c79ef84722fde11fa7cfae2f5091 exec session d9c365a6d44d619031401f02e4115436edd72356f3a25bfd1ddea17092c33231: fail startup: OCI runtime error
```
100% reproducible. Rootless only.

5.11.9-200.fc33
podman-3.1.0-3.fc33
crun-0.19-1.fc33
conmon-2.0.27-2.fc33
containers-common-1-10.fc33

dezza:
https://github.com/containers/podman/issues/22274

PR #291, opened 2021-09-09
exit: fix race zombie reaper

giuseppe:
fix a race condition where all the processes spawned by the OCI
runtime are not terminated yet and reap_children() would not clean
them up.

If an --exit-delay is specified, these proceses would be left
uncleaned until the delay expires.

When used for an exec session, the zombie process would prevent the
container namespaces to be cleaned up.

Closes: https://github.com/containers/conmon/issues/290

### DIFF
--- a/src/ctr_exit.c
+++ b/src/ctr_exit.c
@@ -160,7 +160,7 @@ void do_exit_command()
 			exit_status = 0;
 			ret = waitpid(-1, &tmp, 0);
 			if (ret == exit_pid)
-				exit_status = tmp;
+				exit_status = get_exit_status(tmp);
 		} while ((ret < 0 && errno == EINTR) || ret > 0);
 
 		if (exit_status)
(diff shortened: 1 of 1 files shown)


vrothberg (commented):
LGTM, great work!

src/ctr_exit.c
```diff
@@ -156,9 +150,18 @@ void do_exit_command()
 	if (exit_pid) {
 		int ret, exit_status = 0;
 
-		do
-			ret = waitpid(exit_pid, &exit_status, 0);
-		while (ret < 0 && errno == EINTR);
+		/* even though we've already registered reap_children,
+		 * atexit() runs functions in reverse, so we need to
+		 * manually call here. Repeated calls will not cause trouble
+		 */
```
haircommander, line 156: I think this comment needs updating because we're not calling reap_children here

rhatdan:
LGTM

lsm5:
@giuseppe @haircommander mind cutting a new version with this? /cc @edsantiago

edsantiago:
I'll second this. We are starting to test podman-remote in gating, and gating tests are guaranteed to fail with old conmon.

haircommander:
opened https://github.com/containers/conmon/pull/296

lsm5:
@haircommander could you push the new tag as well?

Also, I see the dependabot security alert for an older buildah in there? Is that something worth handling before we build the new tag?

Issue #293, opened 2021-08-19
Allow For Container Name To Be Shown In Journald 


p0da:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind feature

**Description**

Would it be possible to add a feature that prefixes journald log messages with the relevant container/tag information? It looks like this exact feature was started in: https://github.com/containers/conmon/pull/273 but never got finished. Thanks.

haircommander:
ah I see, I misunderstood the ask. I have opened https://github.com/containers/conmon/pull/294 to fix. can someone test and let me know if it satisfies their use case?

p0da:
Just updated, feature is working perfectly on my end. Thank you for pushing this feature through @haircommander. Feel free to close this issue as needed.

PR #294, opened 2021-09-16
ctr_logs: use container name or ID as SYSLOG_IDENTIFIER for journald

haircommander:
now, when journald is used for the log driver, the logs will appear as:
ctrName[PID]:...
or
partialID[PID]:... (if using a sufficiently old podman)

instead of
conmon[PID]

### DIFF
--- a/src/ctr_logging.c
+++ b/src/ctr_logging.c
@@ -38,13 +38,22 @@ static int k8s_log_fd = -1;
 static char *k8s_log_path = NULL;
 
 /* journald log file parameters */
+// short ID length
 #define TRUNC_ID_LEN 12
+// MESSAGE=
 #define MESSAGE_EQ_LEN 8
+// PRIORITY=x
 #define PRIORITY_EQ_LEN 10
+// CONTAINER_ID_FULL=
 #define CID_FULL_EQ_LEN 18
+// CONTAINER_ID=
 #define CID_EQ_LEN 13
+// CONTAINER_NAME=
 #define NAME_EQ_LEN 15
+// CONTAINER_PARTIAL_MESSAGE=true
 #define PARTIAL_MESSAGE_EQ_LEN 30
+// SYSLOG_IDENTIFIER=
+#define SYSLOG_IDENTIFIER_EQ_LEN 18
 static char short_cuuid[TRUNC_ID_LEN + 1];
 static char *cuuid = NULL;
 static char *name = NULL;
@@ -55,6 +64,8 @@ static char *container_id = NULL;
 static char *container_name = NULL;
 static char *container_tag = NULL;
 static size_t container_tag_len;
+static char *syslog_identifier = NULL;
+static size_t syslog_identifier_len;
 
 typedef struct {
 	int iovcnt;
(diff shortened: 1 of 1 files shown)


haircommander:
fixes https://github.com/containers/conmon/issues/293

rhatdan:
LGTM

rhatdan:
@giuseppe @flouthoc PTAL

src/ctr_logging.c
```diff
@@ -257,6 +274,9 @@ int write_journald(int pipe, char *buf, ssize_t buflen)
 		if (name && writev_buffer_append_segment(dev_null, &bufv, container_name, name_len + NAME_EQ_LEN) < 0)
 			return -1;
 
+		if (writev_buffer_append_segment(dev_null, &bufv, syslog_identifier, syslog_identifier_len) < 0)
```
flouthoc, line 277: No sure but i think we must check if `syslog_identifier` is valid before flushing. Maybe 
```c
if (syslog_identifier && writev_buffer_append_segment(dev_null, &bufv, syslog_identifier, syslog_identifier_len) < 0)
```

src/ctr_logging.c
```diff
@@ -126,6 +137,12 @@ void configure_log_drivers(gchar **log_drivers, int64_t log_size_max_, char *cuu
 			/* save the length so we don't have to compute every sd_journal_* call */
 			name_len = strlen(name);
 			container_name = g_strdup_printf("CONTAINER_NAME=%s", name);
+
+			syslog_identifier = g_strdup_printf("SYSLOG_IDENTIFIER=%s", name);
+			syslog_identifier_len = name_len + SYSLOG_IDENTIFIER_EQ_LEN;
```
giuseppe, line 142: should we account for `'='` as well?

src/ctr_logging.c
```diff
@@ -126,6 +137,12 @@ void configure_log_drivers(gchar **log_drivers, int64_t log_size_max_, char *cuu
...
+		} else {
+			syslog_identifier = g_strdup_printf("SYSLOG_IDENTIFIER=%s", short_cuuid);
+			syslog_identifier_len = TRUNC_ID_LEN + SYSLOG_IDENTIFIER_EQ_LEN;
```
giuseppe, line 145: same here

src/ctr_logging.c
haircommander: I suppose we could but I intentionally made it so that syslog_identifier was always defined (either container name or ID). if it'd make you more comfortable I can update

src/ctr_logging.c
haircommander: SYSLOG_IDENTIFIER_EQ_LEN is 18, which accounts for the `=`

src/ctr_logging.c
haircommander: the one thing I am concerned about is the null byte here, but in other uses we use `TRUNC_ID_LEN + CID_EQ_LEN` which doesn't seem to count null

src/ctr_logging.c
giuseppe: Ok thanks :)

src/ctr_logging.c
flouthoc: @haircommander oh it is completely fine then :) . Thanks for confirming.

flouthoc:
LGTM

giuseppe (approved):
LGTM

giuseppe:
@haircommander are these CI failures expected?  If so, please merge

haircommander:
I was able to reproduce the test failure, we want https://github.com/cri-o/cri-o/pull/5334 first

haircommander:
hold for https://github.com/containers/conmon/pull/297

rhatdan:
Since #297 was merged, I am going to merge this.

PR #303, opened 2021-10-08
conmon: fix error message

giuseppe:
the function called is chmod but the message was referring to chown.


rhatdan:
LGTM

PR #309, opened 2021-12-17
exit_command: unset subreaper attribute before running exit command

martinetd:
staying as subreaper means the exit command would not be able to start new daemonizes processes e.g. another podman container if it's been set to restart.

Fixes: 4d3dba9c4beb ("exit: fix race zombie reaper")
Fixes: https://github.com/containers/podman/issues/12635

I'm not 100% happy with this fix as written in the comment, it means we could still leak some process so #290 might come back... Unfortunately I don't see much other way about it: going back to reaping existing children and waiting only for the exact pid of the exit command as was done previously is just as bad, this is somewhere in between.

Anyway, I'm just mostly happy I finally figured why the double-fork didn't work, feel free to suggest other solutions.

### DIFF
--- a/src/ctr_exit.c
+++ b/src/ctr_exit.c
@@ -12,6 +12,7 @@
 #include <glib-unix.h>
 #include <signal.h>
 #include <stdlib.h>
+#include <sys/prctl.h>
 #include <unistd.h>
 
 volatile pid_t container_pid = -1;
@@ -158,6 +159,16 @@ void do_exit_command()
 	close_fd(&attach_socket_fd);
 	close_fd(&console_socket_fd);
 	close_fd(&attach_pipe_fd);
+
+	/*
+	 * We don't want the exit command to be reaped by the parent conmon
+	 * as that would prevent double-fork from doing its job.
+	 * Unfortunately, that also means that any new subchildren from
+	 * still running processes could also get lost
+	 */
+	if (prctl(PR_SET_CHILD_SUBREAPER, 0) != 0) {
+		nwarn("Failed to disable self subreaper attribute - might wait for indirect children a long time");
+	}
 
 	pid_t exit_pid = fork();
 	if (exit_pid < 0) {


rhatdan:
@haircommander @giuseppe PTAL

src/ctr_exit.c
```diff
@@ -159,6 +160,14 @@ void do_exit_command()
...
 	close_fd(&attach_pipe_fd);
 
+	/*
+	 * We don't want the exit command to be reaped by the parent conmon
+	 * as that would prevent double-fork from doing its job.
+	 * Unfortunately, that also means that any new subchildren from
+	 * still running processes could also get lost
+	 */
+	prctl(PR_SET_CHILD_SUBREAPER, 0);
+
```
haircommander, line 170: Underneath this there's a block
```c
	if (exit_pid) {
		int ret, exit_status = 0;

		/*
		 * Make sure to cleanup any zombie process that the container runtime
		 * could have left around.
		 */
		do {
			int tmp;

			exit_status = 0;
			ret = waitpid(-1, &tmp, 0);
			if (ret == exit_pid)
				exit_status = get_exit_status(tmp);
		} while ((ret < 0 && errno == EINTR) || ret > 0);

		if (exit_status)
			_exit(exit_status);

		return;
	}

```

I believe this block was to cleanup these grandchildren. what processes could we leak in podman cleanup if we drop the subreaper @mheon @giuseppe

giuseppe:
could `clone(CLONE_PARENT)` work here instead of resetting the subreaper flag?

src/ctr_exit.c
mheon: Assuming container restart, there are a few fork/execs - c/storage (reexec code), CNI plugins, and then a fresh Conmon (possible I'm forgetting a few but these are the ones I can recall). All of these could potentially leak if Podman were killed at the right time, though the only one where I'd say this is likely is CNI, which takes a long while to execute and gives a decently long window for this race.

giuseppe:
I am not even sure we care about the exit command return code (and that seems like the only reason why we wait for its completion).

@mheon could we just double fork and exit immediately instead of waiting for the exit command?

mheon:
The only reason I can think of to wait would be to potentially log the return code, which is not something we do right now. Given that, this SGTM

giuseppe:
@martinetd  would something like the following patch also fix the issue you've seen?

```diff
diff --git a/src/ctr_exit.c b/src/ctr_exit.c
index 7f57e7a..a61886b 100644
--- a/src/ctr_exit.c
+++ b/src/ctr_exit.c
@@ -140,61 +140,8 @@ static void close_fd(int *fd)
 	*fd = -1;
 }
 
-void do_exit_command()
+static void run_exit_command()
 {
-	if (sync_pipe_fd > 0) {
-		close(sync_pipe_fd);
-		sync_pipe_fd = -1;
-	}
-
-	if (signal(SIGCHLD, SIG_DFL) == SIG_ERR) {
-		_pexit("Failed to reset signal for SIGCHLD");
-	}
-
-	close_logging_fds();
-	close_fd(&terminal_ctrl_fd);
-	close_fd(&winsz_fd_w);
-	close_fd(&winsz_fd_r);
-	close_fd(&mainfd_stdin);
-	close_fd(&attach_socket_fd);
-	close_fd(&console_socket_fd);
-	close_fd(&attach_pipe_fd);
-
-	/*
-	 * We don't want the exit command to be reaped by the parent conmon
-	 * as that would prevent double-fork from doing its job.
-	 * Unfortunately, that also means that any new subchildren from
-	 * still running processes could also get lost
-	 */
-	prctl(PR_SET_CHILD_SUBREAPER, 0);
-
-	pid_t exit_pid = fork();
-	if (exit_pid < 0) {
-		_pexit("Failed to fork");
-	}
-
-	if (exit_pid) {
-		int ret, exit_status = 0;
-
-		/*
-		 * Make sure to cleanup any zombie process that the container runtime
-		 * could have left around.
-		 */
-		do {
-			int tmp;
-
-			exit_status = 0;
-			ret = waitpid(-1, &tmp, 0);
-			if (ret == exit_pid)
-				exit_status = get_exit_status(tmp);
-		} while ((ret < 0 && errno == EINTR) || ret > 0);
-
-		if (exit_status)
-			_exit(exit_status);
-
-		return;
-	}
-
 	/* Count the additional args, if any.  */
 	size_t n_args = 0;
 	if (opt_exit_args)
@@ -216,9 +163,36 @@ void do_exit_command()
 		sleep(opt_exit_delay);
 	}
 
-	execv(opt_exit_command, args);
+	_exit(execv(opt_exit_command, args));
+}
+
+void do_exit_command()
+{
+	if (sync_pipe_fd > 0) {
+		close(sync_pipe_fd);
+		sync_pipe_fd = -1;
+	}
+
+	if (signal(SIGCHLD, SIG_DFL) == SIG_ERR) {
+		_pexit("Failed to reset signal for SIGCHLD");
+	}
+
+	close_logging_fds();
+	close_fd(&terminal_ctrl_fd);
+	close_fd(&winsz_fd_w);
+	close_fd(&winsz_fd_r);
+	close_fd(&mainfd_stdin);
+	close_fd(&attach_socket_fd);
+	close_fd(&console_socket_fd);
+	close_fd(&attach_pipe_fd);
 
-	/* Should not happen, but better be safe. */
+	pid_t p = fork();
+	if (p)
+		_exit(p);
+	p = fork();
+	if (p)
+		_exit(p);
+	run_exit_command();
 	_exit(EXIT_FAILURE);
 }
 ```

**Comment by giuseppe** [comment]
> ```diff
> +	pid_t p = fork();
> +	if (p)
> +		_exit(p);
> +	p = fork();
> +	if (p)
> +		_exit(p);
> +	run_exit_command();
> ```

and since we exit immediately, we don't need the double fork

**Comment by martinetd** [comment]
> could clone(CLONE_PARENT) work here instead of resetting the subreaper flag?

(a bit moot since different approach was chosen but answering anyway) I think `CLONE_PARENT` would work, but means we can't catch these podman cleanup children either

> @martinetd would something like the following patch also fix the issue you've seen?

My very first idea was to just exec into the exit command directly, without any fork.

This would allow it to be able to handle any existing children if they want to (which would have been handy if podman cleanup itself was also directly exec'ing into conmon in sync mode, but since it's creating conmon as a child it's not actually useful)

I believe either this or what you've done should work, I'll give it a try in a bit

That being said I don't know what inpact this will have on #290 which was the reason for the breaking commit (some namespace not reaped up properly).
It seems difficult to reproduce and I don't use podman remote so I didn't try but someone who worked on it might shed some light -- if the parent conmon process itself exits its children will be reaped by pid1 anyway and it'll work out if they die naturally, but they'll be hard to kill if cleanup is required...

**Comment by martinetd** [comment]
Just tested @giuseppe 's patch removing one fork:
```diff
diff --git a/src/ctr_exit.c b/src/ctr_exit.c
index a61886b4395e..9a3d1e54496e 100644
--- a/src/ctr_exit.c
+++ b/src/ctr_exit.c
@@ -168,6 +168,8 @@ static void run_exit_command()
 
 void do_exit_command()
 {
+       int p;
+
        if (sync_pipe_fd > 0) {
                close(sync_pipe_fd);
                sync_pipe_fd = -1;
@@ -186,12 +188,11 @@ void do_exit_command()
        close_fd(&console_socket_fd);
        close_fd(&attach_pipe_fd);
 
-       pid_t p = fork();
-       if (p)
-               _exit(p);
        p = fork();
-       if (p)
-               _exit(p);
+       if (p > 0)
+               _exit(0);
+       if (p < 0)
+               _exit(EXIT_FAILURE);
        run_exit_command();
        _exit(EXIT_FAILURE);
 }
```

or both
```diff
diff --git a/src/ctr_exit.c b/src/ctr_exit.c
index a61886b4395e..1e39356230df 100644
--- a/src/ctr_exit.c
+++ b/src/ctr_exit.c
@@ -186,12 +186,6 @@ void do_exit_command()
        close_fd(&console_socket_fd);
        close_fd(&attach_pipe_fd);
 
-       pid_t p = fork();
-       if (p)
-               _exit(p);
-       p = fork();
-       if (p)
-               _exit(p);
        run_exit_command();
        _exit(EXIT_FAILURE);
 }
```

In both case I restart worked and left no stray process in my case, if we're going that way I'd say not forking at all is simpler (although forking clears some flags like the prctl subreaper setting...), but I have no preference.
Note double-forking here has no effect because of the subreaper property, the grandchild will become a direct child when the previous child dies. But since we exit anyway it doesn't matter.

(And once again, I haven't tried to reproduce the original leftover namespace issue, but it might be back)

giuseppe:
one risk with my suggested patch is that the conmon process exits immediately, and that could be a problem when running in a systemd unit and having the main process exits before the exit command terminated.

Probably your approach is safer for now, so we have not to worry about these edge cases.

src/ctr_exit.c
```diff
@@ -159,6 +160,14 @@ void do_exit_command()
 	close_fd(&console_socket_fd);
 	close_fd(&attach_pipe_fd);
 
+	/*
+	 * We don't want the exit command to be reaped by the parent conmon
+	 * as that would prevent double-fork from doing its job.
+	 * Unfortunately, that also means that any new subchildren from
+	 * still running processes could also get lost
+	 */
+	prctl(PR_SET_CHILD_SUBREAPER, 0);
```
giuseppe, line 169: could you check for errors?

martinetd: I don't mind checking, but what should we do if the prctl call fails?
In my opinion we need to run the exit command anyway so at most I can add a warning (nwarnf with errno?), is that OK?

martinetd: hm, perhaps run and skip the wait loop if the prctl failed, but I won't have any way of testing this really (error injection is fun but takes a bit more time than I have short term)

giuseppe: I think a warning is enough, at least we have a way to know if we get into this state if we see again the issue you are addressing

martinetd: works for me! added a `nwarn` and force pushed. I've also fixed the issue fix link now the issue has been transferred to #310 (thanks)

giuseppe (approved):
LGTM

haircommander:
LGTM

rhatdan:
Do we need to get the Tests to pass to get this to merge?

haircommander:
CI could use some love, I don't think it's necessary for this one

martinetd:
Thanks!

Issue #167, opened 2020-05-29
infinite loop in eventfds


cgwalters:
I have 

```
$ rpm -q conmon podman
conmon-2.0.16-2.fc32.x86_64
podman-1.9.2-1.fc32.x86_64
```

(On a locally built f32 silverblue style system)

Seeing this after I exited one of my toolbox containers from conmon, it's looping infinitely using 100% of one core:

```
[pid 301357] poll([{fd=6, events=POLLIN}, {fd=10, events=POLLIN}, {fd=13, events=POLLIN}, {fd=14, events=POLLIN}, {fd=16, events=POLLIN}, {fd=19, events=POLLIN}], 6, -1) = 2 ([{fd=6, revents=POLLIN}, {fd=10, revents=POLLNVAL}])
[pid 301357] read(6, "\2\0\0\0\0\0\0\0", 16) = 8
[pid 301357] write(6, "\1\0\0\0\0\0\0\0", 8) = 8
[pid 301357] write(6, "\1\0\0\0\0\0\0\0", 8) = 8
[pid 301357] poll([{fd=6, events=POLLIN}, {fd=10, events=POLLIN}, {fd=13, events=POLLIN}, {fd=14, events=POLLIN}, {fd=16, events=POLLIN}, {fd=19, events=POLLIN}], 6, -1) = 2 ([{fd=6, revents=POLLIN}, {fd=10, revents=POLLNVAL}])
[pid 301357] read(6, "\2\0\0\0\0\0\0\0", 16) = 8
[pid 301357] write(6, "\1\0\0\0\0\0\0\0", 8) = 8
[pid 301357] write(6, "\1\0\0\0\0\0\0\0", 8) = 8
[pid 301357] poll([{fd=6, events=POLLIN}, {fd=10, events=POLLIN}, {fd=13, events=POLLIN}, {fd=14, events=POLLIN}, {fd=16, events=POLLIN}, {fd=19, events=POLLIN}], 6, -1) = 2 ([{fd=6, revents=POLLIN}, {fd=10, revents=POLLNVAL}])
[pid 301357] read(6, "\2\0\0\0\0\0\0\0", 16) = 8
[pid 301357] write(6, "\1\0\0\0\0\0\0\0", 8) = 8
[pid 301357] write(6, "\1\0\0\0\0\0\0\0", 8) = 8
[...]

PR #312, opened 2022-01-04
remove fd sources from glib

haircommander:
or else we may close them without removing them, causing an infinte EPOLL issue

fixes https://github.com/containers/conmon/issues/167

### DIFF
--- a/src/ctr_exit.c
+++ b/src/ctr_exit.c
@@ -6,6 +6,7 @@
 #include "parent_pipe_fd.h"
 #include "globals.h"
 #include "ctr_logging.h"
+#include "close_fds.h"
 
 #include <errno.h>
 #include <glib.h>
@@ -151,6 +152,7 @@ void do_exit_command()
 		_pexit("Failed to reset signal for SIGCHLD");
 	}
 
+	remove_g_unix_fds();
 	close_logging_fds();
 	close_fd(&terminal_ctrl_fd);
 	close_fd(&winsz_fd_w);
--- a/src/cgroup.c
+++ b/src/cgroup.c
@@ -5,6 +5,7 @@
 #include "utils.h"
 #include "cli.h"
 #include "config.h"
+#include "close_fds.h"
 
 #include <fcntl.h>
 #include <glib.h>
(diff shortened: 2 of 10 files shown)


src/ctr_exit.c
```diff
@@ -151,6 +152,7 @@ void do_exit_command()
 		_pexit("Failed to reset signal for SIGCHLD");
 	}
 
+	remove_g_unix_fds();
```
cgwalters, line 155: This does seem like a potential bug, but I'm not sure it's the actual bug here.  This problem can only occur if we reenter `g_main_loop_run()` after closing a fd behind its back.

I am not seeing where we'd being doing that.  It looks like inside `do_exit_command()` here we do some prep stuff and then just `execve()`.

Are you sure there's not another place where we're closing a fd while we're inside the `g_main_loop_run()` invocation?

cgwalters:
Looking at this a little bit more, the logic in `stdio_cb()` seems suspect here; notice we're closing the fd inside the callback for that fd, then returning `g_source_remove()` which should remove the source.  But I'm not sure it's *guaranteed* that the main loop won't try to reenter epoll.

I think what we really want here is (in Rust terms) for the `GSource` to own the fd, not just reference it.  Then we'd only have one static array (of sources) and not two (one of sources, one of fds).

We may be able to do this with `g_unix_fd_add_full()` and pass the fd as user data, and invoke `close()` there.

haircommander:
I have tried two different methods than the one you've described: one where we remove the source immediately after accepting on the corresponding socket https://github.com/containers/conmon/pull/313
and one here where we make sure to remove the source when we close it (here)

haircommander:
oh you said stdio_sb, I focused on attach_cb as I found a case where we didn't remove the source. gotta check stdio_cb too...

haircommander:
> notice we're closing the fd inside the callback for that fd, then returning g_source_remove() which should remove the source. But I'm not sure it's guaranteed that the main loop won't try to reenter epoll.

I feel like we'd be in trouble if G_SOURCE_REMOVE was or wasn't removing the fds. It's a programming error to double remove a source (as they reuse the source guints), and we do it all over the place. We use it all over the place to remove the fds. Except for in attach_cb afaict

rhatdan:
@haircommander is this still needed since other PR fixed the original issue?

haircommander:
nope! closing

PR #313, opened 2022-01-04
conn_sock: remove attach handler

haircommander:
alternative to https://github.com/containers/conmon/pull/312


haircommander:
@mrunalp @giuseppe @mheon do we expect a client to attempt to attach to a container more than once?

mheon:
With `podman attach` being available, it is definitely a possibility.

`podman start --attach` a container, kill that attach session via detach keys, `podman attach` to the same container.

haircommander:
okay we'll need to be smarter then

PR #314, opened 2022-01-09
Avoid mainfd_std{in,out} sharing the same file descriptor.

likan999:
This causes problem that if one is closed, the glib source of the other is not removed, which causes infinite loop and 100% cpu usage.

Fixes: #167

### DIFF
--- a/src/ctrl.c
+++ b/src/ctrl.c
@@ -14,6 +14,7 @@
 #include <sys/socket.h>
 #include <sys/stat.h>
 #include <termios.h>
+#include <unistd.h>
 
 static void resize_winsz(int height, int width);
 static gboolean read_from_ctrl_buffer(int fd, gboolean (*line_process_func)(char *));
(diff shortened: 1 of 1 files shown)


likan999:
@haircommander PTAL, thanks.

haircommander:
LGTM thank you @likan999

@giuseppe @rhatdan @flouthoc PTAL

flouthoc:
LGTM

I was reading the issue really nice finding.  Since I am not familiar with the entire code base but I am assuming we must have a call for `g_source_remove` for mainfd_stdout as well.

CI is red though. @giuseppe @rhatdan PTAL

haircommander:
CI is woefully out of date--kubernetes test passed so that is good signal the PR doesn't break anything

likan999:
> LGTM
> 
> I was reading the issue really nice finding. Since I am not familiar with the entire code base but I am assuming we must have a call for `g_source_remove` for mainfd_stdout as well.
> 
> CI is red though. @giuseppe @rhatdan PTAL

I think the source is removed when `mainfd_stdout` is closed: [code](https://github.com/containers/conmon/blob/b8748f5609b8317cc34d4137c42ce5d4f989bb12/src/ctr_stdio.c#L80). So we don't need to worry about the FD leak here (unless there is some corner case again).

PR #317, opened 2022-01-12
fix: cgroup: Free memory_cgroup_file_path if open fails

sheabot:
In `src/cgroup.c`, the variable `memory_cgroup_file_path` is allocated with `g_build_filename()`, but it is not free'd in the error path. In the happy path, it would be properly free'd in `oom_cb_cgroup_v1()`.

Following that, in `oom_cb_cgroup_v1()` free this memory using `g_free()` instead of `free()`. Functionally, there is no difference, but semantically it's better.


rhatdan:
LGTM

saschagrunert (approved):
LGTM

rhatdan:
Thanks @sheabot

Issue #242, opened 2021-02-27
extra newlines in logged container output


liewegas:
container output, as seen in journald, has random newlines inserted.

**Steps to reproduce the issue:**

There is probably a simpler way to do this, but this is what I did:

1. 
```
cat > /etc/systemd/system/conmon-test.service
[Unit]
Description=test

[Service]
ExecPre=-podman rm conmon-test
ExecStart=podman run -d --rm --name conmon-test --log-driver journald --entrypoint bash centos -c "find /"
```
2. ``systemctl start conmon-test``
3. ``journalctl -u conmon-test``

**Describe the results you received:**

Output has random newlines.  For example:

```
Feb 27 00:37:34 smithi189 conmon[11377]: /proc/bus/pci/ff/10.6
Feb 27 00:37:34 smithi189 conmon[11377]: /proc/bus/pci/ff/10.7
Feb 27 00:37:34 smithi189 conmon[11377]: /proc/bus/pci/ff/12.0
Feb 27 00:37:34 smithi189 conmon[11377]: /proc/bus/pci/ff/12.1
Feb 27 00:37:34 smithi189 conmon[11377]: /proc
Feb 27 00:37:34 smithi189 conmon[11377]: /bus/pci/ff/13.0
Feb 27 00:37:34 smithi189 conmon[11377]: /proc/bus/pci/ff/13.1
Feb 27 00:37:34 smithi189 conmon[11377]: /proc/bus/pci/ff/13.2
Feb 27 00:37:34 smithi189 conmon[11377]: /proc/bus/pci/ff/13.3
```

and

```
Feb 27 00:37:34 smithi189 conmon[11377]: /proc/irq/82/spurious
Feb 27 00:37:34 smithi189 conmon[11377]: /proc/irq/82/smp_affinity
Feb 27 00:37:34 smithi189 conmon[11377]: /proc/irq/82/affinity_hint
Feb 27 00:37:34 smithi189 conmon[11377]: /proc/irq/82/smp_affinity_list
Feb 27 00:37:34 smithi189 conmon[11377]: /proc/irq/82/effective_affinity
Feb 27 00:37:34 smithi189 conmon[11377]: /proc/
Feb 27 00:37:34 smithi189 conmon[11377]: irq/82/effective_affinity_list
Feb 27 00:37:34 smithi189 conmon[11377]: /proc/irq/84
Feb 27 00:37:34 smithi189 conmon[11377]: /proc/irq/84/node
[...]

PR #321, opened 2022-01-20
logging: buffer partial messages to journald

cfsnyder:
Ensures that log lines less than or equal to 8k in size are logged as a single journald message by buffering partial messages from subsequent read calls until a newline is observed (or buffer is full).

Fixes: #242
Signed-off-by: Cory Snyder <csnyder@iland.com>

### DIFF
--- a/src/ctr_logging.c
+++ b/src/ctr_logging.c
@@ -1,6 +1,7 @@
 #define _GNU_SOURCE
 #include "ctr_logging.h"
 #include "cli.h"
+#include "config.h"
 #include <string.h>
 
 // if the systemd development files were found, we can log to systemd
@@ -230,10 +231,19 @@ bool write_to_logs(stdpipe_t pipe, char *buf, ssize_t num_read)
 
 
 /* write to systemd journal. If the pipe is stdout, write with notice priority,
- * otherwise, write with error priority
+ * otherwise, write with error priority. Partial lines (that don't end in a newline) are buffered
+ * between invocations. A 0 buflen argument forces a buffered partial line to be flushed.
  */
 int write_journald(int pipe, char *buf, ssize_t buflen)
 {
+	static char stdout_partial_buf[STDIO_BUF_SIZE];
+	static size_t stdout_partial_buf_len = 0;
+	static char stderr_partial_buf[STDIO_BUF_SIZE];
+	static size_t stderr_partial_buf_len = 0;
+
+	char *partial_buf;
+	size_t *partial_buf_len;
+
 	/* When using writev_buffer_append_segment, we should never approach the number of
 	 * entries necessary to flush the buffer. Therefore, the fd passed in is for /dev/null
 	 */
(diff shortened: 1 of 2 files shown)


rhatdan:
@giuseppe @haircommander @mheon PTAL

haircommander:
amazing, thank you @cfsnyder ! for some reason I had worked myself up and thought the solution would be more complex. Before we merge this, I'd like to get CI a bit more in order to make sure we have a proper smoke test against breaking changes here. I've done a manual test and it works great!

src/ctr_logging.c
```diff
@@ -246,30 +256,47 @@ int write_journald(int pipe, char *buf, ssize_t buflen)
...
 		 * entire string. Copying every line doesn't make very much sense, so instead we do this tmp_line_end
 		 * hack to emulate separate strings.
 		 */
 		char tmp_line_end = buf[line_len];
 		buf[line_len] = '\0';
 
-		_cleanup_free_ char *message = g_strdup_printf("MESSAGE=%s", buf);
-		if (writev_buffer_append_segment(dev_null, &bufv, message, line_len + MESSAGE_EQ_LEN) < 0)
+		ssize_t msg_len = line_len + MESSAGE_EQ_LEN + *partial_buf_len;
+		partial_buf[*partial_buf_len] = '\0';
```
TomSweeneyRedHat, line 292: This is where C bites me.  Do you need a length check here to make sure you're not writing past the buffer?  I'm just concerned with a max length partial_buf.

cfsnyder: Thanks for checking @TomSweeneyRedHat. Verify my logic here, but I believe that the invariant is upheld. Line 280 is the only statement that updates `partial_buf_len` to a non-zero value. On that line, it is updated to the sum of the previous `partial_buf_len` and `line_len`. The if statement surrounding that update guarantees that the maximum value of `line_len` is  `STDIO_BUF_SIZE - partial_buf_len - 1`. Therefore, the maximum value that `partial_buf_len` might obtain is  `partial_buf_len + STDIO_BUF_SIZE - partial_buf_len - 1`, which simplifies to `STDIO_BUF_SIZE - 1`. Of course, an index of `STDIO_BUF_SIZE - 1` is always valid on an array of size `STDIO_BUF_SiZE`.

(In fact, prior to double checking this, I was unnecessarily subtracting 1 from the right side of the conditional on line 278 - so I'm glad you asked!)

src/ctr_logging.c
```diff
@@ -246,30 +256,47 @@ int write_journald(int pipe, char *buf, ssize_t buflen)
...
+		_cleanup_free_ char *message = g_strdup_printf("MESSAGE=%s%s", partial_buf, buf);
+		if (writev_buffer_append_segment(dev_null, &bufv, message, msg_len) < 0)
```
giuseppe, line 294: would it work if you don't allocate a string here (I know it is pre-existing code) but now we are already having an extra buffered copy:
```
if (writev_buffer_append_segment(dev_null, &bufv, "MESSAGE=", MESSAGE_EQ_LEN) < 0)
....
if (writev_buffer_append_segment(dev_null, &bufv, partial_buf, *partial_buf_len) < 0)
...
if (writev_buffer_append_segment(dev_null, &bufv, buf, line_len) < 0)
...
```

cfsnyder: Each structure in the iovec that is passed to sd_journal_sendv must represent one field of the entry to submit. What you have above would create 3 separate structures in the iovec for the MESSAGE field, so that doesn't quite work. Ultimately all components of the entry need to be stored in contiguous memory, one way or another. It would be possible to avoid the heap allocation by placing the message field in a char array on the stack (and memcpy the components of the message there). What do you think?

rhatdan:
If/When this gets merged, I think it is time for a new Conmon Release, to align with Podman 4.0.
@haircommander

haircommander:
I agree I'd deem this 2.1 material :)

haircommander:
heya @cfsnyder can you rebase on latest main? I've fixed most of CI so we should get a better signal on this PR

cfsnyder:
Thanks @haircommander, just rebased!

haircommander:
LGTM, thanks a bunch @cfsnyder

PR #326, opened 2022-01-31
Add journald support to meson build

### LINKED ISSUE #325: Meson build doesn't handle systemd journald
I think the title says it all. When building with meson, conmon is built without support for systemd-journald, even though journald is present in the build environment

AdrianVovk:
Fixes #325


AdrianVovk:
Note: If/once you switch to depending on meson 0.60, you can combine the two library names for systemd like so: `dependency('libsystemd-journal', 'libsystemd', required : false)`

saschagrunert (approved):
LGTM

haircommander:
LGTM

haircommander:
thanks for opening this @AdrianVovk

PR #338, opened 2022-05-26
Port conmon to FreeBSD

dfr:
The main things which I changed were:

* Factor out calls to prctl so that I can replace with FreeBSD procctl
  equivalents without making an ifdef mess.
* Work around the fact that FreeBSD doesn't allow chown on sockets and
  non-named pipes.
* Use a different workaround for socket paths too long for sockaddr_un - the
  Linux O_PATH thing was replaced with a near equivalent use of FreeBSD's
  bindat.
* Some rearranging of ifdefs so that I can cut cgroups and seccomp out of the
  FreeBSD build.
* Reading from pipes and/or ptys can error with EGAIN for some reason - these
  calls can simply be retried similar to EINTR.

### DIFF
--- a/src/conn_sock.c
+++ b/src/conn_sock.c
@@ -112,6 +112,12 @@ static void bind_relative_to_dir(int dir_fd, int sock_fd, const char *path)
 #endif
 
 #ifdef __FreeBSD__
+
+// FreeBSD earlier than 13.1-RELEASE doesn't have O_PATH
+#ifndef O_PATH
+#define O_PATH 0
+#endif
+
 static void bind_relative_to_dir(int dir_fd, int sock_fd, const char *path)
 {
 	struct sockaddr_un addr;
--- a/src/conn_sock.c
+++ b/src/conn_sock.c
@@ -86,10 +86,54 @@ char *setup_seccomp_socket(const char *socket)
 }
 
 #ifdef __linux__
+static void bind_relative_to_dir(int dir_fd, int sock_fd, const char *path)
+{
+	struct sockaddr_un addr;
+
+	/*
+	 * To be able to access the location of the attach socket, without first creating the attach socket
+	 * but also be able to handle arbitrary length paths, we open the parent dir (base_path), and then use
+	 * the corresponding entry in `/proc/self/fd` to act as the path to base_path, then we use the socket_relative_name
+	 * to actually refer to the file where the socket will be created below.
+	 */
+	addr.sun_family = AF_UNIX;
+	if (dir_fd == -1) {
+		strncpy(addr.sun_path, path, sizeof(addr.sun_path) - 1);
+	} else {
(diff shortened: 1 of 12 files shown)


src/conn_sock.c
```diff
@@ -151,6 +152,75 @@ static char *setup_socket(int *fd, const char *path)
...
+	ndebugf("addr{sun_family=AF_UNIX, sun_path=%s}", addr.sun_path);
+
+	/* Bind to the console socket path. */
+	*fd = socket(AF_UNIX, SOCK_STREAM | SOCK_CLOEXEC, 0);
+	if (*fd < 0)
+		pexit("Failed to create socket");
+	if (bindat(sfd == -1 ? AT_FDCWD : sfd, *fd, (struct sockaddr *)&addr, sizeof(addr)) < 0)
+		pexit("Failed to bind to console-socket");
+	if (fchmodat(sfd, addr.sun_path, 0700, AT_SYMLINK_NOFOLLOW))
+		pexit("Failed to change console-socket permissions HERE!!");
```
haircommander, line 217: hm is this HERE!! leftover from debugging

haircommander (commented):
Interesting work here @dfr I am not familiar with the work you're doing, how far along is BSD support for podman and friends?

No real issues other than potential concerns of code readability. maybe we could separate linux and BSD files? 

generally it LGTM though

src/conn_sock.c
dfr: > hm is this HERE!! leftover from debugging

Oops - I'll get rid of it.

dfr:
> Interesting work here @dfr I am not familiar with the work you're doing, how far along is BSD support for podman and friends?
> 
> No real issues other than potential concerns of code readability. maybe we could separate linux and BSD files?
> 
> generally it LGTM though

I have working proof-of-concept builds of podman and buildah, both of which need a little tidying before asking for review. Both are waiting also waiting on a few PRs to vendor'ed modules so I'm in no rush there. I'm also trying to get a FreeBSD OCI working group moving so that we can add FreeBSD definitions to the OCI runtime spec.

For this change, I think the least readable part is in conn_sock.c and that might be better split into platform-specific source files.

dfr:
I reworked the conn_sock.c changes to reduce the duplication between the two platforms and hopefully to improve readability.

haircommander:
LGTM, I'm going to try to fix CI in https://github.com/containers/conmon/pull/341 to get a better signal of any potential breakages

dfr:
Thanks for taking a look. I did successfully run 'make test' on an ubuntu
VM so hopefully I didn't break too much here.

dfr:
The rebase conflict is around the addition of signalfd(). On FreeBSD we can use kqueue for the same purpose so I need to add another tiny abstraction layer.

Issue #337, opened 2022-05-23
2.1.1: build fails because incorrect checking dependency for `libsystemd`


kloczek:
Meson passes without exit 1 anf build fails
```console
+ /usr/bin/meson --buildtype=plain --prefix=/usr --libdir=/usr/lib64 --libexecdir=/usr/libexec --bindir=/usr/bin --sbindir=/usr/sbin --includedir=/usr/include --datadir=/usr/share --mandir=/usr/share/man --infodir=/usr/share/info --localedir=/usr/share/locale --sysconfdir=/etc --localstatedir=/var --sharedstatedir=/var/lib --wrap-mode=nodownload --auto-features=enabled . x86_64-redhat-linux-gnu
The Meson build system
Version: 0.62.1
Source dir: /home/tkloczko/rpmbuild/BUILD/conmon-2.1.1
Build dir: /home/tkloczko/rpmbuild/BUILD/conmon-2.1.1/x86_64-redhat-linux-gnu
Build type: native build
WARNING: You should add the boolean check kwarg to the run_command call.
         It currently defaults to false,
         but it will default to true in future releases of meson.
         See also: https://github.com/mesonbuild/meson/issues/9300
Project name: conmon
Project version: 2.1.1
C compiler for the host machine: /usr/bin/gcc (gcc 12.1.1 "gcc (GCC) 12.1.1 20220507 (Red Hat 12.1.1-1)")
C linker for the host machine: /usr/bin/gcc ld.bfd 2.38-9
Host machine cpu family: x86_64
Host machine cpu: x86_64
Program git found: YES (/usr/bin/git)
Found pkg-config: /usr/bin/pkg-config (1.8.0)
Run-time dependency glib-2.0 found: YES 2.72.1
Checking for function "dlopen" : YES
Found CMake: /usr/bin/cmake (3.23.1)
Run-time dependency libsystemd-journal found: NO (tried pkgconfig and cmake)
Run-time dependency libsystemd found: NO (tried pkgconfig and cmake)
Build targets in project: 1
NOTICE: Future-deprecated features used:
 * 0.56.0: {'meson.source_root'}

conmon 2.1.1

  User defined options
    auto_features : enabled
    bindir        : /usr/bin
    buildtype     : plain
    datadir       : /usr/share
    includedir    : /usr/include
    infodir       : /usr/share/info
    libdir        : /usr/lib64
    libexecdir    : /usr/libexec
    localedir     : /usr/share/locale
    localstatedir : /var
[...]

Issue #275, opened 2021-06-09
errors occurred during building


ghost:
I have encountered a error  building from source code.   Is my gcc version mismatched?

steps:

1.git clone ...
2. make 

os :  ubuntu16.04 
gcc : 5.4.0

![image](https://user-images.githubusercontent.com/33537489/121305157-da6aa480-c92f-11eb-852b-7189d56ddaaa.png)

haircommander:
interesting, this should have been fixed in https://github.com/containers/conmon/pull/268
does your version have that commit?

ccnankai:
I pulled the latest code. 
![image](https://user-images.githubusercontent.com/33537489/121471675-aad0a080-c9f2-11eb-8151-dfc7723ed3ad.png)

haircommander:
@giuseppe any ideas?

giuseppe:
could you try to apply the following patch?

```diff
$ git diff
diff --git a/src/seccomp_notify.h b/src/seccomp_notify.h
index 9dda848..de1d441 100644
--- a/src/seccomp_notify.h
+++ b/src/seccomp_notify.h
@@ -3,6 +3,10 @@
 
 #include "seccomp_notify_plugin.h"
 
+#ifndef SECCOMP_RET_USER_NOTIF
+#undef USE_SECCOMP
+#endif
+
 #if USE_SECCOMP
 
 struct seccomp_notify_context_s;
```

ccnankai:
@giuseppe 
After applying this patch, this error is still reported when compiling. It may be that there is a problem with my system environment, let me troubleshoot again. 
Thank you very much.

![image](https://user-images.githubusercontent.com/33537489/121623864-0d817500-caa3-11eb-8f66-bddb9dd0f539.png)

![image](https://user-images.githubusercontent.com/33537489/121623896-20944500-caa3-11eb-9857-8fa31d3ff608.png)

giuseppe:
this is what I get on Ubuntu 16.04:

```
/usr/include/seccomp.h:#define SECCOMP_RET_USER_NOTIF	0x7fc00000U
root@df988d928f75:/conmon# cat /etc/issue
Ubuntu 16.04.7 LTS \n \l

root@df988d928f75:/conmon# grep -r SECCOMP_RET_USER_NOTIF /usr/include/
/usr/include/seccomp.h:/* SECCOMP_RET_USER_NOTIF was added in kernel v5.0. */
[...]

PR #339, opened 2022-05-27
meson: optionally build seccomp if it supports notify

haircommander:

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -34,6 +34,10 @@ add_project_arguments('-Os', '-Wall', '-Werror',
                       language : 'c')
 
 glib = dependency('glib-2.0')
+seccomp = dependency('libseccomp', version : '>= 2.5.2')
+if seccomp.found()
+  add_project_arguments('-DUSE_SECCOMP=1', language : 'c')
+endif
 
 cc = meson.get_compiler('c')
 null_dep = dependency('', required : false)
@@ -86,7 +90,7 @@ executable('conmon',
             'src/utils.h',
             'src/seccomp_notify.c',
             'src/seccomp_notify.h'],
-           dependencies : [glib, libdl, sd_journal],
+           dependencies : [glib, libdl, sd_journal, seccomp],
            install : true,
            install_dir : join_paths(get_option('libexecdir'), 'podman'),
 )


haircommander:
theoretically this fixes https://github.com/containers/conmon/issues/337

meson.build
```diff
@@ -45,12 +45,19 @@ endif
...
 sd_journal = dependency('libsystemd-journal', required : false)
 if not sd_journal.found()
-	sd_journal = dependency('libsystemd', required : false)
+  sd_journal = dependency('libsystemd', required : false)
 endif
 if sd_journal.found()
 	add_project_arguments('-DUSE_JOURNALD=1', language : 'c')
 endif
 
+if run_command('./hack/seccomp-notify.sh', check: false).returncode() == 0
```
eli-schwartz, line 54: This relies on a shell script to do compile checks, which doesn't respect $CC (there may not necessarily be a `cc` on $PATH, though there likely is) or $PKG_CONFIG (which may point to `pkgconf`).

Maybe it should be ported to something Meson understands natively?

meson.build
```diff
@@ -45,12 +45,19 @@ endif
...
+  seccomp = dependency('libseccomp', required : false)
```
eli-schwartz, line 55: You don't seem to ever add this dependency anywhere?

eli-schwartz (changes requested):

kloczek:
IMO that is not correct solution.
if libseccomp API is used independently from libsystemd is used simple in meson detection should be two checks for those two libraries.
How it was found in https://github.com/containers/conmon/issues/337 containers uses directly libseccomp API even if it was not possible to detect libsystemd  so meson detection should not rely on indirect dependencies 

Using some additional script to check availability of the `seccomp_notif_sizes ()` is as well incorrect because all what dhould be checked is just minimum version of the pkgconfig module.

kloczek:
Just checked and it is not possible to build libseccomp without that `seccomp_notif_sizes ()` routine so checking it is pointless.

haircommander:
> 

Actually , the only reason we link with libseccomp is to use seccomp notify, which requires that structure

meson.build
haircommander: fun fact: I've never used meson before! thanks for the heads up :-)

haircommander:
I've updated with some suggestions, thanks for the reviews y'all

> Using some additional script to check availability of the seccomp_notif_sizes () is as well incorrect because all what dhould be checked is just minimum version of the pkgconfig module.

the problem is we need to know if seccomp notify is supported, but I believe packaged libseccomps don't necessarily correctly reflect that with pkg-config (for instance if support was backported) hence the additional check by compiling

shenghaoyang:
Maybe it would be cleaner to use the built-in test features of meson? See https://mesonbuild.com/Compiler-properties.html - there's a brute force "does snippet compile" function and a separate helper for sizeof: https://mesonbuild.com/Compiler-properties.html#expression-size. Should do the trick for structure presence detection.

Bonus: since it uses the same C compiler meson is setup to use, there shouldn't be any issues with the check being done with one compiler and the build performed by another.

meson.build
eli-schwartz: That's not something Meson understands natively. Also that doesn't help if someone configures Meson using a machine file (e.g. cross-compilation).

haircommander:
thanks for the help everyone, I have attempted with the sizeof function

haircommander:
I think that frankensystem was ubuntu: https://github.com/containers/conmon/issues/275

since 16.04 is out of service, and ubuntu 18.04 is on its way, maybe we won't need it much longer

For now, risk seems low to keep it

meson.build
```diff
@@ -45,12 +45,17 @@ endif
...
 sd_journal = dependency('libsystemd-journal', required : false)
 if not sd_journal.found()
-	sd_journal = dependency('libsystemd', required : false)
+  sd_journal = dependency('libsystemd', required : false)
 endif
 if sd_journal.found()
 	add_project_arguments('-DUSE_JOURNALD=1', language : 'c')
 endif
 
+seccomp = dependency('libseccomp', required : false)
```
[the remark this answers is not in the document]
haircommander: fixed!

rhatdan:
@haircommander What is going on with this?

kloczek:
IM that PR soes not make any sense because notify is supported since exact version of the libseccomp so checking dependencies usimg only `dependency()` covers 100% what is needed.
I have my own version fix of that issue:
```patch
--- a/meson.build
+++ b/meson.build
@@ -33,6 +33,10 @@
                       language : 'c')

 glib = dependency('glib-2.0')
+seccomp = dependency('libseccomp', version : '>= 2.5.2')
+if seccomp.found()
+  add_project_arguments('-DUSE_SECCOMP=1', language : 'c')
+endif

 cc = meson.get_compiler('c')
 null_dep = dependency('', required : false)
@@ -85,7 +86,7 @@
             'src/utils.h',
             'src/seccomp_notify.c',
             'src/seccomp_notify.h'],
-           dependencies : [glib, libdl, sd_journal],
+           dependencies : [glib, libdl, sd_journal, seccomp]
            install : true,
            install_dir : join_paths(get_option('libexecdir'), 'podman'),
 )
```
With that patch hack/seccomp-notify.sh can be deleted.

haircommander:
adapted as suggested. @rhatdan PTAL and I'll cut v2.1.3

kloczek:
With that this shell backed sritpt can be removed as well 😋 (as no longer needed)

haircommander:
I'd prefer to keep it for now as it's still used in the makefile :)

PR #342, opened 2022-06-06
add log-global-size-max option to limit the total output

haircommander:
related to https://github.com/cri-o/cri-o/pull/5933 and https://github.com/cri-o/cri-o/security/advisories/GHSA-fcm2-6c3h-pg6j


haircommander:
@saschagrunert @mrunalp @rhatdan @giuseppe PTAL

rhatdan:
Does global add any value to the option.  Would --logs-size-max work?

haircommander:
> Does global add any value to the option. Would --logs-size-max work?

we already have a log-size-max for the log rotation fields. the global is to differentiate the two

rhatdan:
--logs-size-max 
versus --log-size-max  The s to differentiate the difference.

haircommander:
ahh good idea. Unfortunately we've already shipped cri-o with the flag (and patches to conmon in openshift :see_no_evil: ) so it would end up being tricky to change it now

giuseppe:
@haircommander would it be a breaking change for CRI-O if we change the behavior of `--log-global-size-max` to delete the oldest log file instead of stopping writing so that we could reuse it for https://github.com/containers/podman/issues/23191?

giuseppe:
probably it is still cleaner to add a new option to specify the maximum number of files to keep around

haircommander:
I'm not actually sure. It's used to limit the amount of output on exec probe requests to mitigate a CVE, and we never really concretely promised any behavior on which output would be returned if the exec probe exceeded the max. This change would change the behavior so that we get the end of output instead of beginning. I think you could make an argument in either direction. ultimately, a user would hit this situation if their execs were using *way* too much output, so I am inclined to say not a breaking change? WDYT @saschagrunert @mrunalp @kwilczynski @sohankunkerkar

giuseppe:
yeah I had a quick look at it and I think it will be confusing. It is better a different option to control logs rotation

PR #343, opened 2022-06-15
Stop using g_unix_signal_add() to avoid threads

### LINKED ISSUE #330: Conmon hang after wake from sleep
On my two systems, one Silverblue 35 and one Silverblue 36, I sporadically get a conmon hang, using one core fully, right after waking from sleep. Usually happens in the morning.

```
$ conmon --version
conmon version 2.1.0
commit: 
```

Unsure what info to provide; if debugging is needed I do not mind trying to get a stack trace with debug symbols.

allisonkarlitskaya:
g_unix_signal_add() is implemented via an implicitly-created worker
thread in GLib.  This thread creates an eventfd, and conflicts with
conmon's general approach to taking full ownership of all open fds in
the process.

Stop using g_unix_signal_add() and switch to using g_unix_fd_add() on a
signalfd() as an alternative approach.

Fixes #330

I tested this by running
```
podman --conmon ~/src/conmon/build/conmon exec f36 sh -c 'sleep 1000&'
```

and

```
podman --conmon /usr/bin/conmon exec f36 sh -c 'sleep 1000&'
```

and suspending my laptop.  On resume, `/usr/bin/conmon` was spinning at 100% CPU, but my patched version wasn't.

### DIFF
--- a/src/ctr_exit.c
+++ b/src/ctr_exit.c
@@ -13,16 +13,12 @@
 #include <glib-unix.h>
 #include <signal.h>
 #include <stdlib.h>
+#include <sys/signalfd.h>
 #include <sys/prctl.h>
 #include <unistd.h>
 
 volatile pid_t container_pid = -1;
 volatile pid_t create_pid = -1;
-
-void on_sigchld(G_GNUC_UNUSED int signal)
-{
-	raise(SIGUSR1);
-}
 
 void on_sig_exit(int signal)
 {
@@ -83,9 +79,16 @@ gboolean check_child_processes_cb(gpointer user_data)
 	return G_SOURCE_REMOVE;
 }
 
-gboolean on_sigusr1_cb(gpointer user_data)
+gboolean on_signalfd_cb(gint fd, G_GNUC_UNUSED GIOCondition condition, gpointer user_data)
 {
 	struct pid_check_data *data = (struct pid_check_data *)user_data;
+
+	/* dequeue the signal from the signalfd */
+	struct signalfd_siginfo siginfo;
+	ssize_t s = read(fd, &siginfo, sizeof siginfo);
+	g_assert_cmpint(s, ==, sizeof siginfo);
+	g_assert_cmpint(siginfo.ssi_signo, ==, SIGCHLD);
+
 	check_child_processes(data->pid_to_handler, data->exit_status_cache);
 	return G_SOURCE_CONTINUE;
 }
--- a/src/conmon.c
+++ b/src/conmon.c
@@ -22,6 +22,7 @@
 #include "runtime_args.h"
 
 #include <sys/prctl.h>
+#include <sys/signalfd.h>
 #include <sys/stat.h>
 #include <locale.h>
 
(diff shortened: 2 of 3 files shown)


src/ctr_exit.c
```diff
@@ -83,9 +79,16 @@ gboolean check_child_processes_cb(gpointer user_data)
 	return G_SOURCE_REMOVE;
 }
 
-gboolean on_sigusr1_cb(gpointer user_data)
+gboolean on_signalfd_cb(gint fd, GIOCondition condition, gpointer user_data)
```
giuseppe, line 82: `condition` is not used and the CI barks about it (just prepend G_GNUC_UNUSED)

allisonkarlitskaya: Funny.  I had caught this locally as well, and changed it, but forgot to commit before pushing.  Fixed now :)

allisonkarlitskaya:
I've added the signed-off-by and fixed the unused parameter warning, as well as updating a comment to be more accurate, and adding an error check if `signalfd()` fails (using the same style as similar checks in the same function).

haircommander:
Thanks @allisonkarlitskaya I appreciate you digging in here :smiley: 

```
/usr/bin/ld: src/conmon.o: undefined reference to symbol 'pthread_sigmask@@GLIBC_2.2.5'
/usr/bin/ld: /lib/x86_64-linux-gnu/libpthread.so.0: error adding symbols: DSO missing from command line
```
is one failure in ci
maybe we need to include <pthread.h> too?

allisonkarlitskaya:
> Thanks @allisonkarlitskaya I appreciate you digging in here smiley
> 
> ```
> /usr/bin/ld: src/conmon.o: undefined reference to symbol 'pthread_sigmask@@GLIBC_2.2.5'
> /usr/bin/ld: /lib/x86_64-linux-gnu/libpthread.so.0: error adding symbols: DSO missing from command line
> ```
> 
> is one failure in ci maybe we need to include <pthread.h> too?
> 
> edit: or maybe add `-lpthread` when compiling...

manpage says

```
       #include <signal.h>

       int pthread_sigmask(int how, const sigset_t *set, sigset_t *oldset);

       Compile and link with -pthread.
```

The irony of the fact that I need to link with `-pthread` to avoid creating a thread is not lost on me.

allisonkarlitskaya:
So there's also `sigprocmask()` which is basically the same thing.

A bit more reading though, and I ran into this little gem:

```
       A child created via fork(2) inherits a copy of its parent's
       signal mask; the signal mask is  preserved  across execve(2).
```

This isn't fun anymore :sweat_smile:

haircommander:
LMK if you'd like one of the conmon devs could take it over :smiley:

allisonkarlitskaya:
> LMK if you'd like one of the conmon devs could take it over smiley

I'd be happy to get a bit of input as to if anyone thinks that this is a problem or not.  `conmon` spawns a lot of processes, and having `SIGCHLD` masked in those processes might have all kinds of strange side-effects.

This falls back into the category of "I don't really know enough about what this program is doing"...

giuseppe:
> `conmon` spawns a lot of processes, and having `SIGCHLD` masked in those processes might have all kinds of strange side-effects.

I think that is not a problem.  conmon spawns only two processes (the OCI runtime and the cleanup process).  I think it is safe to assume these programs will reset the signals.

allisonkarlitskaya:
> > `conmon` spawns a lot of processes, and having `SIGCHLD` masked in those processes might have all kinds of strange side-effects.
> 
> I think that is not a problem. conmon spawns only two processes (the OCI runtime and the cleanup process). I think it is safe to assume these programs will reset the signals.

Good enough for me.  I changed to `sigprocmask()` instead (which is already used from the same file).

haircommander:
LGTM, thanks for all of your work @allisonkarlitskaya . @giuseppe for the final approval?

giuseppe (approved):
LGTM

Jmennius:
Can we have this released?
It's been a month since this PR was merged, plus it will take time to got into the distros..
Thanks!

rhatdan:
We are preparing releases right now podman 4.2

PR #349, opened 2022-08-04
Use /usr/bin/env to locate bash

dfr:
On FreeBSD, this lives in /usr/local/bin. Note: this pattern is already
used for the other scripts in hack.


rhatdan:
LGTM

PR #352, opened 2022-08-25
signal: handle SIGUSR1 with signalfd

giuseppe:
we use internally SIGUSR1 as a hack to force the check of the
container processes.

commit d91cc4321797eaa84dcfb7863e91632d2fe26861 changed the way some
signals are handled but forgot to handle SIGUSR1, so that receiving
SIGUSR1 would crash conmon.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=2119072

### DIFF
--- a/src/conmon.c
+++ b/src/conmon.c
@@ -312,9 +312,9 @@ int main(int argc, char *argv[])
 		.pid_to_handler = pid_to_handler,
 		.exit_status_cache = NULL,
 	};
-	int signal_fd = get_signal_descriptor(SIGCHLD);
+	int signal_fd = get_signal_descriptor();
 	if (signal_fd < 0)
-		pexit("Failed to create signalfd for SIGCHLD");
+		pexit("Failed to create signalfd");
 	int signal_fd_tag = g_unix_fd_add(signal_fd, G_IO_IN, on_signalfd_cb, &data);
 
 	if (opt_exit_command)
--- a/src/ctr_exit.c
+++ b/src/ctr_exit.c
@@ -81,9 +81,8 @@ gboolean on_signalfd_cb(gint fd, G_GNUC_UNUSED GIOCondition condition, gpointer 
 {
 	struct pid_check_data *data = (struct pid_check_data *)user_data;
 
-	/* dequeue the signal from the signalfd */
-	int sig = dequeue_signal_event(fd);
-	g_assert_cmpint(sig, ==, SIGCHLD);
+	/* drop the signal from the signalfd */
+	drop_signal_event(fd);
 
 	check_child_processes(data->pid_to_handler, data->exit_status_cache);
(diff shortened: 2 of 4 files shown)


vrothberg (approved):
LGTM

@edsantiago PTAL

src/ctr_exit.c
```diff
@@ -82,8 +82,7 @@ gboolean on_signalfd_cb(gint fd, G_GNUC_UNUSED GIOCondition condition, gpointer
 	struct pid_check_data *data = (struct pid_check_data *)user_data;
 
 	/* dequeue the signal from the signalfd */
-	int sig = dequeue_signal_event(fd);
-	g_assert_cmpint(sig, ==, SIGCHLD);
+	dequeue_signal_event(fd);
```
edsantiago, line 85: Isn't this going to choke at compile time because of missing `(void)`?

edsantiago (commented):
It's been a lot of years since I dealt with signals in C. I don't feel comfortable with a quick review, and can't really devote today to it. One quick comment inline.

src/conmon.c
```diff
@@ -312,7 +312,7 @@ int main(int argc, char *argv[])
 		.pid_to_handler = pid_to_handler,
 		.exit_status_cache = NULL,
 	};
-	int signal_fd = get_signal_descriptor(SIGCHLD);
+	int signal_fd = get_signal_descriptor(SIGCHLD, SIGUSR1);
```
haircommander, line 315: I think I would rather hardcode these in `get_signal_descriptor` then pass two signals arbitrarily

src/conmon.c
```diff
@@ -312,7 +312,7 @@ int main(int argc, char *argv[])
...
 	if (signal_fd < 0)
 		pexit("Failed to create signalfd for SIGCHLD");
```
haircommander, line 317: nit: can we add SIGUSR1 to this log

src/ctr_exit.c
haircommander, line 85: nit: if we're not gonna use this value, then I think we should change the function signature

giuseppe: good idea.  Pushed a new version

src/conmon.c
giuseppe: we have two different versions of `get_signal_descriptor`.  If we want to move the list in `get_signal_descriptor` we would need a way to define them in a single place

src/ctr_exit.c
giuseppe: I've changed the function to be `void`

src/conmon.c
giuseppe: done

src/conmon.c
giuseppe: fixed in the last version

haircommander:
LGTM, thanks!

PR #354, opened 2022-08-29
Packit: initial enablement

lsm5:
This commit will run Copr builds on every PR against all active
releases of Fedora, thus allowing buildability checks before the
PR merges.

Signed-off-by: Lokesh Mandvekar <lsm5@fedoraproject.org>

EDIT: s/Koji scratch/Copr/

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -1,0 +1,44 @@
+# See the documentation for more information:
+# https://packit.dev/docs/configuration/
+
+upstream_package_name: conmon
+downstream_package_name: conmon
+
+jobs:
+  - job: copr_build
+    # Run on every PR
+    trigger: pull_request
+    # Defaults to x86_64 unless architecture is explicitly specified
+    targets:
+      - fedora-rawhide-aarch64
+      - fedora-rawhide-i386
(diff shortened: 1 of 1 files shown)


lsm5:
@edsantiago packit is largely happy here. Couple of jobs are in progress but they should finish soon. Don't think you deal with conmon usually, but this should be another good data point.

@haircommander PTAL :)

lsm5:
> Can't tell if i386/s390x are hung or just slow. Regardless, though, multiple test failures.

both jobs are successful, packit people said it's apparently some issue with the github API. Other test failures are likely something else. I'll rerun them.

haircommander:
LGTM thanks!

lsm5:
@haircommander do you know whatsup with the integration test failures? I'm surely not to blame this time :laughing:

lsm5:
@haircommander thanks a lot!

lsm5:
@haircommander btw, do let me know if this causes you any pain in other PRs, we're working on bringing in packit to our repos in stages. ATM, adding rpm build checks for fedora, maybe later even include gating tests and stuff, if things don't turn out to be a PITA.

PR #356, opened 2022-09-03
signal: Track changes to get_signal_descriptor in the FreeBSD version

dfr:
Since kqueue treats each signal as a separate event, we need to iterate
over the signal set and add events for each.


rhatdan:
LGTM
@giuseppe @haircommander PTAL

giuseppe (approved):
LGTM

PR #359, opened 2022-10-20
Makefile: Fix install for FreeBSD

dfr:
This avoids using install's -D flag since that does something different on FreeBSD. Also, since GNU make is called gmake on FreeBSD, we need to use $(MAKE) for sub-makes.


haircommander:
LGTM

Issue #315, opened 2022-01-10
segmentation fault with podman and ocrmypdf


Fulguritus:
I originally filed the issue in ocrmypdf: https://github.com/ocrmypdf/OCRmyPDF/issues/889

Essentially, when doing 
```bash
podman run --network none --rm -i jbarlow83/ocrmypdf:v13.2.0 --jobs 1 -l deu - - <tmp.pdf >out.pdf
```
the output hangs, and I see a segfault in conmon in dmesg: 
```
[24408.182063] conmon[100696]: segfault at a8a000 ip 00007f025e26d9b5 sp 00007ffcc9e773e8 error 4 in libc.so.6[7f025e11e000+176000]
[24408.182091] Code: fd 74 5f 41 c5 fd 74 67 61 c5 ed eb e9 c5 dd eb f3 c5 cd eb ed c5 fd d7 cd 85 c9 75 48 48 83 ef 80 48 81 ea 80 00 00 00 77 cb <c5> fd 74 4f 01 c5 fd d7 c1 66 90 85 c0 75 5c 83 c2 40 0f 8f c3 00
```

Address within libc: 0x00007f025e26d9b5 - 0x7f025e11e000 = 0x14F9B5
```
$ addr2line -e /usr/lib/libc.so.6 -fCi 0x14F9B5
__GI_netname2host
:?
```

Running the same with log-level=debug, I see
```
DEBU[0270] Sending signal 2 to container b2799f12bc64a5e78947be68ec1c9c37daaa0de03a47c4de4531e87f32d7551f 
2022-01-10T16:46:08.000913691Z: open pidfd: No such process
ERRO[0275] No exit file for container b2799f12bc64a5e78947be68ec1c9c37daaa0de03a47c4de4531e87f32d7551f found: timed out waiting for file /run/user/1000/libpod/tmp/exits/b2799f12bc64a5e78947be68ec1c9c37daaa0de03a47c4de4531e87f32d7551f: internal libpod error 
ERRO[0275] Error forwarding signal 2 to container b2799f12bc64a5e78947be68ec1c9c37daaa0de03a47c4de4531e87f32d7551f: error sending signal to container b2799f12bc64a5e78947be68ec1c9c37daaa0de03a47c4de4531e87f32d7551f: `/usr/bin/crun kill b2799f12bc64a5e78947be68ec1c9c37daaa0de03a47c4de4531e87f32d7551f 2` failed: exit status 1 
```
[...]

PR #361, opened 2022-11-03
logging: do not read more that the buf size

giuseppe:
msg_len doesn't take into account NUL bytes that could be present in
the buffer, while g_strdup_printf("MESSAGE=%s%s", partial_buf, buf)
does and would stop writing the string once it finds a NUL byte.  This
would lead to read after the buffer end.

Build the message string manually so that the calculated length
reflects the buffer size.

Closes: https://github.com/containers/conmon/issues/315

### DIFF
--- a/src/ctr_logging.c
+++ b/src/ctr_logging.c
@@ -286,21 +286,17 @@ static int write_journald(int pipe, char *buf, ssize_t buflen)
 			return 0;
 		}
 
-		/* sd_journal_* doesn't have an option to specify the number of bytes to write in the message, and instead writes the
-		 * entire string. Copying every line doesn't make very much sense, so instead we do this tmp_line_end
-		 * hack to emulate separate strings.
-		 */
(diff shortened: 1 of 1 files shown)


saschagrunert (approved):
LGTM

haircommander:
@giuseppe one lint issue https://github.com/containers/conmon/pull/361/checks?check_run_id=9269929279 otherwise LGTM, thanks for taking this on

evverx:
Thanks! I can't reproduce that issue anymore. I also wrote a (half-baked) fuzz target and it hasn't been able to find anything else either so far. Unfortunately OSS-Fuzz makes it extremely hard to add fuzz targets with external dependencies like `libsystemd` and `glib` so I'm not sure how to bring it there at this point without having to maintain `glib` and `systemd` builds. I have to run some a lot fuzz targets elsewhere because of that as well.

PR #370, opened 2023-01-19
packit: build in dedicated projects

TomasTomecek:
When "identifier" is defined, the job will be built in a dedicated Copr
project. This is a workaround for:

  https://github.com/packit/packit-service/issues/1854

All rpm builds should work consistently after merging this.


haircommander:
LGTM, thanks!

PR #371, opened 2023-01-31
Fix OOM watcher for cgroupv2 `oom_kill` events

saschagrunert:
It may be possible that the container process already got signaled but is still running, whereas we will not get an `oom` entry of `memory.events` but `oom_kill`, which provides another indicator for a possible out of memory kill.

Ref: https://www.kernel.org/doc/Documentation/cgroup-v2.txt

Fixes: https://github.com/cri-o/cri-o/issues/6580
Referring PR in conmon-rs: https://github.com/containers/conmon-rs/pull/1054

### DIFF
--- a/src/cgroup.c
+++ b/src/cgroup.c
@@ -279,19 +279,28 @@ gboolean check_cgroup2_oom()
 	ssize_t read;
 	while ((read = getline(&line, &len, fp)) != -1) {
 		long int counter;
+		const int oom_len = 4, oom_kill_len = 9;
 
-		if (read < 6 || memcmp(line, "oom ", 4))
+		if (read >= oom_kill_len + 2 && memcmp(line, "oom_kill ", oom_kill_len) == 0)
+			len = oom_kill_len;
+		else if (read >= oom_len + 2 && memcmp(line, "oom ", oom_len) == 0)
+			len = oom_len;
+		else
 			continue;
 
-		counter = strtol(&line[4], NULL, 10);
+		counter = strtol(&line[len], NULL, 10);
+
 		if (counter == LONG_MAX) {
-			nwarnf("Failed to parse %s", &line[4]);
+			nwarnf("Failed to parse: %s", &line[len]);
+			continue;
 		}
 
+		if (counter == 0)
+			continue;
+
 		if (counter != last_counter) {
-			if (write_oom_files() == 0) {
+			if (write_oom_files() == 0)
 				last_counter = counter;
-			}
 		}
 		return G_SOURCE_CONTINUE;
 	}


saschagrunert:
@containers/podman-maintainers @haircommander PTAL

src/cgroup.c
```diff
@@ -279,19 +279,35 @@ gboolean check_cgroup2_oom()
...
+		int memcmp_oom, memcmp_oom_kill, len;
+		const int oom_len = 4, oom_kill_len = 9;
 
-		if (read < 6 || memcmp(line, "oom ", 4))
+		if (read < 6)
 			continue;
 
-		counter = strtol(&line[4], NULL, 10);
+		memcmp_oom = memcmp(line, "oom ", oom_len);
+		memcmp_oom_kill = memcmp(line, "oom_kill ", oom_kill_len);
```
giuseppe, line 289: we risk to read after the line length.

The `read < 6` protects only for `oom `.

src/cgroup.c
```diff
@@ -279,19 +279,35 @@ gboolean check_cgroup2_oom()
 	ssize_t read;
 	while ((read = getline(&line, &len, fp)) != -1) {
 		long int counter;
+		int memcmp_oom, memcmp_oom_kill, len;
+		const int oom_len = 4, oom_kill_len = 9;
 
-		if (read < 6 || memcmp(line, "oom ", 4))
+		if (read < oom_len + 2 && read < oom_kill_len + 2)
```
giuseppe, line 285: wouldn't we skip to check `oom `, if the len is `<= 11`?

I think we need to change the check to be something like:

```
if (read >= oom_len + 2)
    memcmp_oom = memcmp(line, "oom ", oom_len);
if (read >= oom_kill_len + 2)
    memcmp_oom = memcmp(line, "oom_kill ", oom_len);
```

instead of the `continue` block.

saschagrunert: Yeah I guess we can merge the `memcmp` with the `read` len checks. Updated the paths a bit to

```c
		if (read >= oom_len + 2 && memcmp(line, "oom ", oom_len) == 0)
			len = oom_len;
		else if (read >= oom_kill_len + 2 && memcmp(line, "oom_kill ", oom_kill_len) == 0)
			len = oom_kill_len;
		else
			continue;
```

Thank you for the hint!

src/cgroup.c
```diff
@@ -279,19 +279,29 @@ gboolean check_cgroup2_oom()
...
 	while ((read = getline(&line, &len, fp)) != -1) {
 		long int counter;
-
-		if (read < 6 || memcmp(line, "oom ", 4))
+		int memcmp_oom, memcmp_oom_kill, len;
+		const int oom_len = 4, oom_kill_len = 9;
+
+		if (read >= oom_len + 2 && memcmp(line, "oom ", oom_len) == 0)
+			len = oom_len;
+		else if (read >= oom_kill_len + 2 && memcmp(line, "oom_kill ", oom_kill_len) == 0)
```
haircommander, line 287: this is a bit pedantic, but why don't we switch these cases? in the oom_kill case, we would always evaluate the first memcmp because `read >= oom_len + 2`, and if we switched them we wouldn't calculate memcmp twice

saschagrunert: Yes makes sense, switched the if cases.

src/cgroup.c
```diff
@@ -279,19 +279,29 @@ gboolean check_cgroup2_oom()
 	ssize_t read;
 	while ((read = getline(&line, &len, fp)) != -1) {
 		long int counter;
-
-		if (read < 6 || memcmp(line, "oom ", 4))
+		int memcmp_oom, memcmp_oom_kill, len;
```
haircommander, line 282: `memcmp_oom` and `memcmp_oom_kill` are never used, are they left over?

saschagrunert: Oh yeah, removed them as well as `len`, because we can re-use it from the lines above.

haircommander:
LGTM, good find here! @giuseppe PTAL

giuseppe (approved):
/lgtm

mimowo:
/retest
Do you know why the CI tests are failed, does it require some fix or rebasing? Trying to restart testing.

haircommander:
tests are in a bad shape, https://github.com/containers/conmon/pull/372 is trying to fix that. We don't even test cgroupv2 here though, so we can merge

saschagrunert:
@haircommander do we wanna release a new conmon version?

com6056:
Sorry to bump an old issue, looks like this is resolved as of https://github.com/containers/conmon/releases/tag/v2.1.6, can we get this library updated to pull in that fix?

PR #384, opened 2023-02-25
Fix leaking symbolic links in the opt_socket_path directory

aleskandro:
When we create a container with crio in OKD/OCP, conmon creates a symbolic link in /var/run/crio to /run/containers/storage/overlay-containers/$cuid/userdata.
This symblic link is never removed when the container is removed, leading to several dangling symbolic links in /var/run/crio, filling up the inodes capacity in /var/run.

This PR fixes this issue by adding a function/callback to remove the symbolic link when `conmon` exits.

Refers okd-project/okd#1497 and https://issues.redhat.com/browse/OCPBUGS-7962

### DIFF
--- a/src/ctr_exit.c
+++ b/src/ctr_exit.c
@@ -1,7 +1,7 @@
 #define _GNU_SOURCE
 
 #include "ctr_exit.h"
-#include "cli.h" // opt_exit_command, opt_exit_delay
+#include "cli.h" // opt_exit_command, opt_exit_delay, opt_socket_path, opt_cuuid
 #include "utils.h"
 #include "parent_pipe_fd.h"
 #include "globals.h"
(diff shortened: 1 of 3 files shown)


rhatdan:
LGTM

haircommander:
LGTM, thanks @aleskandro 

@giuseppe PTAL

src/ctr_exit.c
```diff
@@ -215,6 +215,16 @@ void reap_children()
 		;
 }
 
+void cleanup_socket_dir_symlink()
+{
+	/* A symbolic link might be created at {opt_socket_path}/{opt_cuuid} if the container
+	   is created successfully.
+	   This function will take care of removing the link when the conmon process exits. */
+	char *base_path = g_build_filename(opt_socket_path, opt_cuuid, NULL);
```
cgwalters, line 223: Happened to see this go by in my notifications;
https://github.com/containers/conmon/pull/387

aleskandro: I didn't do that as we're going to exit, but good point about the static analyzers 👍

PR #385, opened 2023-02-27
cgroup: Stumble on if we can't set up oom handling

cgwalters:
OpenShift 4.13 tentatively plans to use cgroupv1 on a RHEL9 era kernel, which breaks only in `PREEMPT_RT` aka `kernel-rt` due to
https://github.com/torvalds/linux/commit/2343e88d238f5de973d609d861c505890f94f22e

For now let's stumble on without oom notifications.


giuseppe (approved):
LGTM

haircommander:
seems to be a fmt error

cgwalters:
> seems to be a fmt error

Fixed

rhatdan:
Why cgroups V1 on RHEL 9?

PR #387, opened 2023-02-28
exit: Free allocated string

cgwalters:
This doesn't matter since we're going to exit, but I'm sure it will show up in static analyzers.


giuseppe (approved):
LGTM

rhatdan:
LGTM

PR #388, opened 2023-03-06
Use default umask `0022`

saschagrunert:
Under non-reproducible circumstances it can happen that the umask of conmon is set to `0000`, which can be a security issue. To avoid that, we now use a default umask and set it directly in `main`.

We had a similar fix in CRI-O in https://github.com/cri-o/cri-o/pull/5904, but we have to do it here again because conmon is a direct parent of PID 1.

Refers to https://issues.redhat.com/browse/OCPBUGS-8057


rhatdan:
LGTM
/approve

haircommander:
LGTM

PR #391, opened 2023-03-15
Switch go version to 1.19

dcermak:


kwilczynski:
@saschagrunert @haircommander, I think we should aim for 1.22 already here. Thoughts?

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/containers/conmon
 
-go 1.18
+go 1.19
```
saschagrunert, line 3: I think switching to go 1.22 is reasonable, right?

haircommander: agreed!

kwilczynski: While at it, I think we should get off from the github.com/pkg/errors package. Some of the @saschagrunert lint magic could help here a lot to identify old and obsolete things. :smile:

kolyshkin: The last suggestion is now implemented by #559

jnovy:
LGTM, we are actually a bit behind the current golang stable version so we might want to bump it again soon.

PR #392, opened 2023-03-16
oom: restore old OOM score

giuseppe:
restore the original OOM score before running the OCI runtime so that the container process will inherit the OOM score from the environment instead of the default value of 0.


giuseppe:
@haircommander it is needed for Podman but it could affect CRI-O workloads if CRI-O is running with the OOM score != 0

haircommander:
it looks like crio runs with 2 on openshift at least. I don't think that's too bad because the oci runtime will set the oom score for the actual container, so it'll just give an oom score to the runtime. WDYT @saschagrunert @mrunalp

saschagrunert (approved):
LGTM

giuseppe:
fine to merge?

PR #419, opened 2023-05-12
fix compile regression on FreeBSD

tendstofortytwo:
`attempt_oom_adjust()` takes two variables as of #392. Since both variables are used only under Linux and the compiler complains if you don't use the second one, this adds a line in to fix that.


emaste:
Great, thank you @tendstofortytwo.

Perhaps we can add a FreeBSD CI job in .cirrus.yml as well, to find these sorts of things automatically?

rhatdan:
LGTM

PR #427, opened 2023-06-06
Fix close_other_fds on FreeBSD

dfr:
While testing podman's restart policy handling, I noticed that containers which were publishing ports did not restart automatically. This was caused by conmon not properly closing the file descriptor which reserved the published port on the host before cleaning up the container. This commit lets conmon discover any extra file descriptors on FreeBSD.

### DIFF
--- a/src/close_fds.c
+++ b/src/close_fds.c
@@ -22,6 +22,12 @@
 
 #include <sys/stat.h>
 
+#ifdef __FreeBSD__
+#define OPEN_FILES_DIR "/dev/fd"
+#else
+#define OPEN_FILES_DIR "/proc/self/fd"
+#endif
+
 static int open_files_max_fd;
 static fd_set *open_files_set;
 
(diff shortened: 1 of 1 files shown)


dfr:
The failing tests look unrelated - maybe flakey?

haircommander:
LGTM test failures are unrelated

src/close_fds.c
```diff
@@ -22,6 +22,12 @@
 
 #include <sys/stat.h>
 
+#ifdef __FreeBSD__
+#define OPEN_FILES_DIR "/dev/fd"
+#else
+#define OPEN_FILES_DIR "/proc/self/fd"
```
kolyshkin, line 28: Sorry for being late to the party, but `/dev/fd` usually works on Linux, too, since it's a symlink to `/proc/self/fd` (at least that's the case with Fedora 38 on my laptop.

Update: this is done that way since systemd v12, here's the commit from 2010: https://github.com/systemd/systemd/commit/5c0532d1cc989d2f78d2cd4a18058daa58143705.

PR #428, opened 2023-06-06
Do not exit main loop before OOM check has finished

kolyshkin:
While debugging conmon OOM detection logic, I saw a case when OOM callback never finished, possibly due to g_main_loop_quit() being called.

Let it finish.

PS I'm not very proficient with this code (or glib), please review carefully.


rhatdan:
@mrunalp @haircommander @giuseppe PTAL

giuseppe:
isn't the event loop single threaded so that once it is in a handler it cannot perform any other action?

kolyshkin:
> isn't the event loop single threaded so that once it is in a handler it cannot perform any other action?

Maybe; alas this is not my area of expertise. Maybe the problem was that `container_exit_cb` was called before `oom_cb_cgroup_v2`.

Closing this PR; will continue investigating.

Issue #348, opened 2022-07-29
podman cannot run [conmon:e] Include journald in compilation path to log to systemd journal


b3n3w:
Hi there,

on one of our machines were not able to run any podman container due to the following error:

[conmon:e] Include journald in compilation path to log to systemd journal
Error: write child: broken pipe

I already reinstalled podman, conmon and checked the permissions of /var/log/journal.
When running as sudo the operation works without problem.

The user is already in the systemd-journal group

Debug output of podman: 
`DEBU[0000] mounted container "fb8f8ce087fb8f8d9fc8ef4aedb53db3e0fb8f98084a85525898743f83a1aef0" at "/home/user/.local/share/containers/storage/overlay/4d5225f292fbc66dff495c31621fcb05193e294a5b81d9188687d441ff88a140/merged" 
DEBU[0000] slirp4netns command: /usr/bin/slirp4netns --disable-host-loopback --mtu=65520 --enable-sandbox --enable-seccomp -c -e 3 -r 4 --netns-type=path /run/user/1000/netns/cni-f4ef9681-4344-7949-fa74-162c74242d8f tap0 
DEBU[0000] Created root filesystem for container fb8f8ce087fb8f8d9fc8ef4aedb53db3e0fb8f98084a85525898743f83a1aef0 at /home/user/.local/share/containers/storage/overlay/4d5225f292fbc66dff495c31621fcb05193e294a5b81d9188687d441ff88a140/merged 
DEBU[0000] /etc/system-fips does not exist on host, not mounting FIPS mode subscription 
DEBU[0000] Setting CGroups for container fb8f8ce087fb8f8d9fc8ef4aedb53db3e0fb8f98084a85525898743f83a1aef0 to user.slice:libpod:fb8f8ce087fb8f8d9fc8ef4aedb53db3e0fb8f98084a85525898743f83a1aef0 
DEBU[0000] reading hooks from /usr/share/containers/oci/hooks.d 
DEBU[0000] Workdir "/" resolved to host path "/home/user/.local/share/containers/storage/overlay/4d5225f292fbc66dff495c31621fcb05193e294a5b81d9188687d441ff88a140/merged" 
[...]

PR #431, opened 2023-06-14
nix: enable systemd in builds for journald support

haircommander:
should fix https://github.com/containers/conmon/issues/348


haircommander:
@saschagrunert have you any advice here?

PR #435, opened 2023-07-06
ensure console socket buffers are properly sized

dfr:
On FreeBSD, the default socket buffer size for unix domain sockets is just 8192 which is too small for largest message used on the console socket which is 8193.  This causes output for processes which emit large amounts of data to be lost on FreeBSD (for an example, see podman's 'podman exec/run - missing output' system test).

To avoid the problem, this commit increases the buffer sizes to CONN_SOCK_BUF_SIZE (32768) which allows for several max-sized messages to be in-flight.


dfr:
GIven that this is not a problem on Linux which defaults to much larger buffers, the calls to setsockopt could be made conditional for FreeBSD builds.

haircommander:
> GIven that this is not a problem on Linux which defaults to much larger buffers, the calls to setsockopt could be made conditional for FreeBSD builds.

yeah I think I would prefer that

rhatdan:
LGTM
@giuseppe @haircommander PTAL

haircommander:
LGTM

giuseppe:
LGTM

PR #441, opened 2023-08-10
stdio: ignore EIO for terminals

giuseppe:
when the terminal is closed, we get EIO.  Let's treat it as a EOF.

Closes: https://github.com/containers/conmon/issues/440

### DIFF
--- a/src/ctr_stdio.c
+++ b/src/ctr_stdio.c
@@ -129,6 +129,13 @@ static bool read_stdio(int fd, stdpipe_t pipe, gboolean *eof)
 			*eof = true;
 		return false;
 	} else if (num_read < 0) {
+		/* Ignore EIO if fd is a tty, since this can happen when the tty is closed
+		   while we are reading from it. */
+		if (errno == EIO && isatty(fd)) {
+			if (eof)
+				*eof = true;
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
@haircommander PTAL

haircommander:
you've an fmt error

src/ctr_stdio.c
```diff
@@ -129,6 +129,9 @@ static bool read_stdio(int fd, stdpipe_t pipe, gboolean *eof)
 			*eof = true;
 		return false;
 	} else if (num_read < 0) {
+		if (errno == EIO && isatty(fd)){
```
haircommander, line 132: really not important, but if we're returning the same value, we could invert this condition and put the warning in here instead. idk if that would make it hard to read though. I think either way, a comment would be helpful

giuseppe: I tried to invert it but it looks more difficult to read.

I've added a comment

edsantiago:
Is there any hope of this fixing https://github.com/containers/podman/issues/10927 ?

giuseppe:
I am not sure :/ the symptoms look very different

PR #444, opened 2023-08-13
Changes to build conmon for `riscv64`

michalbiesek:


haircommander:
@saschagrunert PTAL

saschagrunert (approved):
LGTM, the only difference between arm, ppc and riscv64 is the value of `crossSystem`, but we can cleanup that afterwards.

PR #456, opened 2023-09-25
Remove checks for (long)deprecated libsystemd-journal for libsystemd

rahilarious:
https://bugzilla.redhat.com/show_bug.cgi?id=1350301#c2

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,10 +38,7 @@ override CFLAGS += $(shell $(PKG_CONFIG) --cflags glib-2.0) -DVERSION=\"$(VERSIO
 # "pkg-config --exists" will error if the package doesn't exist. Make can only compare
 # output of commands, so the echo commands are to allow pkg-config to error out, make to catch it,
 # and allow the compilation to complete.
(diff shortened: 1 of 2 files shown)


rahilarious:
ping ... @rhatdan PTAL

rhatdan:
LGTM
@haircommander @mrunalp @giuseppe PTAL

haircommander:
hm I'm pretty sure this and the associated deprecation will break journald driver support in conmon... I think we need to find the package to replace it

giuseppe:
> hm I'm pretty sure this and the associated deprecation will break journald driver support in conmon... I think we need to find the package to replace it

libsystemd should already have the needed symbols, why wouldn't this be enough?

giuseppe:
@rahilarious do we still need this?

rahilarious:
> @rahilarious do we still need this?

you're right. libsystemd has everything. Go may proceed ahead remove libsystemd-journal

giuseppe (approved):
LGTM

@haircommander WDYT?

PR #462, opened 2023-11-01
src/seccomp_notify.c: fix static build

ffontaine:
Fix the following static build failure raised even when seccomp is disabled:

```
src/seccomp_notify.c:10:10: fatal error: dlfcn.h: No such file or directory
   10 | #include <dlfcn.h>
      |          ^~~~~~~~~
```

Fixes:
 - http://autobuild.buildroot.org/results/71b4f35b3150183c7b44bc3897f01b0019e10ebe

### DIFF
--- a/src/seccomp_notify.c
+++ b/src/seccomp_notify.c
@@ -7,7 +7,6 @@
 
 #include <errno.h>
 #include <sys/ioctl.h>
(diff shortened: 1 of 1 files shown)


haircommander:
LGTM thanks!

saschagrunert (approved):

PR #464, opened 2023-11-23
oom-score: restore oom score before running exit command

giuseppe:
When the exit command is executed, the oom score is restored to its original value otherwise the command runs with -1000.

Closes: https://github.com/containers/podman/issues/20765

### DIFF
--- a/src/conmon.c
+++ b/src/conmon.c
@@ -43,7 +43,6 @@ int main(int argc, char *argv[])
 	_cleanup_gerror_ GError *err = NULL;
 	char buf[BUF_SIZE];
 	int num_read;
-	int old_oom_score = 0;
 	_cleanup_close_ int dev_null_r_cleanup = -1;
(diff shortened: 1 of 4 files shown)


rhatdan:
LGTM

giuseppe:
@haircommander PTAL

src/ctr_logging.c
```diff
@@ -614,6 +616,10 @@ static void reopen_k8s_file(void)
 	if (k8s_log_fd < 0)
 		pexitf("Failed to open log file %s", k8s_log_path);
 
+	unlink(k8s_log_path_tmp_old);
+	link(k8s_log_path, k8s_log_path_tmp_old);
+	rename(k8s_log_path_tmp_old, k8s_log_path_old);
```
haircommander, line 621: what's this piece?

giuseppe: shouldn't be there :/

removed and pushed a new version

haircommander:
LGTM

PR #465, opened 2023-11-24
logging: passthrough accepts a tty

giuseppe:
it works in a similar way to passthrough, the only difference is that it is also allowed on a TTY.

### DIFF
--- a/src/ctr_logging.c
+++ b/src/ctr_logging.c
@@ -191,9 +191,6 @@ static void parse_log_path(char *log_config)
 	}
 
 	if (!strcmp(driver, "passthrough")) {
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

haircommander:
what if we had conmon set this up if `-t` is set with `-l passthrough`? why the separate option?

giuseppe:
are you fine with the separate option or should I reuse "passthrough"?

haircommander:
I still lean towards passthrough, as the behavior on conmon side is the same. I think it could be up to podman to warn the user about the risks. WDYT?

PR #466, opened 2023-11-30
src: open all files with O_CLOEXEC

giuseppe:
so they are not propagated to the children.


giuseppe:
@haircommander PTAL

rhatdan:
LGTM

PR #468, opened 2023-12-04
logging: Respect log-size-max immediately after open

martinetd:
When a container keeps restarting, conmon restarts and opens the log file with O_APPEND every single time, resetting bytes_written to 0 with a non-empty file.
If the container fails before writting log-size-max bytes then the log keeps growing indefinitely, potentially using up all the available space.

This should be kept in check: initialize bytes_written to the file's current size, so that it can be reset if required (it's possible that the file gets re-open before the first write if it was already above the limit, but that should almost never happen in practice)

In order to make this simpler make bytes written global variables

---- 
This bit us with podman recently with a bad container config and restart=on-failure; I thought it wasn't a problem before? but checking the code I don't see how this would be a new problem, so I probably failed to check this when adding the podman `--log-opt max-size=xxx` parameter...

Thanks!

### DIFF
--- a/src/ctr_logging.c
+++ b/src/ctr_logging.c
@@ -3,6 +3,7 @@
 #include "cli.h"
 #include "config.h"
 #include <string.h>
+#include <sys/stat.h>
 
 // if the systemd development files were found, we can log to systemd
 #ifdef USE_JOURNALD
@@ -40,6 +41,8 @@ static int64_t log_global_size_max = -1;
 /* k8s log file parameters */
 static int k8s_log_fd = -1;
 static char *k8s_log_path = NULL;
+static int64_t k8s_bytes_written;
+static int64_t k8s_total_bytes_written;
 
 /* journald log file parameters */
 // short ID length
@@ -113,6 +116,15 @@ void configure_log_drivers(gchar **log_drivers, int64_t log_size_max_, int64_t l
 		k8s_log_fd = open(k8s_log_path, O_WRONLY | O_APPEND | O_CREAT | O_CLOEXEC, 0600);
 		if (k8s_log_fd < 0)
 			pexit("Failed to open log file");
+
+		struct stat statbuf;
+		if (fstat(k8s_log_fd, &statbuf) == 0) {
+			k8s_bytes_written = statbuf.st_size;
+		} else {
+			nwarnf("Could not stat log file %s, assuming 0 size", k8s_log_path);
+			k8s_bytes_written = 0;
+		}
+		k8s_total_bytes_written = k8s_bytes_written;
 
 		if (!use_journald_logging && tag)
 			nexit("k8s-file doesn't support --log-tag");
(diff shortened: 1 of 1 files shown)


src/ctr_logging.c
```diff
@@ -40,6 +41,8 @@ static int64_t log_global_size_max = -1;
 /* k8s log file parameters */
 static int k8s_log_fd = -1;
 static char *k8s_log_path = NULL;
+int64_t k8s_bytes_written;
+int64_t k8s_total_bytes_written;
```
haircommander, line 45: let's keep these static as they're only used in this file

src/ctr_logging.c
```diff
@@ -114,6 +117,15 @@ void configure_log_drivers(gchar **log_drivers, int64_t log_size_max_, int64_t l
...
 			pexit("Failed to open log file");
 
+		struct stat statbuf;
+		if (fstat(k8s_log_fd, &statbuf) == 0) {
+			k8s_bytes_written = statbuf.st_size;
+		} else {
+			nwarnf("Could not stat log file %s, assuming 0 size", k8s_log_path);
+			k8s_bytes_written = 0;
+		}
+		k8s_total_bytes_written = k8s_bytes_written;
```
haircommander, line 127: why not write directly to the variable?

haircommander:
couple of nits, feel free to ignore the test failures :)

src/ctr_logging.c
martinetd: Both variables are used for different purposes so both need to be set -- did you mean something like `k8s_total_bytes_written = k8s_bytes_written = statbuf.st_size` ?

src/ctr_logging.c
```diff
@@ -606,6 +614,9 @@ static void reopen_k8s_file(void)
 	/* Close the current k8s_log_fd */
 	close(k8s_log_fd);
 
+	/* Open with O_TRUNC: reset bytes written */
+	k8s_bytes_written = 0;
```
haircommander, line 618: wouldn't we want to reset *total_* as well here? Otherwise, the file rotation will rotate but won't be written to again

martinetd: The current code doesn't so I didn't think very hard about it -- from my understanding of the current code, `total_` indeed isn't reset, so logging stops as soon as the container has output log_global_size_max bytes (this patch makes it "minus the original file size" as I thought using just log global size max alone ought to limit the log size as well without rotating)
in other words:
 - log_size_max is used to cap the file size, rotating as necessary
 - log_global_size_max either caps the file size if used alone or limits the number of times the file is rotated

Looking at the commit that introduced it - 98058b14d9b1a2d6ecaaf0f26e4c1dde5e32865a "add log-global-size-max option to limit the total output conmon processes" - it might make more sense to not init k8s_total_bytes_written to the original file size, but then a user with only that option would risk the file growing indefinitely when container restarts so I'm not sure that's intended?
Either way we probably shouldn't just reset the total bytes written, if we do both vars would just be identical all the time and could be folded back together.

haircommander: ah sorry it's been a while since I've looked here, I forgot there were two cli flags for each situation.

haircommander:
LGTM, @giuseppe PTAL

giuseppe (approved):
LGTM

giuseppe:
could you please rebase so the CI (will hopefully) pass?

martinetd:
> could you please rebase so the CI will (hopefully) pass?

It looks like that did the trick.

I don't suppose there are plans for a 2.1.10 a couple of days after the 2.1.9 is there? :)

giuseppe (approved):

haircommander:
this is included in 2.1.10 :)

Issue #475, opened 2023-12-14
conmon 2.1.9 not working with podman 4.8.2 and ubuntu 22.04


dmdx86:
```
root@nas:~/conmon-2.1.8# cd ../conmon-2.1.9/
root@nas:~/conmon-2.1.9# make install
<snip>
root@nas:~/conmon-2.1.9# podman run --rm -it ubi9-init /bin/bash
Error: container create failed (no logs from conmon): conmon bytes "": readObjectStart: expect { or n, but found , error found in #0 byte of ...||..., bigger context ...||...
```

Same error on any other container, new or existing.

It is OK on 2.1.8

```
root@nas:~/conmon-2.1.9# cd ../conmon-2.1.8/
root@nas:~/conmon-2.1.8# make install
<snip>
root@nas:~/conmon-2.1.8# podman run --rm -it ubi9-init /bin/bash
[root@e9ded0c5b94a /]# 
exit
root@nas:~/conmon-2.1.8# 
```

I compile and build podman/conmon/crun/buildah, do not use the versions in the Ubuntu repositories. Everything was good up until I updated conmon this morning.

This is trivial for me to reproduce, LMK if anything else is needed from me or if I am doing something dumb to cause this.

```
root@nas:~/conmon-2.1.8# podman version
Client:       Podman Engine
Version:      4.8.2
API Version:  4.8.2
Go Version:   go1.20.3
Built:        Thu Dec 14 16:53:00 2023
OS/Arch:      linux/amd64
```

flyingfishflash:
I logged to say the same. I'm on archlinux, but all of my containers now fail to start with the same exact error.

Reverting back to 2.1.8 with podman 4.8.2 works as expected.

```
Client:       Podman Engine
Version:      4.8.2
API Version:  4.8.2
Go Version:   go1.21.5
Git Commit:   aa546902fa1a927b3d770528565627d1395b19f3-dirty
Built:        Wed Dec 13 17:07:26 2023
OS/Arch:      linux/amd64
```

martinetd:
hit the same problem when trying to check my patch just now...
The problem was introduced in [a recent static analysis fix](https://github.com/containers/conmon/pull/474) ; the free is incorrect if the value comes straight from options.

I've fixed it in https://github.com/containers/conmon/pull/476

PeterUpfold:
I can also reproduce this issue on Arch.

```
$ podman run -t ubuntu echo "Hello"

[...]

PR #476, opened 2023-12-15
Fix incorrect free in conn_sock

martinetd:
Earlier commit freed socket_parent_dir()'s result which is correct in the case it returns a path from g_build_filename, but when it returns opt_bundle_path the string should not be freed.

Make the function always return an allocated string that can be freed

Fixes: #475
Fixes: fad6bac8e65f ("fix some issues flagged by SAST scan")


martinetd:
(alternatively we could only free parent_dir if (!use_full_attach_path) in bind_unix_socket, but that seems more error-prone)

Luap99:
@haircommander @giuseppe PTAL

giuseppe (approved):
LGTM

PR #484, opened 2024-01-08
drop --tty on exec

haircommander:
as gvisor doesn't use it, and runc/crun don't need it, as they'll only use it if the process args aren't defined, which conmon always does

fixes https://github.com/containers/conmon/issues/483


giuseppe (approved):
LGTM

rhatdan:
LGTM

PR #492, opened 2024-02-28
Add support for s390x

saschagrunert:
Adding support for s390x via nix to ship the binary as part of the CRI-O packages.


saschagrunert:
We hit https://github.com/systemd/systemd/issues/30448 here.

PR #505, opened 2024-04-30
Bump version to 2.1.11

saschagrunert:
I forgot to bump the version to the latest release :-/ 

Should we create a follow-up tag because of that?


haircommander:
Should we bump again so we can rebuild without replacing the artifacts ? (2.1.12 now)

saschagrunert:
> Should we bump again so we can rebuild without replacing the artifacts ? (2.1.12 now)

Yes, done :+1:

PR #506, opened 2024-05-06
Packit: enable downstream sync to CentOS Stream 10

lsm5:
This commit enables downstream syncing to CentOS Stream 10. The centos maintainer will need to run `packit propose-downstream` and `centpkg build` until better centos integration is in place.

The copr jobs have been split distro-wise so that Fedora and CentOS builds / tests can still be triggered automatically regardless of PR author's access to the repo, while only the RHEL jobs will need manual triggering.

EL8 jobs have also been deleted as CentOS Stream 8 will go EOL end of May and no further updates are expected to be shipped there.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -2,13 +2,24 @@
 # See the documentation for more information:
 # https://packit.dev/docs/configuration/
 
-specfile_path: rpm/conmon.spec
+downstream_package_name: conmon
 upstream_tag_template: v{version}
+
+packages:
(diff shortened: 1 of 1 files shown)


lsm5:
@haircommander @saschagrunert PTAL

lsm5:
> /lgtm

@saschagrunert maybe it needs a manual merge given the 2 integration tests failed?

haircommander:
actually conmon never got prow support :)

PR #512, opened 2024-07-04
logging: remove unuseful fsync

giuseppe:
the file is going to be removed few lines later with rename():

```
	/* Replace the previous file */
	if (rename(k8s_log_path_tmp, k8s_log_path) < 0) {
		pexit("Failed to rename log file");
	}
```

so there is no point in the expensive sync call since the file is not supposed to be persisted after a reboot.


saschagrunert (approved):

giuseppe:
@haircommander PTAL

haircommander:
we did hit issues in the past removing an fsync but I think it was because of the other one before the container exits. this should be okay

Issue #504, opened 2024-04-29
conmon writes oom file to cwd


Luap99:
I have no idea if there is a reason for it but this seems rather strange behavior, why do we write the oom file to the cwd? This clutters the working directory of users with a random oom file in case of a oom container kill which makes no sense to a end user.

https://github.com/containers/conmon/blob/e21e7c85b7637e622f21c57675bf1154fc8b1866/src/cgroup.c#L331-L334

For podman something simple as
`podman run --rm --memory 1m --oom-score-adj 1000 quay.io/libpod/testimage:20240123 sort /dev/urandom` will reproduce the issue and writes it the cwd as podman does not change the cwd for conmon.

I tried to lookup the history and chased this all the way back to the inial commit, then continued in crio and eventually found https://github.com/cri-o/cri-o/pull/535/commits/7700a62347956ba634938ff91cd88cb195762291

I am not sure did cri-o set a different cwd for each conmon instance so cri-o could just read this without conflicts?

IMO I like to remove this assuming cri-o does no depend on it? Using the persistent path should be better?
Alternatively do we have to change the cwd in podman before launching conmon? Not sure if this breaks anything.

giuseppe:
seems weird to create it twice, I've opened a PR: https://github.com/containers/conmon/pull/513

PR #513, opened 2024-07-04
cgroup: do not write twice the oom file

giuseppe:
if a persist_path was provided, then use only that to write the oom file instead of creating it also in the current working directory.

Closes: https://github.com/containers/conmon/issues/504


Luap99 (commented):
Thanks, LGTM

I find it weird that the function is named write and that the error says write too but it really just creates an empty file, but this is pre-existing so I don't mind.

giuseppe:
> I find it weird that the function is named write and that the error says write too but it really just creates an empty file, but this is pre-existing so I don't mind.

good idea, I've renamed it to `create_oom_file`

giuseppe:
@haircommander PTAL

giuseppe:
maybe we can fix it in Podman, and run the conmon process from the bundle path itself, or the persistent dir

giuseppe:
alternative fix for Podman: https://github.com/containers/podman/pull/23243

PR #514, opened 2024-07-10
conmon: do not create oom file under cwd

giuseppe:
instead use the bundle path to create the second (shurgh!) file since this is what CRI-O uses.

Closes: https://github.com/containers/conmon/issues/504

### DIFF
--- a/src/cgroup.c
+++ b/src/cgroup.c
@@ -33,7 +33,8 @@ static void setup_oom_handling_cgroup_v2(int pid);
 static void setup_oom_handling_cgroup_v1(int pid);
 static gboolean oom_cb_cgroup_v2(int fd, GIOCondition condition, G_GNUC_UNUSED gpointer user_data);
 static gboolean oom_cb_cgroup_v1(int fd, GIOCondition condition, G_GNUC_UNUSED gpointer user_data);
-static int write_oom_files();
+static int create_oom_files();
+static int create_oom_file(const char *base_path);
 
 void setup_oom_handling(int pid)
 {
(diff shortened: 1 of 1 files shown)


src/cgroup.c
```diff
@@ -306,33 +306,39 @@ gboolean check_cgroup2_oom()
...
- * this can be used for v1 and v2 OOMS
+static int create_oom_file(const char *base_path, const char *path_type)
+{
+	if (base_path == NULL || base_path[0] == '\0')
+		return 0;
+
+	_cleanup_free_ char *ctr_oom_file_path = g_build_filename(base_path, "oom", NULL);
+	_cleanup_close_ int ctr_oom_fd = open(ctr_oom_file_path, O_CREAT | O_CLOEXEC, 0666);
+	if (ctr_oom_fd < 0) {
+		nwarnf("Failed to write oom file to the %s path", path_type);
```
Luap99, line 325: If I read such a message using `ctr_oom_file_path` would make the most sense as a user is likely not knowing where this path would be and if there is a full file path a user could check this directory quickly.

Luap99 (commented):
one nit, otherwise LGTM

src/cgroup.c
giuseppe: replaced `path_type` with `base_path` itself

Luap99 (commented):
LGTM

src/cgroup.c
```diff
@@ -306,33 +306,39 @@ gboolean check_cgroup2_oom()
...
 				last_counter = counter;
 		}
 		return G_SOURCE_CONTINUE;
 	}
 	return G_SOURCE_REMOVE;
 }
 
-/* write the appropriate files to tell the caller there was an oom event
- * this can be used for v1 and v2 OOMS
+static int create_oom_file(const char *base_path)
```
haircommander, line 317: nit: can this be defined underneath where it's called?

giuseppe: done

haircommander:
let's try to get ci running to double check this works with crio https://github.com/containers/conmon/pull/515

haircommander:
@giuseppe can you rebase to get test fixes please?

giuseppe:
rebased

haircommander:
thanks!

PR #524, opened 2024-09-02
RPM: Remove crio

lsm5:


lsm5:
@saschagrunert @haircommander do we need the `Packaging for Fedora ...` cirrus jobs if we have packit?

saschagrunert:
@lsm5 I don't think so.

mdbooth (commented):
IIUC this meets the packaging requirements, namely not to touch `%{_libexecdir}/crio`, which is owned by crio. If the podman tests still pass, which from the discussion we seem confident they will, this looks good to me.

lsm5:
> IIUC this meets the packaging requirements, namely not to touch `%{_libexecdir}/crio`, which is owned by crio. If the podman tests still pass, which from the discussion we seem confident they will, this looks good to me.

Could you please point to the exact `podman tests` in this repo? We can and probably should replace any Cirrus tests with TMT.

lsm5:
PTAL.

mdbooth:
> > IIUC this meets the packaging requirements, namely not to touch `%{_libexecdir}/crio`, which is owned by crio. If the podman tests still pass, which from the discussion we seem confident they will, this looks good to me.
> 
> Could you please point to the exact `podman tests` in this repo? We can and probably should replace any Cirrus tests with TMT.

Sorry, this was just my hand-wavy assumption that they must exist somewhere and will be executed here. I believe the primary concern about this change is that it could break podman (but we don't believe it will because podman is using a different conmon binary).

I did notice that there appear to be some podman tests in the Fedora repo here: https://src.fedoraproject.org/rpms/conmon/blob/rawhide/f/tests. I don't know what they do, though.

lsm5:
> > > IIUC this meets the packaging requirements, namely not to touch `%{_libexecdir}/crio`, which is owned by crio. If the podman tests still pass, which from the discussion we seem confident they will, this looks good to me.
> > 
> > 
> > Could you please point to the exact `podman tests` in this repo? We can and probably should replace any Cirrus tests with TMT.
> 
> Sorry, this was just my hand-wavy assumption that they must exist somewhere and will be executed here. I believe the primary concern about this change is that it could break podman (but we don't believe it will because podman is using a different conmon binary).
> 
> I did notice that there appear to be some podman tests in the Fedora repo here: https://src.fedoraproject.org/rpms/conmon/blob/rawhide/f/tests. I don't know what they do, though.

Ack. I guess we can leave TMT enablement and cirrus removal for another PR. This PR should be good to go.

saschagrunert (approved):
@haircommander PTAL

mdbooth:
Awesome, thanks! @lsm5 Any idea which version of Fedora this is likely to land in?

lsm5:
> Awesome, thanks! @lsm5 Any idea which version of Fedora this is likely to land in?

@saschagrunert @haircommander is it worth cutting a new conmon for this? If not, I can add it manually.

mdbooth:
> > Awesome, thanks! @lsm5 Any idea which version of Fedora this is likely to land in?
> 
> @saschagrunert @haircommander is it worth cutting a new conmon for this? If not, I can add it manually.

To reiterate my use case: I cannot switch to the k8s cri-o packages until I update my base OS (Fedora CoreOS) to a release including this change.

lsm5:
> To reiterate my use case: I cannot switch to the k8s cri-o packages until I update my base OS (Fedora CoreOS) to a release including this change.

@mdbooth updates created https://bodhi.fedoraproject.org/updates/?search=conmon .

PR #529, opened 2024-09-20
vendor: drop libpod in runner package

haircommander:
initially the hope of the runner package was to be a library that podman and cri-o used, but that never happened. now, it's just for tests. As such, we can use different errors and reduce the vendor surface area of conmon


rhatdan:
LGTM

runner/conmon/pipes.go
```diff
@@ -9,10 +9,24 @@ import (
...
 
-	"github.com/containers/podman/v4/libpod/define"
 	"github.com/pkg/errors"
 )
 
+// These errors are adapted from github.com/containers/podman:libpod/define
+// And copied to reduce the vendor surface area of this library
+var (
+	// ErrInternal indicates an internal library error
+	ErrInternal = fmt.Errorf("internal libpod error")
```
kwilczynski, line 19: Would it make sense to rename the error message to:

```suggestion
	ErrInternal = fmt.Errorf("internal error")
```

PR #532, opened 2024-10-08
Packit: constrain downstream jobs to the fedora package

lsm5:
Without this constraint, packit triggers dup jobs for downstream updates.


jnovy:
LGTM, thanks.

Issue #454, opened 2023-09-08
podman with healthcheck fills journal with error


HerrFrutti:
# problem
My containers are healthy but my journal gets filled with error messages. The error seems to be related with healthchecks.

# workaround
For now I disabled all healthchecks and the error disappeared.

# logs
example line from my journal:
```
conmon[11328]: conmon 62de48d99d5858ab629f <error>: Unable to send container stderr message to parent Broken pipe
```
![266649948-aee6269b-69e2-4ec6-846c-6f215ba270eb](https://github.com/containers/conmon/assets/6975702/cf295dcb-fa4a-4c99-9260-ffd9b8fb894d)

# additional info
```
conmon version 2.1.7
podman version 4.6.1
podman-compose version 1.0.6
OS: Fedora release 37 (Thirty Seven)
```
I'm using podman-compose to deploy my containers.

hdonnay:
This is still a problem on Fedora 39

mufeedali:
I have a couple of health checks and memory usage for journald is way beyond normal due to this. Until this bug gets fixed, I might have to disable my health checks, which is kind of sad :(

benblasco:
@mufeedali How have you been disabling the health checks?

shyzus:
@benblasco If you are dealing with images like the gitlab image which provides health checks by default. You can use this [option](https://docs.podman.io/en/latest/markdown/podman-create.1.html#health-cmd-command-command-arg1) to disable it `--health-cmd=none`

stryan:
I've been noticing this problem on my OpenSUSE Leap 15.5 box. Containers are run through quadlet with podman 4.7.1.

symysak:
Ubuntu 22.04 LTS 64-bitでもこの問題が発生しています。
raspberrypi@raspberrypi:~$ podman -v
podman version 4.6.2

benblasco:
> @benblasco If you are dealing with images like the gitlab image which provides health checks by default. You can use this [option](https://docs.podman.io/en/latest/markdown/podman-create.1.html#health-cmd-command-command-arg1) to disable it `--health-cmd=none`

Thanks for sharing this info. I will have to figure out how to do this with quadlet!

benblasco:
Going through the documentation it appears as though the `healthcmd` directive can only be used for container units, and not pod units. Is there a way to override the healthcmd for a pod unit? Am I just reading the documentation incorrectly?

rhatdan:
[...]

PR #536, opened 2024-12-06
Avoid bogus journal filling errors

jnovy:
Issue #454 is likely caused by retryable_error() not catching all possible retryable errors from the write(2) syscall.

This PR should be FeeBSD compatible. The only retriable result in question is ENOBUFS which might be bound to network latencies. Hope these are acceptable.

Signed-off-by: Jindrich Novy <jnovy@redhat.com>


jnovy:
@haircommander @saschagrunert PTAL

PR #538, opened 2024-12-26
Tmt

lsm5:


lsm5:
We can speed up tmt tests by only running the podman tests relevant to conmon. Let me know which ones and I can adjust the test config.

lsm5:
@saschagrunert @haircommander With packit rpm build jobs in place, do we need the cirrus `Packaging for Fedora N[-1]` jobs? If no, I guess we could also delete `contrib/` and `.rpmbuild` dirs along with those 2 jobs.

lsm5:
Marking ready for review but maybe best to wait until https://github.com/containers/podman/pull/24369 is merged. I'll then update the TMT plan to fetch tests from podman main branch instead of my own fork.

saschagrunert:
> @saschagrunert @haircommander With packit rpm build jobs in place, do we need the cirrus `Packaging for Fedora N[-1]` jobs? If no, I guess we could also delete `contrib/` and `.rpmbuild` dirs along with those 2 jobs.

I'd say no, but I'm not exactly sure how we should handle the CI here.

lsm5:
> > @saschagrunert @haircommander With packit rpm build jobs in place, do we need the cirrus `Packaging for Fedora N[-1]` jobs? If no, I guess we could also delete `contrib/` and `.rpmbuild` dirs along with those 2 jobs.
> 
> I'd say no, but I'm not exactly sure how we should handle the CI here.

Doesn't look like any Cirrus job depends on `fedora_packaging`. So, I think it should be safe to remove. I can leave that for a followup though.

The current PR should be good to go. PTAL. I'll leave CentOS and aarch64 for a followup as well.

lsm5:
Ah, there's still the fact that I'm fetching tests from my podman fork. I'll leave it to you if you feel comfy merging with that. I can keep my fork active and change the URL here once the podman PR is merged. Your call.

saschagrunert:
> @saschagrunert @haircommander How much does conmon care about nix package builds? Do we still need to keep the static build jobs?

Yeah I would prefer keeping them because we still ship conmon this way.

PR #541, opened 2025-01-09
Change permissions of logs from 0600 to 0640

jnovy:
Changing log permissions to 0640 would allow the administrator to set sticky group on the log directory, and for a selected log-users (in a specific group) without root-permissions to read the log files.

Fixes #539


haircommander:
LGTM

PR #543, opened 2025-01-15
Timestamp fix

jnovy:
Make timestamp generation never fail.
    
If, for whatever reason, quering current time fails then the whole log message is dropped.

This PR assures the log message is not lost and the output is done with the default timestamp.


jnovy:
@haircommander PTAL

jnovy:
Fair enough - there's no need to set it twice - just left the zero initialization @haircommander PTAL again.

haircommander:
can you run `make fmt` please?

jnovy:
Done @haircommander

PR #547, opened 2025-02-24
Install some packages to fix CI

bitoku:
Imported from https://github.com/cri-o/cri-o/blob/release-1.31/scripts/github-actions-packages


haircommander:
thank you!

Issue #490, opened 2024-02-02
conmon c82ee1934e559bd5868f <nwarn>: stdio_input read failed Resource temporarily unavailab


feidasheng:
[root@node-1 ~]# journalctl -xe  podman-coaster-celery-agent.service 
Failed to add match 'podman-coaster-celery-agent.service': Invalid argument
[root@node-1 ~]# journalctl -xe -u podman-coaster-celery-agent.service 
-- 
-- podman-coaster-celery-agent.service 单元已结束启动。
-- 
-- 启动结果为“done”。
2月 02 15:46:57 node-1.domain.tld systemd[1]: Starting Celery Service via podman...
-- Subject: podman-coaster-celery-agent.service 单元已开始启动
-- Defined-By: systemd
-- Support: https://lists.freedesktop.org/mailman/listinfo/systemd-devel
-- 
-- podman-coaster-celery-agent.service 单元已开始启动。
2月 02 15:46:59 node-1.domain.tld podman[51690]: celery multi v3.1.19 (Cipater)
2月 02 15:46:59 node-1.domain.tld podman[51690]: > Starting nodes...
2月 02 15:46:59 node-1.domain.tld podman[51690]:         > celery_1@celery: OK
2月 02 15:46:59 node-1.domain.tld conmon[51722]: conmon c82ee1934e559bd5868f <nwarn>: stdio_input read failed Resource temporarily unavailab>
2月 02 15:46:59 node-1.domain.tld conmon[51722]: conmon c82ee1934e559bd5868f <nwarn>: stdio_input read failed Resource temporarily unavailab>
2月 02 15:47:00 node-1.domain.tld podman[51818]: celery multi v3.1.19 (Cipater)
2月 02 15:47:00 node-1.domain.tld podman[51818]: > celery_1@node-1.domain.tld: DOWN
2月 02 15:47:00 node-1.domain.tld systemd[1]: podman-coaster-celery-agent.service: Succeeded.
-- Subject: Unit succeeded
-- Defined-By: systemd
-- Support: https://lists.freedesktop.org/mailman/listinfo/systemd-devel
-- 
-- The unit podman-coaster-celery-agent.service has successfully entered the 'dead' state.
2月 02 15:47:00 node-1.domain.tld systemd[1]: Started Celery Service via podman.
-- Subject: podman-coaster-celery-agent.service 单元已结束启动
-- Defined-By: systemd
-- Support: https://lists.freedesktop.org/mailman/listinfo/systemd-devel
-- 
[...]

PR #551, opened 2025-03-24
Handle descriptor in non-blocking mode properly.

jnovy:
Resolves: #490


haircommander:
LGTM, @giuseppe would you PTAL

jnovy:
Ok, let's keep it simple - I will remove the polling - it adds more complexity and potentially could slow things down (+the 500ms magic constant).

giuseppe (approved):
LGTM

PR #553, opened 2025-04-14
Add container labels to JournalD logs

OZoneGuy:
Add the ability to add the container labels to the journald logs as extra parameters.

Relies on the caller to pass the label key and value.

Usage:
```
conmon ... --log-labels label1=val1,label2=val2
```

The labels and values will be added to the journald entry.

This is my first contribution to a C codebase. Feedback is most welcome.


src/ctr_logging.c
```diff
@@ -169,6 +173,22 @@ void configure_log_drivers(gchar **log_drivers, int64_t log_size_max_, int64_t l
...
+		if (log_labels) {
+			// parse labels in the format "KEY=VALUE,KEY_2=VALUE_2,..."
+			container_labels_count = 1;
+			const char *ptr = log_labels;
+			while ((ptr = strchr(ptr, ',')) != NULL) {
+				container_labels_count++;
+				ptr++;
+			}
+			container_labels = g_alloca((sizeof(char *)) * container_labels_count);
+			container_labels_lengths = g_alloca((sizeof(int *)) * container_labels_count);
```
giuseppe, line 185: when are these ever going to be freed?  In this way they are just leaked

src/ctr_logging.c
```diff
@@ -169,6 +173,22 @@ void configure_log_drivers(gchar **log_drivers, int64_t log_size_max_, int64_t l
...
+			char *pair = strtok(log_labels, ",");
```
giuseppe, line 186: `strtok` modifies the buffer, so `log_labels` will be modified.  The correct way to do it is to `g_strdup` the string before passing it to `strtok`.

src/ctr_logging.c
```diff
@@ -325,6 +345,12 @@ static int write_journald(int pipe, char *buf, ssize_t buflen)
 		/* per docker journald logging format, CONTAINER_PARTIAL_MESSAGE is set to true if it's partial, but otherwise not set. */
 		if (partial && writev_buffer_append_segment_no_flush(&bufv, "CONTAINER_PARTIAL_MESSAGE=true", PARTIAL_MESSAGE_EQ_LEN) < 0)
 			return -1;
+		if (container_labels) {
+			for (int i = 0; i < container_labels_count; i++) {
+				if (writev_buffer_append_segment_no_flush(&bufv, container_labels[i], container_labels_lengths[i]) < 0)
+					return -1;
+			}
```
giuseppe, line 352: I suggest to not create the arrays before and then have to worry about cleaning them up.

Just strdup container_labels here and call `writev_buffer_append_segment_no_flush` while you parse it.

giuseppe (changes requested):

src/ctr_logging.c
giuseppe: that is fine, could you just add an inline comment saying it is by design?

src/ctr_logging.c
OZoneGuy: Going back to this. Since the array will live for the entire of the program. Is this needed? Not to challenge, just want to get more familiar with C and this codebase.

p12tic:
This PR can be closed due to https://github.com/containers/conmon/pull/562 being merged

PR #555, opened 2025-04-22
Logging fixups and cleanups

kolyshkin:
I was looking into how conmon logs things and found out a few things that can be improved, so here's this.

Please review commit by commit; see individual commits for details.

### DIFF
--- a/src/cmsg.c
+++ b/src/cmsg.c
@@ -36,19 +36,14 @@
 #define ECOMM EINVAL
 #endif
 
-#define error(s) \
+#define errorf(fmt, ...) \
 	do { \
-		fprintf(stderr, "conmon: %s %s\n", s, strerror(errno)); \
+		fprintf(stderr, "conmon: " fmt "\n", ##__VA_ARGS__); \
 		errno = ECOMM; \
 		goto err; /* return value */ \
 	} while (0)
 
-#define errorf(fmt, ...) \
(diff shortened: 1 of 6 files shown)


src/cmsg.c
```diff
@@ -120,7 +115,7 @@ struct file_t recvfd(int sockfd)
 	/* TODO: Make this dynamic with MSG_PEEK. */
 	file.name = malloc(TAG_BUFFER);
 	if (!file.name)
-		error("recvfd: failed to allocate file.tag buffer");
+		errorf("recvfd: failed to allocate file.tag buffer: %m");
```
jnovy, line 118: `malloc()` failure doesn't set errno according to ISO C standard. The errno is set only on POSIX compliant systems. At this point I'm not sure if somebody would run conmon on non-POSIX systems but it's good to keep things portable if possisble I think. Having said that I'd just skip %m as "failed to allocate file.tag buffer" says it all if file.name is NULL.

jnovy (changes requested):
Thanks for this, just one comment related to malloc() + errno.

src/cmsg.c
kolyshkin: Thanks! I was on the verge whether to use %m or not, and there was a version without it :) went back to it

jnovy:
All good now, thanks!

PR #557, opened 2025-04-23
seccomp_accept_cb: fix memory leak

kolyshkin:
Users of recvfd are expected to free the `.name`.


jnovy:
Yeah, the whole concept of `file_t` is a bit unfortunate... This is a memleak indeed. Thanks for noticing! LGTM

PR #558, opened 2025-04-27
include the correct header for syscall definitions

eli-schwartz:
This file needs to utilize `__NR_seccomp`, which is defined in the linux uapi headers, not seccomp.h, even though seccomp.h does itself indirectly cause this header to be included as well.

Nothing else in this particular file needs seccomp.h so drop that include altogether since it's now entirely unused.

### DIFF
--- a/src/seccomp_notify.c
+++ b/src/seccomp_notify.c
@@ -21,7 +21,7 @@
 #include <dlfcn.h>
 #include <sys/sysmacros.h>
 #include <linux/seccomp.h>
-#include <seccomp.h>
+#include <asm/unistd.h>
 
 #include "seccomp_notify.h"
 


jnovy:
Assuming more appropriate is to `#include <syscall.h>` or `#include <sys/syscall.h>` as `seccomp_notify.c` needs this to call the `__NR_seccomp` syscall?

src/seccomp_notify.c
```diff
@@ -21,7 +21,7 @@
 #include <dlfcn.h>
 #include <sys/sysmacros.h>
 #include <linux/seccomp.h>
-#include <seccomp.h>
+#include <asm/unistd.h>
```
jnovy, line 24: Assuming more appropriate is to #include <syscall.h> or #include <sys/syscall.h> as seccomp_notify.c needs this to call the __NR_seccomp syscall?

jnovy (changes requested):

kolyshkin (commented):
So, this does not fix anything, except maybe simplifies things a little.

There must be an underlying reason for this change, can you please share it @eli-schwartz ?

thesamesam:
The motivation here is to be able to drop the `libseccomp` dependency in our packaging, given there's already a reliance on `linux-headers` here, so we may as well get the same information from that.

eli-schwartz:
Hi, sorry for the delay in responding (thanks Sam for prodding me).

As Sam said, the underlying goal here was trying to remove unneeded deps. I was trying to figure out where libseccomp was(n't) used and this seemed to jump out at me early on.

I've looked a bit deeper, and it seems this could in theory be useful for EOL kernels, but shouldn't try to support unsupported EOL kernels. However if we do want that it should be a configure time kernel check... I can update to do that instead if you prefer.

WDYT about requiring kernel 5.0?

jnovy:
@eli-schwartz Are you planning to update your PR based on your comment?

>  However if we do want that it should be a configure time kernel check... I can update to do that instead if you prefer.

kolyshkin:
We've dropped seccomp in https://github.com/containers/conmon/pull/671, so this one can be closed

PR #561, opened 2025-05-05
Require at least golang 1.23

jnovy:


kolyshkin:
@jnovy you need to run `go mod tidy` after bumping go version in `go.mod`.

kolyshkin:
Also, go 1.22 is no longer supported since go 1.24 is out (i.e. since February 2025).

jnovy:
Thanks @kolyshkin - silly me forgetting this.

For golang-1.24 we have a bit of a chicken and egg problem - it still hasn't landed to major current distros. E.g we have still go-1.23.6 in CentOS/RHEL.

rhatdan:
@jnovy Still working on this?  At least use go  mod tidy

jnovy:
Go mod tidy was already done, maybe it makes sense to bump it even more to golang 1.23 now.

jnovy:
@kolyshkin Maybe we should make this part of your https://github.com/containers/conmon/pull/569 ?

runner/conmon_test/suite_test.go
```diff
@@ -60,7 +60,7 @@ func getConmonOutputGivenOptions(options ...conmon.ConmonOption) (string, string
 
 	alljournalout, err := getConmonJournalOutput(pid, -1)
 	Expect(err).To(BeNil())
-	fmt.Fprintf(GinkgoWriter, alljournalout+"\n")
+	fmt.Fprintf(GinkgoWriter, "%s\n", alljournalout)
```
kolyshkin, line 63: Perhaps it's better/simpler to use `fmt.Fprintln` here?

```suggestion
	fmt.Fprintln(GinkgoWriter, alljournalout)
```

jnovy:
@kolyshkin Thanks, I updated the printf statements, I think it's fair to update to 1.25 soon but merge this one first.

kolyshkin:
> @kolyshkin Thanks, I updated the printf statements, I think it's fair to update to 1.25 soon but merge this one first.

That's not what I meant. Let me try again.

The `go <version>` directive in go.mod specifies the minimally required go version. This means that anything older is not supported.

So we should not use 1.24 in there (as we still want 1.23 to be supported), and we should not switch it to 1.25 as soon as it's out (we still want 1.24 to be supported).

In general, we should be conservative here, meaning at the very least support all officially supported Go versions (those are 1.23 and 1.24 now). Meaning, today I would put `go 1.23` to go.mod, not `go 1.24`.

jnovy:
Ok, merging with requiring at least go 1.23 to continue with your PR.

PR #570, opened 2025-07-18
Fix k8s-file log corruption during log rotation

jnovy:
The k8s-file log driver was corrupting log entries when the log file reached max_size and needed rotation. This manifested as garbled output with repeated timestamps and broken log entries.

The root cause was in writev_buffer_flush() which incorrectly handled partial writes. When writev() returned a partial write, the function would modify the original iovec base pointers, corrupting the buffer state. During log rotation, this corrupted state would carry over to the new file descriptor, causing subsequent log entries to be malformed.

Changes:
- Simplify writev_buffer_flush() to use individual write() calls instead of complex writev() partial write handling
- Always reset buffer state after log rotation to ensure clean state with new file descriptor
- Remove conditional buffer reset that could leave corrupted state

This fixes the issue where long-running containers (like GitLab CE) would produce corrupted logs after reaching the configured max_size.

Fixes: Log corruption with --log-driver k8s-file --log-opt max_size=20mb

### DIFF
--- a/runner/conmon/options.go
+++ b/runner/conmon/options.go
@@ -97,6 +97,12 @@ func WithLogLevel(level string) ConmonOption {
 	}
 }
 
+func WithLogSizeMax(sizeMax int64) ConmonOption {
+	return func(ci *ConmonInstance) error {
+		return ci.addArgs("--log-size-max", fmt.Sprintf("%d", sizeMax))
+	}
+}
+
 func WithSocketPath(path string) ConmonOption {
 	return func(ci *ConmonInstance) error {
 		// TODO verify path is right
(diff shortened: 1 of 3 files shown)


haircommander:
@jnovy do you feel up for adding a conmon integration test? it is a golang suite, but it could help verify this behavior and ensure we don't regress

jnovy:
Like this @haircommander ? PTAL

haircommander:
yeah that's a good start, though it's not testing the contents of this PR. you could create a container that logs higher than the max (maybe setting the max very low) and making sure the two files are valid?

jnovy:
This should test what it should - PTAL @haircommander

haircommander:
looks great thank you!

Issue #545, opened 2025-02-24
Crio-conmon.scope still running despite crio-.scope not


xuegege5290:
Ref:https://github.com/opencontainers/runc/issues/4620

> sorry there's a lot to catch up on--when we see the cgroup.procs error, is the corresponding conmon still running? as in crio-conmon.scope still running despite crio-.scope not?

> If so, then it may be a conmon issue of missing the container's exit. From the logs above, it looks like conmon may still be running though that would be very strange. If conmon missed the exit, then it wouldn't notify crio the container exited, which means crio's state wouldn't be updated, causing the kubelet to continue to do exec probes on the container (and for crictl to allow you to run exec on a container process that is gone).

xuegege5290:
I added some log information. Under normal circumstances, once crio.xxx.scope receives an exit message, conmon should be able to capture it, just like this:
```bash
Feb 24 09:37:54 syx conmon[16909]: conmon 9cb5e3ff7214dcc7dfd3 <ninfo>: container 16936 exited with status 143

Feb 24 09:37:54 syx systemd[1]: crio-9cb5e3ff7214dcc7dfd393366a5dba5517d425b6edeb9c0749de7d9252aabc1f.scope: Succeeded.
```

xuegege5290:
However, under abnormal circumstances, after crio-xxx.scope exits, conmon does not receive any signals.
Journalctl log
```bash
[...]

PR #571, opened 2025-07-30
Fix container exit detection in systemd scope environments

jnovy:
When container processes are not direct children of conmon conmon fails to detect container exits because it never receives SIGCHLD signals.

This fixes issue where conmon processes remain running after container exit in certain systemd cgroup manager configurations.

Fixes: #545


giuseppe:
how would this happen?

conmon is setting `set_subreaper()` so that any child process, even not direct, is reported to conmon

giuseppe:
does it break when there are `exec` sessions?

jnovy:
Good catch! Let me add some bits to handle this.

giuseppe:
I am not really a fan of the probe part, can we understand how that configuration happens? Because conmon should get the signal, even with exec sessions (the container exits, all the processes in the pid namespace are killed and their parent gets the sigchld)

giuseppe:
how does the cgroup affect the parent->child relationship?  I don't think it should matter in what cgroup the process ends up, it should still be reported to the conmon (assuming set_reaper() was used correctly) used for the exec session.

You can verify it by running `podman run ....`, then do `podman exec ...` and find in the process list the exec'ed process:

```
$ podman exec -lti sleep 12345 &
$ grep -i ppid /proc/$(pgrep -f 12345 | tail -n1)/status
PPid:   380778
```

and PPid is the conmon process.

jnovy:
Comment adjusted, thanks!

PR #573, opened 2025-08-11
Add optional systemd support for static builds

jnovy:
Implement conditional systemd linking to fix static build failures. Static builds can now disable systemd with DISABLE_SYSTEMD=1. Nix derivation updated with enableSystemd parameter for flexibility. Provides backward compatibility while solving static linking issues.

Fixes: #348


giuseppe (approved):
LGTM

Issue #574, opened 2025-08-11
Most log macros functions in utils.h are incorrect


mtrmac:
In things like 
```
fprintf(stderr, "[conmon:e]: %s %m\n", s); \
if (use_syslog) \
	syslog(LOG_ERR, "conmon %.20s <error>: %s %m\n", log_cid, s); \
```
`fprintf` modifies `errno` on error, and it _may_ modify it even on success, so the second `%m` might be logging an incorrect error cause.

PR #575, opened 2025-08-12
Fix errno race condition and logging macro issues

jnovy:
- Fix errno race condition in logging macros
- Fix pwarnf macro structure to prevent control flow issues
- Add missing log level check to pwarn macro
- Fix syslog tag inconsistency in pwarnf
- Remove trailing spaces from syslog format strings
- Add missing colon in nexit stderr format

Fixes: #574

### DIFF
--- a/src/utils.h
+++ b/src/utils.h
@@ -6,6 +6,7 @@
 #include <syslog.h>
 #include <stdbool.h>
 #include <unistd.h>
+#include <errno.h>
 #include <glib.h>
 #include <glib-unix.h>
 #include <sys/uio.h>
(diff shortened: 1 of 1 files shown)


jnovy:
@mtrmac PTAL

src/utils.h
```diff
@@ -79,25 +83,31 @@ extern gboolean use_syslog;
...
 	do { \
-		fprintf(stderr, "[conmon:w]: %s %s\n", s, strerror(errno)); \
-		if (use_syslog) \
-			syslog(LOG_INFO, "conmon %.20s <pwarn>: %s %s\n", log_cid, s, strerror(errno)); \
+		if (log_level >= WARN_LEVEL) { \
+			int saved_errno = errno; \
+			fprintf(stderr, "[conmon:w]: %s %s\n", s, strerror(saved_errno)); \
+			if (use_syslog) \
+				syslog(LOG_INFO, "conmon %.20s <pwarn>: %s %s\n", log_cid, s, strerror(saved_errno)); \
+			errno = saved_errno; \
```
mtrmac, line 97: The restores seem unnecessary? (2 cases)

mtrmac (commented):
Locally this looks good, except that `strerror` is not thread-safe. Is this all guaranteed to run on the main thread?

I didn’t _find_ anything to suggest multi-threaded use, the code seems to be centered around a `GMainLoop`, but I am not familiar with the codebase.

src/utils.h
jnovy: Yes, I can remove these as it seems an overkill. I'd need to add a similar thing to all debug() and info() macros if we want to go ahead with a paradigm that logging functions themselves shouldn't affect errno.

jnovy:
The conmon is single thread - just updated the implementation so that it's thread safe if we decide to go that way.

haircommander:
LGTM

src/utils.h
mtrmac: I generally think that the most reliable implementation approach is: if anything needs an `errno` value, it _must_ capture it immediately after a failing call itself; a rule that “some functions preserve errno” is too difficult to maintain, and tools won’t help us.

src/utils.h
```diff
@@ -35,41 +38,53 @@ extern log_level_t log_level;
 extern char *log_cid;
 extern gboolean use_syslog;
 
+static inline const char *get_errno_string(void)
+{
+	static __thread char errbuf[256];
+	int saved_errno = errno;
+	snprintf(errbuf, sizeof(errbuf), "%s", strerror(saved_errno));
```
mtrmac, line 45: `if … >= sizeof(errbuf) { abort() }` or something? (`sprintf("[unexpected error %d"])` is an option but unreasonable)

src/utils.h
```diff
@@ -35,41 +38,53 @@ extern log_level_t log_level;
 extern char *log_cid;
 extern gboolean use_syslog;
 
+static inline const char *get_errno_string(void)
+{
+	static __thread char errbuf[256];
```
mtrmac, line 43: (Absolutely unblocking: The `__thread` works but it seems unnecessary, the caller can allocate a buffer on stack.)

src/utils.h
mtrmac, line 45: This does not avoid the usage of `strerror`, so I don’t think this is an improvement compared to the previous version?!

Using `%m` here might work.

mtrmac:
Fine :) I’ll read that carefully later, but this clearly works.

Issue #426, opened 2023-06-02
OOM detection on cgroup v2 not always working


kolyshkin:
Continuing the investigation that started in https://github.com/cri-o/cri-o/pull/6973

Apparently, sometimes systemd removes the cgroup before conmon can read its `memory.events`. I don't have any ideas about how to work around that race (in conmon; that is).

Another small issue I found is sometimes OOM can be misdetected because the code use the same static counter variable for both `oom` and `oom_kill`, which appears to be wrong.

https://github.com/containers/conmon/blob/ebc6717071589813b3eb4d2f18ef5999b7d263b0/src/cgroup.c#L269-L310

PR #581, opened 2025-09-01
Fix OOM detection on cgroup v2

jnovy:
- Use separate counters for oom and oom_kill events
- Handle race condition when systemd removes cgroup
- Prevent buffer corruption in getline() usage
- Add comprehensive test coverage

Fixes: #426


jnovy:
@kolyshkin PTAL

Issue #335, opened 2022-05-12
Long lines will be splitted when journald logs


xeor:
In https://github.com/containers/conmon/blob/2b087348289664b75c874ce44796f762d636c060/src/ctr_logging.c#L320-L322, `CONTAINER_PARTIAL_MESSAGE` is set static to `true`. I find no way to turn this off.

This can be problematic example when using ansible tower/awx, which uses conmon/podman to start jobs inside containers running its job since those jobs often contains a bunch of json from ansible. You will end up with a splitted message in journald, which is not even in order..

If you send journald messages to another system as well, you will have no way of putting it together, and it will often break parsing and cause all sorts of problems since there are unbalanced quotes per line.

There should be a way to log without setting `CONTAINER_PARTIAL_MESSAGE` to true

haircommander:
thanks for opening the issue @xeor , do you have any interest in opening a PR?

xeor:
sorry for the delay..

I honestly don't think I should try to make a PR for this. I have not been doing a lot (any c programming).
I am working around this by using the logging functionality inside awx/tower instead of journald. So this is still a problem.

PR #583, opened 2025-09-02
Add --no-container-partial-message option

jnovy:
- Add `--no-container-partial-message` CLI option to disable CONTAINER_PARTIAL_MESSAGE=true in journald logs                                                              - This prevents splitting of long log lines into multiple journal entries
- Particularly useful for systems parsing structured logs like JSON
- Option only affects journald log driver, warns if used without journald                  

Fixes: #335


giuseppe (approved):
LGTM

PR #584, opened 2025-09-02
Fix conmon exec exit status handling

### LINKED ISSUE #328: conmon exec not handling runtime failures
Looks like at least when doing exec, conmon is ignoring the exit status of the runtime and returns 0. That made it really hard to pin point an issue we're having.

Using /bin/false to easily reproduce:

```
# conmon -c <id> -n /runtime.v1alpha2.RuntimeService/Exec -p /mnt/crio/overlay-containers/<id>/userdata/pidfile -e --socket-dir-path /var/run/crio --exec-process-spec exec-spec.json -r /bin/false --log-path /dev/stdout
# echo $?
0
```

jnovy:
For exec operations, use runtime exit status instead of container status and skip runtime failure check since non-zero exit is expected behavior.

The fix addresses two core issues:
- early exit when runtime returns non-zero (treating exec failures as runtime failures)
- using container_status instead of runtime_status for final exit code determination

Fixes #328

### DIFF
--- a/src/conmon.c
+++ b/src/conmon.c
@@ -365,7 +365,10 @@ int main(int argc, char *argv[])
 		}
 	}
 
-	if (!WIFEXITED(runtime_status) || WEXITSTATUS(runtime_status) != 0) {
+	/* For exec operations, the runtime exit status reflects the exit status of the exec'd command,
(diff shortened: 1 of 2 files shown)


giuseppe (approved):
LGTM

Issue #582, opened 2025-09-01
Assorted suspicions


mtrmac:
This is a random collection of unclear / risky / suspect code I noticed while toying with conmon vs. Rust.

For all I know, none of them might really _matter_, and I don’t really have the capacity to test / reproduce / fix any of that properly. But maybe someone else is interested.

Warning: I have, at best, one-function-at-a-time understanding of conmon. I couldn’t tell you what it does or in what order or under which conditions.

[conmon-fixmes.patch](https://github.com/user-attachments/files/22079783/conmon-fixmes.patch)

PR #587, opened 2025-09-04
Fix code quality issues

jnovy:
- Fix strncpy missing null termination in conn_sock.c
- Replace malloc with g_malloc for consistency in ctr_exit.c
- Add bounds validation for sscanf parsing in ctrl.c
- Fix signal safety: change volatile pid_t to sig_atomic_t
- Add null check for cgroup2_path before use
- Clean up outdated TODO/FIXME comments

These changes address potential buffer safety, memory allocation consistency, signal safety, and logic issues flagged during code review.

Fixes: #582

### DIFF
--- a/src/cli.c
+++ b/src/cli.c
@@ -196,7 +196,6 @@ void process_cli()
 	}
 
 	// we should always override the container pid file if it's empty
-	// TODO FIXME I removed default_pid_file here. shouldn't opt_container_pid_file be cleaned up?
 	if (opt_container_pid_file == NULL)
 		opt_container_pid_file = g_strdup_printf("%s/pidfile-%s", cwd, opt_cid);
 
--- a/src/conmon.c
+++ b/src/conmon.c
@@ -253,8 +253,6 @@ int main(int argc, char *argv[])
 			if (workerfd_stdout != dev_null_w && fchmod(STDOUT_FILENO, 0777) < 0 && errno != EINVAL)
 				nwarn("Failed to chmod stdout");
 
-			if (workerfd_stderr < 0)
-				workerfd_stderr = workerfd_stdout;
 			if (dup2(workerfd_stderr, STDERR_FILENO) < 0)
 				_pexit("Failed to dup over stderr");
 			if (workerfd_stderr != dev_null_w && fchmod(STDERR_FILENO, 0777) < 0 && errno != EINVAL)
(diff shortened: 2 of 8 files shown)


src/conmon.c
```diff
@@ -180,11 +180,11 @@ int main(int argc, char *argv[])
 
 	if (opt_seccomp_notify_socket != NULL) {
 #ifdef USE_SECCOMP
-		pexit("seccomp support not present");
-#else
 		if (opt_seccomp_notify_plugins == NULL)
 			pexit("seccomp notify socket specified without any plugin");
 		seccomp_listener = setup_seccomp_socket(opt_seccomp_notify_socket);
+#else
+		pexit("seccomp support not present");
```
mtrmac, line 187: E.g. this is still printing an `errno`-derived value.

mtrmac (commented):
The changes here LGTM.

Note that this _by no means_ handles everything from #582 , e.g. see the example below. But, also, I’m not going to be able to follow up on everything there, so the resolution, if any, is entirely up to you.

src/cli.c
```diff
@@ -117,6 +118,7 @@ GOptionEntry opt_entries[] = {
 	 "Path to the socket where the seccomp notification fd is received", NULL},
 	{"seccomp-notify-plugins", 0, 0, G_OPTION_ARG_STRING, &opt_seccomp_notify_plugins,
 	 "Plugins to use for managing the seccomp notifications", NULL},
+	{"seccomp-notify-allowed", 0, 0, G_OPTION_ARG_NONE, &opt_seccomp_notify_allowed, "Allow seccomp notify annotations", NULL},
```
mtrmac, line 121: Does the man page need updating for the new option?

(If this didn’t work for … years, my first instinct would be to delete the code wholesale, but I know nothing of the history.)

jnovy: I don't know much about the history either. @giuseppe ?

giuseppe: this was added to experiment with crun-style seccomp notification plugins: https://github.com/containers/crun/blob/main/contrib/seccomp-notify-plugin-example/full.c

Not sure if anyone is using it, we didn't follow up to expose it in Podman because using them is quite difficult

giuseppe: but why do we need opt_seccomp_notify_allowed?

If they are specified in `opt_seccomp_notify_plugins` we use them, otherwise not.  Why another bool?

If anything, it should be a separate commit

jnovy: Ok, I removed the `seccomp-notify-allowed` as `opt_seccomp_notify_plugins` already handles that.

Issue #589, opened 2025-09-04
Investigate regression uncovered by podman testing


jnovy:
The podman issue for reference: https://github.com/containers/podman/issues/26981

PR #590, opened 2025-09-04
Revert 'Fix conmon exec exit status handling'

jnovy:
The original fix introduced a regression where exec operations always returned exit code 0 instead of the actual command exit code. This broke podman health checks and exec exit code propagation.

The original code was working correctly - container_status contains the actual exit code of executed commands, while runtime_status only indicates whether the runtime successfully executed the command.

Testing shows:
- Before fc0a342: exec commands return correct exit codes - OK
- After fc0a342: exec commands always return 0 - NOT OK
- After this revert: exec commands return correct exit codes - OK

Related issue: https://github.com/containers/podman/issues/26981

Fixes: #589


test/06-exec-exit-status.bats
```diff
@@ -12,40 +12,84 @@ teardown() {
...
+    # Create a test container
+    local container_id
+    container_id=$(podman run -dt busybox:latest sleep 30)
+
+    if [ -z "$container_id" ]; then
+        skip "Failed to create test container"
+    fi
+
+    # Test 1: Success case
+    if ! podman exec "$container_id" true; then
```
Honny1, line 69: Does this work if podman is run without the --conmon option?

Makefile
```diff
@@ -116,7 +116,7 @@ install.podman: bin/conmon
 .PHONY: fmt
 fmt:
 	git ls-files -z \*.c \*.h | xargs -0 clang-format -i
-
+	git ls-files -z '*.bats' | xargs -0 sed -i 's/[[:space:]]\+$$//'
```
Luap99, line 119: should this be in this commit?

Luap99 (commented):
reverting LGTM, I cannot really comment on the testing details here.

Makefile
jnovy: This one is just missing format fix for BATS tests to avoid trailing spaces - intentional.

PR #594, opened 2025-09-10
Show the code coverage in `make test-binary` output.

jankaluza:
This commit changes the `make test-binary` to recompile the conmon with `--coverage` flag so it generates the `.gcno and `.gcda` (these files are included in the `.gitignore` now) while running the tests.

After that, the `gcovr` is executed to show the code coverage based on those files.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -89,6 +89,19 @@ build_task:
         - make
         - make test
 
+# Test coverage
+coverage_task:
+    # Runs within Cirrus's "community cluster"
+    container:
+        image: "${FEDORA_CONTAINER_FQIN}"
+        cpu: 1
+        memory: 4
+
+    script:
+        - dnf install -y make glib2-devel git gcc pkg-config systemd-devel libseccomp-devel gcovr
+        - cd $CIRRUS_WORKING_DIR
+        - make test-coverage
+
 
 # Verify code was fmt'ed properly
 fmt_task:
--- a/.gitignore
+++ b/.gitignore
@@ -51,6 +51,11 @@ Module.symvers
 Mkfile.old
 dkms.conf
 
+# --coverage files
+*.gcno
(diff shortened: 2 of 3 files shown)


jankaluza:
Output is visible at the end of https://cirrus-ci.com/task/6362001716281344.

jnovy:
@jankaluza Can we use just normal `gcov` so that we don't need to add a new package dependency and maybe adding a separate `test-coverage` target instead?

Something like this?

```
diff --git a/Makefile b/Makefile
index 90ff8f0..72b25a0 100644
--- a/Makefile
+++ b/Makefile
@@ -79,6 +79,15 @@ test-binary: bin/conmon
 test: bin/conmon
        CONMON_BINARY="$(MAKEFILE_PATH)bin/conmon" test/run-tests.sh
 
+.PHONY: test-coverage
+test-coverage: clean
+       $(MAKE) bin/conmon DEBUGFLAG="--coverage"
+       CONMON_BINARY="$(MAKEFILE_PATH)bin/conmon" test/run-tests.sh
+       @echo "=== Coverage Summary ==="
+       @for file in src/*.c; do \
+               gcov $$file 2>/dev/null | grep -E "(^File|^Lines executed)" | paste - - | sed 's/File '\''src\/\([^.]*\)\.c'\''.*Lines executed:\([0-9.]*%\).*/\1.c: \2/'; \
+       done | grep -E "\.c: [0-9]" | sort
+
 bin:
        mkdir -p bin
 
@@ -90,7 +99,7 @@ docs:
 
 .PHONY: clean
 clean:
-       rm -rf bin/ src/*.o
+       rm -rf bin/ src/*.o src/*.gcda src/*.gcno *.gcov
        $(MAKE) -C docs clean
 
 .PHONY: install install.bin install.crio install.podman podman crio
```

Issue #211, opened 2020-10-06
Improve --log-size-max handling


rhatdan:
Currently conmon just truncates the log file when max is reached, this means we loose all of the data.

I think we should allow for a backup file.  So when log-size-max is reached we rename the existing file to a .1 version.  Now when the user reads the full log file they would read the .1 file first and then the main log file.  This would mean we would always have log-size-max data available, after we fill it up. If .1 already existed, we would loose that data.

Since user asked for log-size-max, we should probably do the rename at log-size-max/2 to grant their wishes.

@haircommander @giuseppe @vrothberg @mheon WDYT?

rhatdan:
Docker also handles an addition field to indicate the number of backup files, I believe.

vrothberg:
> Docker also handles an addition field to indicate the number of backup files, I believe.

That answers the immediate question I had in mind: How do we set a limit to protect from running out of disk space?

SGTM

haircommander:
the approach sounds good, as long as the behaviour is gated by a flag. In the k8s world, kubelet handles the renaming of the files, so we should preserve existing behaviour unless otherwise specified

haircommander:
another thing to think about is how to handle the metadata. does docker truncate based on file size or based on size of log line? regardless, we'll have differing behaviour until conmon actually supports json-file

exekias:
++ on this issue, we (Filebeat dev here) are seeing more issues coming from k8s users using alternative runtimes, like CRI-o. Having truncate as the mechanism for rotation makes it impossible to guarantee delivery of all log messages. A rename + create approach like the one describe in this issue allows agents to keep reading both the old and the new file after a rotation event.

> does docker truncate based on file size or based on size of log line?

Docker does this based on size, which probably makes sense, as it allows system administrators to allocate disk for logs in a predictable way, regardless of log line lengths.

Also a clarification about Docker rotation strategy: Docker doesn't truncate files, this is important. It will rename the existing log file and create a new one for the following logs.

PR #596, opened 2025-09-11
Add log rotation functionality as alternative to log truncation

jnovy:
This commit implements comprehensive log rotation functionality for conmon:

- Add --log-rotate and --log-max-files CLI options
- Implement rotate_k8s_file() and shift_backup_files() functions
- Add comprehensive BATS test suite covering rotation and truncation scenarios
- Maintain backward compatibility with existing truncation behavior
- Update existing tests to work with new functionality

The log rotation creates backup files (.1, .2, etc.) instead of truncating when --log-rotate is enabled, providing better log preservation.

Fixes: #211

### DIFF
--- a/docs/conmon.8.md
+++ b/docs/conmon.8.md
@@ -63,11 +63,24 @@ Leave stdin open when the attached client disconnects.
 **--log-level**
 Print debug logs based on the log level.
 
+**--log-max-files**
+Maximum number of log backup files to keep when log rotation is enabled. Default is 1.
+
+**--log-rotate**
+Enable log rotation instead of log truncation. When enabled, log files are rotated
+with numbered suffixes (.1, .2, etc.) instead of being truncated when they reach
+the maximum size.
+
 **--log-size-max**
 Maximum size of the log file (in bytes).
 
 **--log-tag**
 Additional tag to use for logging.
+
+**--log-allowlist-dir**
+Specifies allowed directories for log file creation. This option can be specified multiple times to allow
+multiple directories. When configured, log files can only be created within these allowed directories or
+their subdirectories. If not specified, log files can be created in any location.
 
 **--log-label**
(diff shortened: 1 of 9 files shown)


jnovy:
Updated, PTAL @jankaluza

test/06-log-management.bats
```diff
@@ -0,0 +1,230 @@
...
+    local allowed_dir="$TEST_TMPDIR/conmon-allowed-$$"
+    if ! mkdir -p "$allowed_dir" 2>/dev/null; then
+        skip "Cannot create test directory for allowlist test"
+    fi
+
+    local allowed_log="$allowed_dir/test.log"
+
+    run_conmon --cid "$CTR_ID" --cuuid "$CTR_ID" --runtime "$VALID_PATH" \
+        --log-path "k8s-file:$allowed_log" \
+        --log-allowlist-dirs "$allowed_dir" \
```
jankaluza, line 222: Can you also test second `--log-allowlist-dirs`?

```
--log-allowlist-dirs "$allowed_dir"
--log-allowlist-dirs /foo
```

So we know that multiple options are enabled and /foo does not override the first one.

docs/conmon.8.md
```diff
@@ -63,6 +63,14 @@ Leave stdin open when the attached client disconnects.
...
 Print debug logs based on the log level.
 
+**--log-max-files**
+Maximum number of log backup files to keep when log rotation is enabled. Default is 1.
+
+**--log-rotate**
+Enable log rotation instead of log truncation. When enabled, log files are rotated
+with numbered suffixes (.1, .2, etc.) instead of being truncated when they reach
+the maximum size.
+
```
jankaluza, line 73: I think the `log-allowlist-dir` should still be documented here.

jankaluza:
LGTM.

I think this is OK from the code point of view and the test coverage is also solid.

PR #599, opened 2025-09-13
healthcheck feature

samifruit514:
context:
We are running a multi-container application with docker-compose, but through podman (`podman system service --log-level debug unix:///tmp/podman.sock`). The apps definitions inside the docker-compose.yml file contains a bunch of health checks and dependencies. Since we run that in a CI, WITHOUT systemd, there is no healthchecks (no unit are created for healthchecks). Because of that, the multi-container app doesnt run.

According to podman people, conmon should handle health checks, or at least, conmon would be a great candidate to do the healthchecks.

This PR accepts `--enable-healthcheck` in conmon args, enabling the healthchecks from conmon to podman, through the unix pipe (the same one that sends the PID to link).

For more info on healthcheck handling by podman: https://github.com/containers/podman/pull/27033

### DIFF
--- a/src/healthcheck.c
+++ b/src/healthcheck.c
@@ -1,0 +1,533 @@
+#define _GNU_SOURCE
+
+#include "healthcheck.h"
+#include "utils.h"
+#include "ctr_logging.h"
+#include "parent_pipe_fd.h"
+#include "globals.h"
+#include "cli.h"
+#include "ctr_exit.h"
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include <errno.h>
+#include <time.h>
+#include <signal.h>
+#include <sys/wait.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <fcntl.h>
+#include <limits.h>
+#include <glib.h>
+
+/* Healthcheck validation constants */
+#define HEALTHCHECK_INTERVAL_MIN 1
+#define HEALTHCHECK_INTERVAL_MAX 3600
+#define HEALTHCHECK_TIMEOUT_MIN 1
+#define HEALTHCHECK_TIMEOUT_MAX 300
+#define HEALTHCHECK_START_PERIOD_MIN 0
+#define HEALTHCHECK_START_PERIOD_MAX 3600
+#define HEALTHCHECK_RETRIES_MIN 0
+#define HEALTHCHECK_RETRIES_MAX 100
+
+/* Validate healthcheck configuration parameters */
+bool healthcheck_validate_config(const healthcheck_config_t *config)
+{
+	if (config == NULL) {
+		return false;
+	}
+
+	/* Validate interval */
+	if (config->interval < HEALTHCHECK_INTERVAL_MIN || config->interval > HEALTHCHECK_INTERVAL_MAX) {
+		nwarnf("Healthcheck interval %d is out of range [%d, %d]", config->interval, HEALTHCHECK_INTERVAL_MIN,
+		       HEALTHCHECK_INTERVAL_MAX);
+		return false;
+	}
+
+	/* Validate timeout */
+	if (config->timeout < HEALTHCHECK_TIMEOUT_MIN || config->timeout > HEALTHCHECK_TIMEOUT_MAX) {
+		nwarnf("Healthcheck timeout %d is out of range [%d, %d]", config->timeout, HEALTHCHECK_TIMEOUT_MIN,
+		       HEALTHCHECK_TIMEOUT_MAX);
+		return false;
+	}
+
+	/* Validate start period */
+	if (config->start_period < HEALTHCHECK_START_PERIOD_MIN || config->start_period > HEALTHCHECK_START_PERIOD_MAX) {
+		nwarnf("Healthcheck start period %d is out of range [%d, %d]", config->start_period, HEALTHCHECK_START_PERIOD_MIN,
+		       HEALTHCHECK_START_PERIOD_MAX);
+		return false;
+	}
+
+	/* Validate retries */
+	if (config->retries > HEALTHCHECK_RETRIES_MAX) {
+		nwarnf("Healthcheck retries %u is out of range [%d, %d]", config->retries, HEALTHCHECK_RETRIES_MIN,
+		       HEALTHCHECK_RETRIES_MAX);
+		return false;
+	}
+
+	/* Validate that timeout is not greater than interval */
+	if (config->timeout > config->interval) {
+		nwarnf("Healthcheck timeout %d cannot be greater than interval %d", config->timeout, config->interval);
+		return false;
+	}
+
+	return true;
+}
+
+/* Static string constants for healthcheck statuses */
+const char *healthcheck_status_strings[] = {"none", "starting", "healthy", "unhealthy"};
+
+/* Global healthcheck timer (one per conmon instance) */
+healthcheck_timer_t *active_healthcheck_timer = NULL;
+
+
+/* Cleanup healthcheck subsystem */
+void healthcheck_cleanup(void)
+{
+	if (active_healthcheck_timer != NULL) {
+		healthcheck_timer_stop(active_healthcheck_timer);
+		healthcheck_timer_free(active_healthcheck_timer);
+		active_healthcheck_timer = NULL;
+	}
+}
+
+/* Free healthcheck configuration */
+void healthcheck_config_free(healthcheck_config_t *config)
+{
+	if (config == NULL) {
+		return;
+	}
+
+	if (config->test != NULL) {
+		for (int i = 0; config->test[i] != NULL; i++) {
+			free(config->test[i]);
+		}
+		free(config->test);
+		config->test = NULL;
+	}
+	// Don't free config itself - it's a local variable on the stack
+}
+
+/* Create a new healthcheck timer */
+healthcheck_timer_t *healthcheck_timer_new(const char *container_id, const healthcheck_config_t *config)
+{
+	if (container_id == NULL || config == NULL) {
+		return NULL;
+	}
+
+	healthcheck_timer_t *timer = calloc(1, sizeof(healthcheck_timer_t));
+	if (timer == NULL) {
+		nwarn("Failed to allocate memory for healthcheck timer");
+		return NULL;
+	}
+
+	timer->container_id = strdup(container_id);
+	if (timer->container_id == NULL) {
+		free(timer);
+		return NULL;
+	}
+
+	timer->config = *config;
+	timer->status = HEALTHCHECK_NONE;
+	timer->consecutive_failures = 0;
+	timer->start_period_remaining = config->start_period;
+	timer->timer_active = false;
+	timer->last_check_time = 0;
+
+	/* Copy the test command array */
+	if (config->test != NULL) {
+		int argc = 0;
+		while (config->test[argc] != NULL) {
+			argc++;
+		}
+
+		timer->config.test = calloc(argc + 1, sizeof(char *));
+		if (timer->config.test == NULL) {
+			free(timer->container_id);
+			free(timer);
+			return NULL;
+		}
+
+		for (int i = 0; i < argc; i++) {
+			timer->config.test[i] = strdup(config->test[i]);
+			if (timer->config.test[i] == NULL) {
+				/* Clean up on error */
+				for (int j = 0; j < i; j++) {
+					free(timer->config.test[j]);
+				}
+				free(timer->config.test);
+				free(timer->container_id);
+				free(timer);
+				return NULL;
+			}
+		}
+		timer->config.test[argc] = NULL;
+	}
+
+	return timer;
+}
+
+/* Free healthcheck timer */
+void healthcheck_timer_free(healthcheck_timer_t *timer)
+{
+	if (timer == NULL) {
+		return;
+	}
+
+	/* Stop the timer if it's still active */
+	if (timer->timer_active) {
+		healthcheck_timer_stop(timer);
+	}
+
+	/* Free container ID */
+	if (timer->container_id != NULL) {
+		free(timer->container_id);
+	}
+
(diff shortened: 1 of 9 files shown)


samifruit514:
Thanks for taking a look mheon! its really appreciated. I've made the changes to handle healthchecks in starting phase. Let me know your feedback! Thanks

samifruit514:
I've made a PR on podman side: It would read from the conmon's pipe: https://github.com/containers/podman/pull/27067
Sorry if its too noisy. Set that PR to draft for now. Its to give a global overview on how it would work

src/healthcheck.c
```diff
@@ -0,0 +1,645 @@
...
+
+/* Create a new healthcheck configuration */
+healthcheck_config_t *healthcheck_config_new(void) {
+    healthcheck_config_t *config = calloc(1, sizeof(healthcheck_config_t));
+    if (config == NULL) {
+        nwarn("Failed to allocate memory for healthcheck config");
+        return NULL;
+    }
+    
+    config->interval = 30;      /* Default 30 seconds */
```
mheon, line 195: I'd want all of this passed explicitly. We shouldn't keep defaults in two places; Podman should pass every argument every time instead.

src/healthcheck.c
```diff
@@ -0,0 +1,645 @@
...
+    ht->count++;
+    
+    return true;
+}
+
+/* Global healthcheck timers hash table */
+struct hash_table *active_healthcheck_timers = NULL;
+
+/* Check if systemd is available and running */
+bool healthcheck_is_systemd_available(void) {
```
mheon, line 131: Feels unnecessary, this should be in Podman. If we tell Conmon to do a healthcheck it should do it. If we don't, it's because systemd is available and we're preferring that.

mheon:
I'm a bit hesitant about this one because we are strongly contemplating a Conmon rewrite to be released alongside Podman 6 in May - https://github.com/containers/podman/pull/27053 - so additional work on the existing, C conmon seems a bit wasted in light of that. But if this can be done for our November release of 5.7 and that's of use to you, I'm not entirely opposed?

src/healthcheck.c
samifruit514: cleaned up. thanks

src/healthcheck.c
```diff
@@ -457,52 +457,77 @@ bool healthcheck_parse_oci_annotations(const char *annotations_json, healthcheck
...
-            if (strcmp(cmd_type->valuestring, "CMD") == 0 || strcmp(cmd_type->valuestring, "CMD-SHELL") == 0) {
-                /* Create test command array */
-                config->test = calloc(2, sizeof(char*));
-                config->test[0] = strdup(cmd_value->valuestring);
-                config->test[1] = NULL;
-            }
-        }
+    if (!cJSON_IsArray(test_array) || cJSON_GetArraySize(test_array) < 2) {
+        nwarn("Healthcheck configuration missing required 'test' command");
+        cJSON_Delete(json);
```
jnovy, line 472: `healthcheck_config_free(config);` is missing here?

src/healthcheck.c
```diff
@@ -457,52 +457,77 @@ bool healthcheck_parse_oci_annotations(const char *annotations_json, healthcheck
...
+        return false;
+    }
+    
+    cJSON *cmd_type = cJSON_GetArrayItem(test_array, 0);
+    cJSON *cmd_value = cJSON_GetArrayItem(test_array, 1);
+    
+    if (!cJSON_IsString(cmd_type) || !cJSON_IsString(cmd_value)) {
+        nwarn("Healthcheck test command must be an array of strings");
+        cJSON_Delete(json);
```
jnovy, line 481: `healthcheck_config_free(config);` is missing here too?

src/healthcheck.c
```diff
@@ -457,52 +457,77 @@ bool healthcheck_parse_oci_annotations(const char *annotations_json, healthcheck
...
+        return false;
     }
     
-    /* Parse Interval (now in seconds) */
+    if (strcmp(cmd_type->valuestring, "CMD") == 0 || strcmp(cmd_type->valuestring, "CMD-SHELL") == 0) {
+        /* Create test command array */
```
jnovy, line 486: Please validate the command string here, e.g. `strlen(cmd_value->valuestring) == 0 || strlen(cmd_value->valuestring) > 4096` ?

src/healthcheck.c
```diff
@@ -457,52 +457,77 @@ bool healthcheck_parse_oci_annotations(const char *annotations_json, healthcheck
...
+        config->test = calloc(2, sizeof(char*));
+        config->test[0] = strdup(cmd_value->valuestring);
```
jnovy, line 488: Test calloc and strdup for failure here.

src/healthcheck.c
```diff
@@ -457,52 +457,77 @@ bool healthcheck_parse_oci_annotations(const char *annotations_json, healthcheck
...
+        config->test[1] = NULL;
+    } else {
+        nwarnf("Unsupported healthcheck command type: %s (only CMD and CMD-SHELL supported)", cmd_type->valuestring);
+        cJSON_Delete(json);
```
jnovy, line 492: missing `healthcheck_config_free(config);`

src/healthcheck.c
```diff
@@ -457,52 +457,77 @@ bool healthcheck_parse_oci_annotations(const char *annotations_json, healthcheck
...
+        return false;
+    }
+    
+    /* Parse Interval (now in seconds) - REQUIRED */
     cJSON *interval = cJSON_GetObjectItem(json, "interval");
-    if (cJSON_IsNumber(interval)) {
-        config->interval = (int)interval->valuedouble;
+    if (!cJSON_IsNumber(interval) || interval->valuedouble <= 0) {
```
jnovy, line 498: Maybe add here also check for the max value here?

src/healthcheck.c
```diff
@@ -457,52 +457,77 @@ bool healthcheck_parse_oci_annotations(const char *annotations_json, healthcheck
...
+        return false;
     }
+    config->interval = (int)interval->valuedouble;
     
-    /* Parse Timeout (now in seconds) */
+    /* Parse Timeout (now in seconds) - REQUIRED */
     cJSON *timeout = cJSON_GetObjectItem(json, "timeout");
-    if (cJSON_IsNumber(timeout)) {
-        config->timeout = (int)timeout->valuedouble;
+    if (!cJSON_IsNumber(timeout) || timeout->valuedouble <= 0) {
```
jnovy, line 507: Also max value check here please + missing `healthcheck_config_free(config);` again

src/healthcheck.c
```diff
@@ -457,52 +457,77 @@ bool healthcheck_parse_oci_annotations(const char *annotations_json, healthcheck
...
+        return false;
     }
+    config->timeout = (int)timeout->valuedouble;
     
-    /* Parse StartPeriod (now in seconds) */
+    /* Parse StartPeriod (now in seconds) - REQUIRED (can be 0) */
     cJSON *start_period = cJSON_GetObjectItem(json, "start_period");
-    if (cJSON_IsNumber(start_period)) {
-        config->start_period = (int)start_period->valuedouble;
+    if (!cJSON_IsNumber(start_period) || start_period->valuedouble < 0) {
```
jnovy, line 516: Also max value check here please.

src/healthcheck.c
```diff
@@ -457,52 +457,77 @@ bool healthcheck_parse_oci_annotations(const char *annotations_json, healthcheck
...
+        return false;
     }
+    config->start_period = (int)start_period->valuedouble;
     
-    /* Parse Retries */
+    /* Parse Retries - REQUIRED */
     cJSON *retries = cJSON_GetObjectItem(json, "retries");
-    if (cJSON_IsNumber(retries)) {
-        config->retries = (int)retries->valuedouble;
+    if (!cJSON_IsNumber(retries) || retries->valuedouble < 0) {
```
jnovy, line 525: Also max value check here please.

src/healthcheck.c
```diff
@@ -0,0 +1,670 @@
...
+        nwarn("Invalid parameters for healthcheck command execution");
+        return false;
+    }
+    
+    /* For now, we'll use a simple approach: check if the container is running */
+    /* This is a placeholder implementation - in a real implementation, we would */
+    /* need to execute the command inside the container using the container runtime API */
+    
+    /* Simple healthcheck: just return success for now */
+    /* TODO: Implement proper container command execution */
```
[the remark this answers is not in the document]
samifruit514: wow. This is crazy. I can't believe I didnt catch that.
for json, cursor was doing a custom parsing without any JSON lib, so I asked to use cJSON because it was the only one I knew 😆

Thanks for all the tips. I will do the changes. Once I go back into the related podmon PR, I will make sure I didnt mess up with AI

samifruit514:
integration tests failed for cri-o. Would it be possible that its some race condition? I don't see how this is related to the changes in this PR 🤔
AI suggests to put some sleep in the bats tests on cri-o side

samifruit514:
I was able to run the "crio restore with missing config.json" test locally:
```
Running the test...
restore_test_cgroupfs.bats
 ✓ crio restore with missing config.json

1 test, 0 failures

✓ Test passed successfully!
Cleaning up...
Test completed successfully!
```
its unclear what happened but I have the feeling that running it again could pass? Do we experience flaky in CI sometimes?

@jnovy Sorry to bother you again, could you run the integration test if you get a chance?

src/healthcheck.c
```diff
@@ -0,0 +1,422 @@
...
+{
+	if (container_id == NULL) {
+		return false;
+	}
+
+	const char *status_str = healthcheck_status_to_string(status);
+
+	/* Create simple JSON message for status update */
+	char json_msg[1024];
+	snprintf(json_msg, sizeof(json_msg),
```
jnovy, line 346: I'd remove this JSON part and use just minimalistic conmon protocol - JSON is over-complication here and poses injection risks.

Something like:
```
/* Include exit code in message for debugging when status indicates failure */
char message[256];
if (exit_code != 0 && status == HEALTHCHECK_UNHEALTHY) {
      snprintf(message, sizeof(message), "%s (exit_code: %d)", status_str, exit_code);
      write_or_close_sync_fd(&sync_pipe_fd, status, message);
 } else {
       write_or_close_sync_fd(&sync_pipe_fd, status, status_str);
 }
```

src/healthcheck.c
```diff
@@ -0,0 +1,422 @@
...
+		/* Execute the runtime command */
+		if (execvp(runtime_path, runtime_argv) == -1) {
+			/* If execvp fails, exit with error code */
+			_exit(127); /* Command not found */
+		}
+	} else {
+		/* Parent process - wait for child to complete */
+		close(stderr_pipe[1]); /* Close write end of stderr pipe */
+		int status;
+		pid_t wait_result = waitpid(pid, &status, 0);
```
jnovy, line 273: The `healthcheck_execute_command()` function doesn't enforce the configured timeout parameter. The `waitpid(pid, &status, 0) which waits indefinitely, making the timeout parameter non-functional.

src/healthcheck.c
```diff
@@ -0,0 +1,422 @@
...
+gboolean healthcheck_timer_callback(gpointer user_data)
+{
+	healthcheck_timer_t *timer = (healthcheck_timer_t *)user_data;
+	if (timer == NULL || !timer->timer_active) {
+		return G_SOURCE_REMOVE; /* Stop the timer */
+	}
+
+	/* Check if we're still in start period */
+	if (timer->start_period_remaining > 0) {
+		timer->start_period_remaining -= timer->config.interval;
```
jnovy, line 367: This countdown approach creates timing inconsistencies. If the interval is 30s but start period is 45s, the logic becomes unreliable, use actual elapsed time: `time(NULL) - timer->start_time`

src/healthcheck.c
```diff
@@ -0,0 +1,422 @@
...
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <fcntl.h>
+#include <time.h>
+#include <signal.h>
+#include <glib.h>
+
+/* Healthcheck validation constants */
+#define HEALTHCHECK_INTERVAL_MIN 1
+#define HEALTHCHECK_INTERVAL_MAX 3600
```
jnovy, line 26: Review these constants again - HEALTHCHECK_INTERVAL_MIN/MAX, etc. are defined but never used to validate user input, allowing invalid parameters to pass through.

samifruit514: hey @jnovy , once again, thank you. I've made a bunch of changes; I made the changes from the findings you reported. While working in start_period, noticed that interval was not starting at the beginning, so I fixed that. for the minimalistic conmon protocol, I took id -100 + the health status. Also spent time on "integration" BATS tests. Trying to cover your points above. Im not just sure what you meant by "Resource cleanup verification"

Thanks for your patience.

samifruit514: draft PR updated as well, in case you want to try conmon+podman: https://github.com/containers/podman/pull/27067/files

Luap99:
Personally I don't see the point for this, we are in the process rewriting in rust so I rather not add large features here that increase the scope of the rewrite. 

But also more importantly I don't see how this can help podman here, from a quick look this is calling the oci runtime directly instead of the proper `podman healthcheck run` command. Podman does a lot more as it tracks the metadata, stores the stdout/err and so on. It also has two different timers startup and the regular healtcheck so I really don't see how this here will solve the podman problem at all.

> enabling the healthchecks from conmon to podman, through the unix pipe (the same one that sends the PID to link).

Podman is not a daemon, `podman run -d` exists right away after container start so the pipe will be broken and useless. 
So overall I don't think the design makes sense for podman.

If we want to add healtchcheck to conmon then I think we first need a proper design discussion.

senpro-ingwersenk:
@mheon Thank you very much for your elaboration - I fully understand.

In that case, what'd be the best way to observe the progress of the Rust port itself? Should I keep an eye out on a certain branch or is there some kind of milestone ticket? I would love to keep up to date with that implementation. :)

PR #603, opened 2025-09-18
Add tests for --exec-attach

jankaluza:
This commit adds three new tests for the --exec-attach option. It also
improves start_conmon_with_default_args function so it be executed
multiple times in a single test-case for single container.


jnovy:
LGTM, assuming it doesn't make sense to implement helpers for pipe operations - the pipe operations are one off?

jankaluza:
@jnovy , I will see in the next commit when I add more pipeline operations. I might end up with helpers, but we will see.

PR #605, opened 2025-09-23
Add tests for --terminal and ctrl.c

jankaluza:
This commit adds tests for handling of the tty resize messaages and
log reopen messages sent to ctl fifo.

Signed-off-by: Jan Kaluza <jkaluza@redhat.com>

### DIFF
--- a/test/10-ctrl.bats
+++ b/test/10-ctrl.bats
@@ -1,0 +1,141 @@
+#!/usr/bin/env bats
+
+load test_helper
+
+setup() {
+    check_conmon_binary
+    check_runtime_binary
+    setup_container_env "while [ ! -f /tmp/test.txt ]; do /busybox sleep 0.1; done; /busybox stty size" "true"
+    generate_process_spec "echo 'Hello from exec!' && echo 'Hello there!' > /tmp/test.txt"
+}
+
+teardown() {
+    cleanup_test_env
+}
+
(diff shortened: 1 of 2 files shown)


test/10-ctrl.bats
```diff
@@ -0,0 +1,118 @@
...
+# Helper function to send the resize command. Fails if the resize command
+# does not trigger the tty resize.
+test_resize_command_ok() {
+    local command="$1"
+    test_ctl_command "$command"
+
+    # Check that the main process noticed the /tmp/test.txt.
+    assert_file_exists "$LOG_PATH"
+    run cat "$LOG_PATH"
+    assert "${output}" =~ "2 2"
```
jnovy, line 54: This is wrong? `"2 2"` is hardcoded, shouldn't `test_resize_command_ok()` be invoked with `$1` - command and `$2` - expected size? E.g.:

```
@test "ctrl: resize with leading zeros" {
    test_resize_command_ok "1 0010 0020" "10 20"
}
```

jnovy (changes requested):
It seems that the test suite tests only a very limited edge cases. Maybe something like this would make sense too?

```
@test "ctrl: resize with floating point dimensions" {
    test_resize_command_fail "1 10.5 20.3"
}

@test "ctrl: resize with hex numbers" {
    test_resize_command_fail "1 0x10 0x20"
}
```

PR #606, opened 2025-09-24
Document all CLI options

jnovy:
Some of CLI options were missing from conmon.8.md. This commit adds all available options there.


jnovy:
Seems the removal is the best if it never worked.

jnovy:
I just updated the man page not mentioning these options @giuseppe

giuseppe (approved):
LGTM

PR #607, opened 2025-09-25
Add more tests for ctr_logging.c

jankaluza:
This commit adds tests for:
- the `--log-label` option.
- the `--log-tag` option.
- the parsing of journald priority level.
- the partial message in both journald and k8s-file.
- the log rotation using the `ctl` command.


test/02-ctr-logs.bats
```diff
@@ -74,4 +75,112 @@ run_conmon_with_log_opts() {
...
+@test "ctr logs: journald with --log-tag" {
+    run_conmon_with_default_args \
+        --log-path "journald:" \
+        --log-tag "tag_$CTR_ID"
+
+    run journalctl --user CONTAINER_TAG="tag_$CTR_ID"
+    assert "${output}" =~ "hello from busybox"
+}
+
+@test "ctr logs: journald priority level" {
```
jnovy, line 153: Isn't this a duplicate of what we test in `06-priority-parsing.bats`?

jankaluza: True. I've removed that test.

Issue #38, opened 2019-05-29
implement double buffering scheme for logs


haircommander:
cross post of: https://github.com/cri-o/cri-o/issues/1778

jnovy:
Is there anything to be done on conmon side for this one @haircommander ?

PR #608, opened 2025-09-26
Replace pthread-based logging buffer with signal masking

jnovy:
This commit replaces the pthread mutex-based synchronization in the async logging buffer with signal masking, eliminating the pthread dependency while maintaining thread safety. This should reduce conmon latency overhead by ~15%.

Fixes: #38


src/ctr_logging_buffer.c
```diff
@@ -0,0 +1,375 @@
...
+
+static void flush_log_buffer_locked(void)
+{
+	if (!g_log_buffer || g_log_buffer->count == 0) {
+		return;
+	}
+
+	/* Allocate temporary buffer to avoid race conditions */
+	size_t flush_count = g_log_buffer->count;
+	log_entry_t *temp_entries = malloc(flush_count * sizeof(log_entry_t));
```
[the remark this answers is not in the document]
jnovy: nice catch, let me `_cleanup_free` this

nix/derivation.nix
```diff
@@ -24,7 +24,7 @@ with pkgs; stdenv.mkDerivation rec {
     # Static builds will use PKG_CONFIG_PATH approach instead
   ];
   prePatch = ''
-    export CFLAGS='-static -pthread'
+    export CFLAGS='-static'
```
portante: For the last 4 years this reference to `-pthread` has been there without the need for it?

If so, perhaps this is worthy of pulling out into a separate commit.

src/ctr_logging_buffer.c
```diff
@@ -0,0 +1,482 @@
...
+	/* Try to add to buffer */
+	bool success = add_entry_to_buffer_locked(pipe, buf, num_read);
+
+	if (!success) {
+		/* Buffer full - flush and try again */
+		flush_buffer_contents_locked();
+		success = add_entry_to_buffer_locked(pipe, buf, num_read);
+	}
+
+	if (!success) {
```
portante: This check appears to only be required to check the return value from line 204, as the value of `success` will still be true if it is true at line 201.

Is it here to save indentation?

Can we consider moving this to follow line 204 in that same block?

PR #609, opened 2025-09-29
Restore use of `writev()` system call

portante:
The use of the sequential `write()` system calls to replace `writev()` in commit 29d17be might be simpler to understand, but can lead to unwanted CPU overhead when a high frequency of newlines exists in the data.  The original bugfix for the log file corruption was the proper reset of the `iovcnt` on error.  We restore the reset but always do it as part of the `writev_buffer_flush()` method which avoids "leaking" knowledge of how it works.

See #262 for the effects that behavioral analysis.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@ sudo yum install -y \
   glib2-devel \
   glibc-devel \
   libseccomp-devel \
+  systemd-devel \
   make \
   pkgconfig \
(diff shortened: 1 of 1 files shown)


jnovy:
LGTM, the READM.md can be further amended in the following commit. Integration errors are not related to this PR.

Issue #610, opened 2025-10-01
Test don't run on architectures other than x86_64


ricardobranco777:
I'm getting this from 04-runtime on aarch64 & ppc64le:

```
#/vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
#|     FAIL: 'hello from busybox' found in the log
#| expected: =~ hello from busybox
#|   actual:    2025-10-01T10:25:54.025335792-04:00 stderr F exec container process \`/busybox\`: Exec format error
#\^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

This may be due to the use of an Intel binary for busybox.

ricardobranco777:
I applied this workaround: Install busybox-static and copy it to `/tmp/conmon-test-images/busybox` but now I get:

- aarch64: https://openqa.opensuse.org/tests/5355640
- ppc64le: https://openqa.opensuse.org/tests/5355641

PR #611, opened 2025-10-01
Fix tests on non-x86_64 architectures

jnovy:
Make busybox binary download architecture-aware to fix test failures on aarch64 and ppc64le. Previously, tests hardcoded an x86_64 binary URL which caused "Exec format error" on other architectures.

The fix detects the system architecture and selects the appropriate busybox binary from busybox.net, with support for x86_64, aarch64, ppc64le, i686, and a fallback for other architectures.

Fixes: #610

### DIFF
--- a/test/test_helper.bash
+++ b/test/test_helper.bash
@@ -22,7 +22,31 @@ assert_failure() {
 # Default paths and variables
 CONMON_BINARY="${CONMON_BINARY:-/usr/bin/conmon}"
 RUNTIME_BINARY="${RUNTIME_BINARY:-/usr/bin/runc}"
-BUSYBOX_SOURCE="https://busybox.net/downloads/binaries/1.31.0-i686-uclibc/busybox"
+
+# Detect architecture and set appropriate busybox source
+ARCH=$(uname -m)
+case "$ARCH" in
+    x86_64)
+        BUSYBOX_SOURCE="https://busybox.net/downloads/binaries/1.31.0-i686-uclibc/busybox"
+        ;;
+    aarch64|arm64)
+        BUSYBOX_SOURCE="https://busybox.net/downloads/binaries/1.31.0-defconfig-multiarch-musl/busybox-armv8l"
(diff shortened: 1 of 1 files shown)


ricardobranco777:
It seems to solve the busybox issue but we have other failures:

- aarch64: https://openqa.opensuse.org/tests/5355651
- ppc64le: https://openqa.opensuse.org/tests/5355652
- x86_64: https://openqa.opensuse.org/tests/5355653 (which is weird as it passed here https://openqa.opensuse.org/tests/5355489)

Note: We're backporting these patches to v2.1.13: 579, 597, 602, 603, 604, 605, 611

Given that busybox.net is sometimes down, we'll use this workaround instead:

```
sudo zypper in busybox-static
mkdir /tmp/conmon-test-images
cp /usr/bin/busybox-static /tmp/conmon-test-images/
```

ricardobranco777:
> Given that busybox.net is sometimes down, we'll use this workaround instead:
> 
> ```
> sudo zypper in busybox-static
> mkdir /tmp/conmon-test-images
> cp /usr/bin/busybox-static /tmp/conmon-test-images/
> ```

This workaround doesn't work.  For some reason we need an older BusyBox.  https://openqa.opensuse.org/tests/5355733

ricardobranco777 (approved):

jankaluza:
This code is fragile in general. If you kill the tests while it's downloading, the broken busybox file can end up on fs and it never gets redownloaded until you remove that broken busybox binary. That was something I also wanted to fix one day in the future :-).

Maybe that might be the reason for broken binary @ricardobranco777  sees?

jnovy:
Hmm, thinking more about all this - there doesn't seem any particular reason why we use busybox and it's been unreliable for some time. It's time to switch to something else. I'd propose UBI as the registry from which it's pulled from is rarely down and it's properly tested and functional.

jnovy:
I will create a new PR, let's skip this one.

PR #612, opened 2025-10-02
Switch tests from busybox to UBI10 and add consistent terminal size v…

jnovy:
…alidation

- Update Containerfile to use UBI10 base image instead of busybox
- Replace busybox references with UBI10 commands in all test files
- Add upper bound validation (1000) for terminal window size in ctrl.c to prevent wraparound bugs
- Update CI/CD dependencies to include podman, bats, jq, and socat
- Fix test assertions to expect proper terminal size behavior

Fixes: #610

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -84,7 +84,7 @@ build_task:
         disk: 200
 
     script:
-        - dnf install -y make glib2-devel git gcc pkg-config systemd-devel libseccomp-devel
+        - dnf install -y make glib2-devel git gcc pkg-config systemd-devel libseccomp-devel podman bats jq socat
         - cd $CIRRUS_WORKING_DIR
         - make
         - make test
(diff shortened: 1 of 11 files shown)


jnovy:
@ricardobranco777 @jankaluza This PR changes the tests to run solely on UBI10-micro removing busybox as a whole. This will allow us to actually run with on top of full-featured and tested OS components. Reliability is a nice bonus.

For that I needed to amend few tests and actually found a bug - we don't test upper limit for terminal size and kernel accepts only `unsigned short` via `ioctl(TIOCSWINSZ)`. In order to avoid OOM/DoS via setting a massive terminal sizes I maxed the possible size to be 1000 characters for each dimension.

ricardobranco777:
> @ricardobranco777 @jankaluza This PR changes the tests to run solely on UBI10-micro removing busybox as a whole. This will allow us to actually run on top of full-featured and tested OS components. Reliability is a nice bonus.
> 
> For that I needed to amend few tests and actually found a bug - we don't test upper limit for terminal size and kernel accepts only `unsigned short` via `ioctl(TIOCSWINSZ)`. In order to avoid OOM/DoS via setting a massive terminal sizes I maxed the possible size to be 1000 characters for each dimension.

Unfortunately I can't properly test this PR because we're backporting these to an existing release (1.24) and the tests will fail as it relies on modified code.  It would help if the migration to UBI10 is done in a separate PR to just test that part.

jnovy:
@ricardobranco777  These fixes unfortunately need to happen together because the previous tests were relying on fake busybox utility behaviour (not receiving a TTY). With UBI we got TTY so we can check live values. This is needed change otherwise tests in UBI10 would fail.

PR #614, opened 2025-10-08
Remove need for extra bytes for remote consoles

portante:
We use `writev()` for remote consoles to remove the need for extra bytes surrounding the data to be written.

We really don't want to "leak" the protocol requirement for remote consoles to the rest of the code base.  To hide that, and avoid two `write_all()` method calls in a row, we promote `writev_buffer_flush()` to a `utils` method called `writev_all()` to leverage I/O vectors in one call (memory copies will need to occur for remote sockets, so using `writev()` avoids that).

Further, we use the `g_ptr_array_foreach()` method instead of doing it ourselves to avoid breaking the `GPtrArray` encapsulation.

Signed-off-by: Peter Portante <peter.portante@redhat.com>

----

FWIW, this PR is a pre-requisite to #615, which in turn will make it easier to finally implement #264.

### DIFF
--- a/src/conn_sock.c
+++ b/src/conn_sock.c
@@ -39,7 +39,7 @@ static char *setup_socket(int *fd, const char *path);
   setup_attach_socket() is responsible for setting the correct remote FD and pushing it onto the queue.
 */
 static struct local_sock_s local_mainfd_stdin = {&mainfd_stdin, true, NULL, "container stdin", NULL};
-struct remote_sock_s remote_attach_sock = {
+static struct remote_sock_s remote_attach_sock = {
 	SOCK_TYPE_CONSOLE,   /* sock_type */
 	-1,		     /* fd */
 	&local_mainfd_stdin, /* dest */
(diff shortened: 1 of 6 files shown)


jnovy:
Would it be possible to make it atomic?

jnovy (changes requested):
Also please squash both commits.

src/utils.c
```diff
@@ -80,6 +80,59 @@ ssize_t write_all(int fd, const void *buf, size_t count)
...
+		} while (res == -1 && errno == EINTR /*retryable_error(errno)*/);
+
+		if (res <= 0) {
+			/*
+			 * Any unflushed data is lost (this would be a good place to add a counter for how many times
+			 * this occurs and another count for how much data is lost).
+			 *
+			 * Note that if writev() returns a 0, this logic considers it an error.
+			 */
+			ntracef("writev_all(fd = %d, iovcnt = %lu) errno = %d", fd, iovcnt, errno);
```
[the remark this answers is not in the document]
portante: Yes, removed.  Thanks for noticing.

TomSweeneyRedHat:
LGTM
but would like a head nod from @jnovy

src/conn_sock.c
```diff
@@ -349,27 +349,32 @@ char *socket_parent_dir(gboolean use_full_attach_path, size_t desired_len)
...
 	}
 }
 
+void write_back_to_remote_consoles(stdpipe_t pipe, char *buf, size_t buflen)
+{
+	if (local_mainfd_stdin.readers == NULL)
+		return;
+	struct iovec iov[2] = {{&pipe, 1}, {buf, buflen}};
+	writev_iov_t wviov = {2, 2, iov};
+	g_ptr_array_foreach(local_mainfd_stdin.readers, write_remote_sock, &wviov);
```
jankaluza: The `wviov` is created few lines above, passed as pointer down to `write_remote_sock` for each `local_mainfd_stdin.reader`. I think the issue is that the first `writev_all` call changes the `wviov` and the second call of `writev_all` for the second reader from `local_mainfd_stdin.readers` will get a `wviov` with `buf->iovcnt = 0;`, effectively writing nothing.

I think we will have to rebuild the writev_iov_t in the `write_remote_sock` or change the `writev_all` to operate on a local copy and do not mutate it.

I think it would be also great to add a bats test for two readers, so this use-case is covered.

jankaluza: I think another issue here is that change of order of array iteration.

Previously, the `local_mainfd_stdin.readers` was iterated in the reverse order. I believe the reason was that the `remote_sock_shutdown` can remove the reader from the array (It calls `g_ptr_array_remove`).

The proposed code uses `g_ptr_array_foreach`. The documentation for this says `func must not add elements to or remove elements from the array.`, but this is what can actually happen in the `remote_sock_shutdown`.

I think the way forward is to keep the reverse iteration.

kolyshkin:
@portante that's the current status of this? Needs to be rebased at the very least.

PR #615, opened 2025-10-08
Use the actual pipe size instead of 8,192

portante:
The default for linux pipes is typically something much larger than 8,192 bytes.
We take advantage of this to fact to avoid multiple rounds of reading from a
full pipe.

The constant, `STDIO_BUF_SIZE` is removed in favor of `fcntl(F_GETPIPE_SZ)`
calls for pipes / fifos, an adding `DEF_STDIO_BUF_SIZE` for the case where
`stdout` is not a PIPE (perhaps character special file) using a 64K buffer to
consume more data per system when possible.

Signed-off-by: Peter Portante <peter.portante@redhat.com>

----

The goal of this work is to make it much easier to implement #264.

### DIFF
--- a/src/conmon.c
+++ b/src/conmon.c
@@ -165,6 +165,10 @@ int main(int argc, char *argv[])
 				pexit("Failed to create !terminal stdin pipe");
 
 			mainfd_stdin = fds[1];
+			ret = fcntl(mainfd_stdin, F_GETPIPE_SZ);
+			if (ret < 0)
+				pexit("main stdin pipe size determination failed");
+			mainfd_stdin_size = (size_t)ret;
 			workerfd_stdin = fds[0];
 
 			if (g_unix_set_fd_nonblocking(mainfd_stdin, TRUE, NULL) == FALSE)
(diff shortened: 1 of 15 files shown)


src/conmon.c
```diff
@@ -194,6 +202,13 @@ int main(int argc, char *argv[])
 		pexit("Failed to create stderr pipe");
 
 	mainfd_stderr = fds[0];
+	ret = fcntl(mainfd_stderr, F_GETPIPE_SZ);
+	if (ret < 0)
+		pexit("main stderr pipe size determination failed");
+	mainfd_stderr_size = (size_t)ret;
+	if ((mainfd_stdout >= 0) && (mainfd_stderr_size != mainfd_stdout_size)) {
```
jankaluza: Out of curiosity, why do we care about the sizes to be the same?

jnovy (changes requested):

kolyshkin:
@portante is this still a draft? Also needs to be rebased.

PR #619, opened 2025-10-27
Fix remaining busybox references in tests

jnovy:
Complete the migration from busybox to ubi10-micro by removing remaining busybox references that were missed in the initial switch. This updates test assertions and commands to use standard utilities available in ubi10-micro.


jnovy:
Merging this one to unbreak the CI - it is trivial fix.

PR #621, opened 2025-10-29
Check memory.events file exists before adding inotify watch

### LINKED ISSUE #620: Check if memory.events file exists before attempting to add inotify watch
I run a rootless podman container on OpenSUSE MicroOS. The container is working fine although I get the following warning every 30s when conmon tries to check the `memory.events` file for the container

```
conmon 6d47ba2f69f3682e855c <nwarn>: Failed to add inotify watch for /sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/user.slice/user-libpod_pod_....slice/libpod-....scope/memory.events
```

The folder exists, but not the `memory.events` file. I suspect this is because
`/sys/fs/cgroup/user.slice/cgroup.subtree_control` (and below) only contains `pids` but not `memory` like `/sys/fs/cgroup/cgroup.subtree_control` does.

I don’t think that I do anything special when launching the container, so I’m surprised that I get these warnings. Is it correct that conmon assumes memory to be enabled for user.slice? Is it MicroOS that is unusual in this?

_Originally posted by @andreasjarliden in https://github.com/containers/conmon/discussions/616_

jnovy:
Fixes #620

### DIFF
--- a/src/cgroup.c
+++ b/src/cgroup.c
@@ -124,7 +124,11 @@ static void setup_oom_handling_cgroup_v2(int pid)
 	}
 
 	if (inotify_add_watch(ifd, memory_events_file_path, IN_MODIFY) < 0) {
-		nwarnf("Failed to add inotify watch for %s", memory_events_file_path);
+		if (errno == ENOENT) {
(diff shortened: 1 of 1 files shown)


giuseppe (approved):
LGTM

PR #623, opened 2025-11-05
Add an easy cleanup for tests

portante:
The `make clean` target now cleans up the `/tmp` area a bit.  It leaves around the cache of the `busybox` image to avoid fetching it too often.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -106,6 +106,7 @@ docs:
 .PHONY: clean
 clean:
 	rm -rf bin/ src/*.o src/*.gcno src/*.gcda *.gcov
+	$(MAKE) -C test clean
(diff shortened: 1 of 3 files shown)


jnovy:
Maybe better approach would be to just use podman for image caching without any usage of `/tmp`?

Benefits:
- No /tmp files to manage at all
- Podman's image cache is the single source of truth
- make clean only needs to clean actual test artifacts
- No cache directory needed

This would need: Removal of `ROOTFS_CACHE_DIR` and `ROOTFS_CACHE_MARKER` variables, `cache_ubi10_rootfs()` to rely solely on podman's built-in image cache and removal of redundant marker file creation and checks.

Does it make sense?

PR #625, opened 2025-12-02
Fix SIGABRT crash in drop_signal_event

jnovy:
Replace g_assert_cmpint with proper error handling to prevent abort when signalfd read fails.

Fixes: #2348843


giuseppe:
> Fixes: #2348843

this seems not correct.

Code LGTM

PR #629, opened 2026-02-03
Fix k8s-file log format for terminating F-sequence

jnovy:
The k8s-file format requires the space separator between tag and content to be present even when content is empty. The terminating F-sequence generated for partial log lines was missing this mandatory space, causing podman logs to fail with "is not a valid container log line".

Changed terminating F-sequence from "F\n" to "F \n" and updated byte counts accordingly. Fixed test assertions to validate the correct format.

Fixes containers/podman#28014

### DIFF
--- a/src/ctr_logging.c
+++ b/src/ctr_logging.c
@@ -510,14 +510,14 @@ static int write_k8s_log(stdpipe_t pipe, const char *buf, ssize_t buflen)
 
 		if (writev_buffer_append_segment(k8s_log_fd, &bufv, tsbuf, TSBUFLEN - 1) >= 0) {
 			timestamp_written = true;
-			if (writev_buffer_append_segment(k8s_log_fd, &bufv, "F\n", 2) >= 0) {
+			if (writev_buffer_append_segment(k8s_log_fd, &bufv, "F \n", 3) >= 0) {
 				f_sequence_written = true;
 			}
 		}
 
 		if (timestamp_written && f_sequence_written) {
-			k8s_bytes_written += TSBUFLEN - 1 + 2;
-			k8s_total_bytes_written += TSBUFLEN - 1 + 2;
+			k8s_bytes_written += TSBUFLEN - 1 + 3;
+			k8s_total_bytes_written += TSBUFLEN - 1 + 3;
 		} else {
 			if (!timestamp_written) {
 				nwarn("failed to write timestamp for terminating F-sequence");
--- a/test/09-partial-log-test.bats
+++ b/test/09-partial-log-test.bats
@@ -80,7 +80,7 @@ teardown() {
             return 1
         }
 
-        echo "$log_content" | grep -q "stdout F$" || {
+        echo "$log_content" | grep -q "stdout F $" || {
             echo "Expected F-sequence not found"
             return 1
         }
@@ -133,7 +133,7 @@ teardown() {
 
     # For normal output with newlines, should NOT have standalone F-sequences
     # (F-sequences should only appear for partial line termination)
-    ! echo "$log_content" | grep -q "stdout F$" || {
+    ! echo "$log_content" | grep -q "stdout F $" || {
         echo "Unexpected standalone F-sequence found for normal output"
         return 1
     }


jnovy:
@ricardobranco777 PTAL

Luap99:
> I think it works. The podman e2e test fails differently now:
> 
> https://openqa-assets.opensuse.org/tests/5649499/file/podman_e2e-localintegration.txt
> 
> ```
>   [FAILED] Timed out after 0.042s.
>   The function passed to Eventually failed at /var/tmp/podman/test/e2e/logs_test.go:465 with:
>   Expected
>       <string>: "...a72be
>       "
>   to equal               |
>       <string>: "...a72be"
>   In [It] at: /var/tmp/podman/test/e2e/logs_test.go:466 @ 02/04/26 07:31:55.76
> 
>   Full Stack Trace
>     github.com/containers/podman/v5/test/e2e.init.func45.21()
>     	/var/tmp/podman/test/e2e/logs_test.go:466 +0x3fc
>   < Exit [It] podman logs partial log lines: json-file - /var/tmp/podman/test/e2e/logs_test.go:447 @ 02/04/26 07:31:55.76 (458ms)
> ```

well this is exactly what I have mentioned above, this is a behavioural change by injecting an additional newline, we do explicitly test that this is not the case

```
			// Important: do not use OutputToString(), this will remove the trailing newline from the output.
			// However this test must make sure that there is no such extra newline.
			Expect(string(logc.Out.Contents())).To(Equal(content))
```

so if conmon now does this we must chnage the podman test to allow that, but I like to actually have some input from others if conmon doing this now is really sane

At the very least this here makes the log drivers behave differently!

```
$ podman --conmon /home/pholzing/CODE/conmon/bin/conmon  run --name test --log-driver journald  quay.io/libpod/testimage:20241011 printf "hola" && echo ---; podman logs test; echo ---
test
hola---
hola---
$ podman --conmon /home/pholzing/CODE/conmon/bin/conmon  run --name test --log-driver k8s-file  quay.io/libpod/testimage:20241011 printf "hola" && echo ---; podman logs test; echo ---
test
hola---
hola
---
```

IMO it would be best to just revert the original fix instead and no do that to preserve the behavior.

jnovy:
The CRI spec text you quoted doesn't explicitly define container termination behavior - you're correct. However, the spec's scope is limited to runtime-initiated line splitting, not container termination scenarios. The evidence that this behavior is expected comes from filed bugs against both conmon and containerd showing that **missing terminating F-sequences break log consumers in production**:

**Conmon issue #252**: Reports that running `echo blah | tr '\n' ' '` produces only:
```
2021-03-29T21:16:20.753110058-04:00 stdout P blah
```

The bug report explicitly states the expected behavior should be:
```
2021-03-29T21:16:20.753110058-04:00 stdout P blah
2021-03-29T21:16:20.753110059-04:00 stdout F
```

and concludes: *"conmon should be writing out a terminating F-sequence into the log if it sees a container pipe closing without a trailing newline"* what I consider justified.

**Containerd issue #4813**: Shows containerd has the identical bug. When running `printf "Hello from init container!"`, containerd outputs:
```
2020-12-07T15:32:38.994574754Z stdout P Hello from init container!
```
and stops. This breaks fluentd's concat plugin with timeout errors.

**Grafana Loki issue #10998**: Documents that when a container outputs `echo -n "fuga"` and terminates, promtail's CRI stage cannot flush the partial line. It remains buffered indefinitely (100+ minutes) until the buffer is forcibly flushed, at which point Loki may reject it as too old.

Regarding driver inconsistency - the journald driver doesn't inject the newline because it writes directly to journald using native journald fields, bypassing the CRI log format entirely. The k8s-file driver must comply with CRI format to be parseable by kubectl/log consumers.

This is a pre-existing architectural difference: journald has always had different semantics than file-based drivers. The proper solution would be documenting this difference or making journald also emit CRI-compliant logs (which would require significant rework and I'm not planning to do that).

**Sources:**
- Missing terminating F-sequence in conmon: https://github.com/containers/conmon/issues/252
- CRI wrong partial log line in containerd: https://github.com/containerd/containerd/issues/4813
- Promtail partial line flushing issue: https://github.com/grafana/loki/issues/10998

PR #654, opened 2026-05-04
ci: replace Cirrus CI with GitHub Actions

timcoding1988:
cirrus-ci is out of support by June. we are replacing cirrus-ci with native GHA


.github/workflows/packaging.yml
```diff
@@ -0,0 +1,54 @@
+name: packaging
```
lsm5: I think we can remove this entire `packaging.yml` . Packit already handles packaging, so there's no need for this.

lsm5 (approved):
LGTM with nit to remove `packaging.yml`. 

Also, I think we should not even have GHA as everything should be manageable via Packit. This repo isn't moving to CNCF, is it?

lsm5:
@saschagrunert @haircommander @jnovy PTAL

jnovy:
@timcoding1988 I agree with @lsm5 on removing `packaging.yml` - Packit already handles that.

Regarding @lsm5's broader question about whether we need GHA at all: Packit covers packaging/COPR builds, but it doesn't run integration tests, coverage, or static binary builds. So we do still need GHA (or some  CI) for those. The existing `integration.yml` and `validate.yml` already handle the build/test and lint jobs that Cirrus had, so this PR  correctly focuses on the remaining gaps: coverage and static builds.

**coverage.yml:** `hack/github-actions-setup` installs a lot of things that `make test-coverage` doesn't need (runc, crun, cri-tools compiled from source, CRIU, etc.). The original Cirrus coverage task only installed build deps + gcovr. Consider replacing it with targeted apt installs:

```
- name: Install dependencies
run |
  sudo apt-get update
  sudo apt-get install -y gcovr libglib2.0-dev libseccomp-dev libsystemd-dev pkg-config
```

This would cut the job time significantly.

**static.yml:** The `CACHIX_AUTH_TOKEN` secret needs to be configured in the repo settings (Settings/Secrets/Actions) for the Nix cache to work. Worth noting in the PR description so whoever merges this knows to set it up. Also, for consistency with the other workflows, consider adding an `all-done` gate job.

jnovy (changes requested):
Requesting changes as per comment above.

lsm5:
> Packit covers packaging/COPR builds, but it doesn't run integration tests, coverage, or static binary builds. So we do still need GHA (or some CI) for those.

Sorry, I meant we _could_ do those things via Packit and Testing Farm itself, so there shouldn't be any need for GHA as such. But since @timcoding1988 has it in place already with this PR, we could leave all that for later.

.github/workflows/packaging.yml
timcoding1988: removed. correct this is not going to CNCF.

jnovy:
LGTM, nice work! I will continue to set the `CACHIX_AUTH_TOKEN` secret will need to be configured in the repo settings.

jnovy:
@saschagrunert seems that you own the "conmon" cache on cachix - would you mind to update the auth token here?

PR #658, opened 2026-05-18
signal: replace raise(SIGUSR1) with self-pipe to fix SIGABRT on glibc >= 2.42

### LINKED ISSUE #657: conmon 2.2.1 SIGABRT in __syscall_cancel during ppoll (94 crashes in 4 days, deterministic)
## Bug Description

`conmon` 2.2.1-2.fc43 crashes deterministically with **SIGABRT** in `__internal_syscall_cancel()` while inside `ppoll()` during the main GLib event loop. **94 crashes captured in 4 days** (12-16 May 2026) on a single rootless Podman host, all with **identical stack traces** at identical binary offsets.

The host container appears as `Up X hours (healthy)` in `podman ps`, but the conmon monitor process is dead. `podman exec <container> true` then fails with `OCI runtime error: container is not running`, and `podman restart` fails with `conmon process killed`. Only `podman rm -f` + recreate recovers.

This is not OOM (host has 182 GB free, monitored container uses ~40 MB).

## Stack Trace (identical across all 94 coredumps)

```
PID: <varies>
Signal: 6 (ABRT)
Executable: /usr/bin/conmon
Package: conmon/2.2.1-2.fc43
Build-ID: 69e402d7183efd44fb1bd32b1c7fa5d4fb84aa8e

Stack trace of thread <main>:
#0  0x00007f...bbe __internal_syscall_cancel (libc.so.6 + 0x6ebbe)
#1  0x00007f...be4 __syscall_cancel           (libc.so.6 + 0x6ebe4)
#2  0x00007f...686 ppoll                      (libc.so.6 + 0xe8686)
#3  0x00007f...3a0 g_main_context_iterate_unlocked.isra.0 (libglib-2.0.so.0 + 0x4a3a0)
#4  0x00007f...55f g_main_loop_run            (libglib-2.0.so.0 + 0x4a55f)
#5  0x000055...19d main                       (/usr/bin/conmon + 0x419d)
#6  0x00007f...5b5 __libc_start_call_main     (libc.so.6 + 0x35b5)
#7  0x00007f...668 __libc_start_main@@GLIBC_2.34
#8  0x000055...785 _start                     (/usr/bin/conmon + 0x5785)
```

Binary offsets (`0x6ebbe`, `0x6ebe4`, `0xe8686`, `0x4a3a0`, `0x4a55f`, `0x419d`, `0x5785`) are **byte-for-byte identical** across all 94 coredumps — confirming this is a single, deterministic codepath, not random memory corruption.

## Environment

| Component | Version |
|---|---|
| conmon | 2.2.1-2.fc43.x86_64 |
| podman | 5.8.2-1.fc43.x86_64 |
| crun | 1.27.1-1.fc43.x86_64 |
| glibc | 2.42 |
| glib2 | 2.86.5-1.fc43.x86_64 |
| systemd | 258.7-1.fc43 |
| Kernel | 7.0.7-100.fc43.x86_64 |
| Distro | Fedora 43 Server (x86_64) |
| User | rootless (UID 1000) |
| Cgroup | systemd, v2 |

Container being monitored: a Go binary (~40 MB resident) running a long-lived process with WebSocket connections. Open-source HTTP wrapper around `tulir/whatsmeow`. Possibly relevant: `tulir/whatsmeow#1004` describes Go-side lock contention that could leave the monitored process slow to handle signals — speculative correlation with the conmon trigger.

## Frequency / Temporal Pattern

- **94 crashes between 2026-05-12 and 2026-05-16** (4 days), one host.
- Crashes cluster heavily during night hours (02:00–08:00 local time), correlating with:
  - Scheduled PostgreSQL backups (~03:00)
  - Low application traffic
- **Distinctive pattern: crash clusters every exact 2 minutes**, e.g.:
  ```
  07:17:39, 07:19:39, 07:21:39, 07:23:39, 07:25:39
  08:26:39, 08:28:39, 08:30:39, 08:32:40, 08:34:40
  ```
  This 2-minute periodicity strongly suggests an external trigger, not internal conmon timer.

## Suspected Trigger

A user-space watchdog runs `podman exec <container> true` every 60 seconds to detect "zombie" state (container marked Up but unresponsive). When the container is healthy, exec succeeds and conmon stays alive. When the container enters a degraded state (cause unknown — possibly related to `tulir/whatsmeow#1004` LID lock contention on the monitored Go workload), the next `podman exec` triggers `__syscall_cancel` race in conmon's `ppoll` loop and conmon SIGABRTs.

Once conmon is dead, `podman restart` fails with `conmon process killed`. Only `podman rm -f` + recreate (which spawns a new conmon for the new container) recovers.

The 2-minute spacing in clusters likely reflects: (1) watchdog detects zombie via exec, (2) attempts `podman restart` (which fails because conmon is killed), (3) cleanup leaves a half-recovered state, (4) watchdog next cycle detects it again, repeats. Crashes stop only when out-of-band manual `podman rm -f` + recreate is performed.

## Reproducer (not yet minimal)

Not yet isolated. Suspicion is that sustained `podman exec` against a container whose internal process is slow to respond to signals can trigger this. Reproducer attempts to follow.

## What I've Tried

- ABRT confirmed signal is SIGABRT (signal 6), not SEGV.
- Verified binary integrity (`rpm -V conmon` clean).
- Implemented exponential backoff + automatic recreate in our watchdog to recover within ≤60s (workaround).
- Tried `--init` (tini as PID 1 in the monitored container) — no impact on conmon stability.
- Considering switch `crun` → `runc`, upgrade Fedora 43 → 44 (glibc 2.43+), but neither addresses the conmon `ppoll` cancellation race directly.

## Asks

1. Is this a known conmon issue? I couldn't find an exact match — searched issues for `__syscall_cancel`, `ppoll`, `SIGABRT`, `zombie container` (closest hits #501, #625, #632 are all different bugs).
2. Is conmon's main loop expected to handle `EINTR` / cancellation during `ppoll` more gracefully, or is the assumption that `ppoll` will never be cancelled from outside?
3. Would patching `g_main_context_iterate_unlocked` invocation in conmon to mask cancellation during the ppoll syscall be appropriate? Happy to attempt a PR if pointed at the right area.
4. Any guidance on capturing additional diagnostics (e.g., `strace` on the live conmon before the crash) would be appreciated.

I can provide:
- Full coredump (`.zst`, ~65 KB each) on request — no client data, just container monitoring state
- `journalctl` logs around crash events
- Anonymized timeline of all 94 crash events

Thank you.

jnovy:
### Problem

When `on_sig_exit()` cannot forward a signal (SIGTERM/SIGQUIT/SIGINT) to the container process because `kill()` fails, it needs to wake the GLib main loop. Previously this was done via `raise(SIGUSR1)`.

On **glibc >= 2.42**, calling `raise()` from a signal handler while the main thread is in `ppoll()` triggers glibc's `__syscall_cancel` mechanism, causing conmon to abort with **SIGABRT**.

### Fix

Replace `raise(SIGUSR1)` with the **self-pipe trick** — an async-signal-safe pattern:
- Create a pipe at startup, register a GLib IO source on the read end
- In the signal handler, write one byte to the pipe instead of calling `raise()`
- The callback drains the pipe and triggers `check_child_processes()`

The self-pipe logic is extracted into a reusable module (`src/self_pipe.h` / `src/self_pipe.c`) with `self_pipe_init()`, `self_pipe_wake()`, and `self_pipe_fini()`. This makes it available for any future signal handler that needs to wake the main loop safely.

Fixes: https://github.com/containers/conmon/issues/657


giuseppe (approved):
LGTM after the CI failure is addressed

PR #668, opened 2026-08-13
test: don't let a failed image pull look like a passing test run

kolyshkin:
While debugging a runc CI failure ([opencontainers/runc#5399][1]) I found
that the conmon test suite can report success without having run a single
container test.

Each container test pulls the UBI10-micro image and exports its rootfs in
its own `setup()`, so a full run does dozens of registry round-trips --
all of them concurrent under `--jobs`. When the pull fails, the failure is
reported with `skip()`, so bats prints `ok`, `run-tests.sh` prints
"All tests passed!", and the job goes green.

That is exactly what happened in runc's CI: on the runs where the pull
succeeded, all 38 container tests ran and failed (for an unrelated,
environmental reason); on the runs where the pull failed, everything was
skipped and the job passed. A deterministic breakage looked like a ~50%
flake.

The three commits here:

1. Move the pull and rootfs export into `setup_suite()`, which runs once
   per run and, unlike a per-test `skip`, aborts the whole run when it
   fails. The rootfs is kept as a tarball that each test extracts into its
   own bundle, so tests stay independent. podman's output is no longer
   sent to `/dev/null` -- when obtaining the image fails, the reason is the
   whole point. `UBI10_MICRO_IMAGE` becomes overridable, for a local mirror
   or for testing this failure path.

2. Turn infrastructure failures in `06-exec-exit-status.bats` ("podman is
   here but the container could not be created") into hard failures.
   `skip()` should mean "this environment can't run this test by design",
   not "our fixture broke".

3. Add `--strict` / `CONMON_TEST_STRICT=1`, which turns every skip into a
   failure, so CI notices when tests silently stop running.

Tested locally against runc, rootless:

- full suite with `--strict -j 4`: 119/119 pass
- with `UBI10_MICRO_IMAGE=localhost/no-such-image:nope`:
  `not ok 1 setup_suite`, no tests executed, exit 1, and podman's actual
  error is printed. Before this series, the same situation was a silent
  all-green run.

[1]: https://github.com/opencontainers/runc/issues/5399

### DIFF
--- a/test/06-exec-exit-status.bats
+++ b/test/06-exec-exit-status.bats
@@ -46,22 +46,23 @@ teardown() {
     local conmon_path="$(dirname "$CONMON_BINARY")/conmon"
 
     if [ ! -f "$conmon_path" ]; then
-        skip "conmon binary not found for integration testing at $conmon_path"
+        die "conmon binary not found for integration testing at $conmon_path"
     fi
 
-    # Check if we can create a simple container for testing
-    if ! timeout 10 podman --conmon $conmon_path run --rm registry.access.redhat.com/ubi10/ubi-micro:latest true >/dev/null 2>&1; then
-        skip "Cannot create test containers with podman"
+    # Check if we can create a simple container for testing.
+    run timeout 10 podman --conmon $conmon_path run --rm "$UBI10_MICRO_IMAGE" true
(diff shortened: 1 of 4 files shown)


jnovy:
LGTM, thanks

PR #669, opened 2026-08-13
ci: fix runc build by disabling libpathrs build tag

jnovy:
runc v1.5.1 added `libpathrs` to its default build tags (`BUILDTAGS := seccomp urfave_cli_no_docs libpathrs`), which requires the `libpathrs` C library to be installed on the system. Since `libpathrs-dev` is not available in the Ubuntu runner environment, the CI fails at the `install_runc` step with:

```
Package pathrs was not found in the pkg-config search path.
Package 'pathrs', required by 'virtual:world', not found
make: *** [Makefile:85: runc-bin] Error 1
```

This fix passes `RUNC_BUILDTAGS=-libpathrs` to `make`, which uses runc's built-in mechanism (Makefile lines 15-19) to remove `libpathrs` from the build tags. runc falls back to a pure-Go path resolver when `libpathrs` is not enabled.

Fixes the CI failure introduced by the runc v1.4.2 to v1.5.1 bump.

### DIFF
--- a/hack/github-actions-setup
+++ b/hack/github-actions-setup
@@ -128,7 +128,7 @@ install_cni_plugins() {
 install_runc() {
     git clone --depth 1 --branch "${VERSIONS["runc"]}" https://github.com/opencontainers/runc
     pushd runc
-    make
+    make RUNC_BUILDTAGS=-libpathrs
     sudo install -D -m0755 runc /usr/sbin/runc
     popd
     rm -rf runc


kolyshkin (approved):
LGTM

kolyshkin:
> Run sudo hack/github-actions-setup
> sudo: hack/github-actions-setup: command not found

Hmm. It seems to worked before; wonder why it started failing

kolyshkin:
> Hmm. It seems to worked before; wonder why it started failing

Ah! @jnovy looks like you've removed the `x` bits; see https://github.com/containers/conmon/pull/669.diff:

```
diff --git a/hack/github-actions-setup b/hack/github-actions-setup
old mode 100755
new mode 100644
index cb9de090..079ee18d
--- a/hack/github-actions-setup
+++ b/hack/github-actions-setup
```

kolyshkin:
@jnovy I forced-pushed here to fix the file mode issue.

PR #653, opened 2026-04-03
meson: fix deprecation warnings

acsezen:
## Summary
- add explicit `check` arguments to `run_command()` calls
- replace deprecated `meson.source_root()` with `meson.project_source_root()`
- raise the declared minimum Meson version to `0.56` to match the APIs already in use

## Testing
- `meson setup build-pr-check .`

PR #670, opened 2026-08-13
Fix meson.build, add CI job to validate it

kolyshkin:
Carrying #653.

The meson build is not covered by CI at all -- `validate` runs `make fmt`, `coverage` and `integration` build via the Makefile, and `static` builds via nix. Nothing invokes meson, so `meson.build` can silently rot between releases. That is how the deprecated `run_command` and `meson.source_root()` usages fixed in #653 accumulated in the first place.

This PR:

1. Carries @acsezen's commit from #653 (`meson: fix deprecation warnings`) unchanged.
2. Adds a `meson-build` job to the `validate` workflow, plus a `/build/` entry in `.gitignore` so the meson build directory does not dirty local trees.
3. Switches `-Os`, `-Wall` and `-Werror` from raw `c_args` to the corresponding meson built-in options, which silences the last three warnings and lets the CI job run with `--fatal-meson-warnings`. Future deprecations now fail the build instead of accumulating unnoticed.

The compiler flags are unchanged in effect. Before:

```
-Wall -Winvalid-pch -std=c99 -O0 -g -Os -Wall -Werror
```

After:

```
-Wall -Winvalid-pch -Werror -std=c99 -Os -g
```

The duplicate `-Wall` collapses (`warning_level=1` already emitted one), and `-O0` from the default debug buildtype disappears since it was being overridden by the trailing `-Os` anyway.

One deliberate behavior change worth reviewer attention: because these are now `default_options`, they become overridable -- `-Doptimization=2` and `-Dwerror=false` are honored, whereas `add_project_arguments` forced them unconditionally. This is generally desirable for distro packaging, which often needs `-Werror` off to survive new compiler releases, but it is a real change rather than a pure refactor.

Verified locally from a clean clone with meson 1.11.2 and gcc 16.1.1: `meson setup --fatal-meson-warnings build` exits 0 with no warnings or notices, `ninja` links `conmon` under `-Werror`, and the binary reports the correct version and commit.

---

**Update**: rebased onto main, and two more commits added.

4. `meson: drop duplicate ctr_logging sources` -- `src/ctr_logging.c` and `src/ctr_logging.h` were listed twice in the `executable()` source list. Meson deduplicates them silently today, but with `--fatal-meson-warnings` now in CI, a future meson release warning about duplicate sources would break the job. Reported by @jnovy.

5. `meson: drop seccomp notify leftovers` -- the rebase promptly proved the point of this PR. 9a57dda ("Drop seccomp notify support", #671) removed `src/seccomp_notify.c` and `src/seccomp_notify.h` but not the corresponding `meson.build` bits, so the meson build on main is currently broken:

```
meson.build:59:0: ERROR: File src/seccomp_notify.c does not exist.
```

Removed the two sources, the `libseccomp` dependency and the `-DUSE_SECCOMP=1` define, plus the now-unused `libdl` lookup (`seccomp_notify.c` was the only user of `dlopen`/`dlsym`). This matches the Makefile, which links against glib and, optionally, libsystemd only.

This is exactly the kind of rot the new CI job exists to catch: nothing in CI built with meson, so the breakage sailed through #671 unnoticed.

Re-verified locally with meson 1.11.2 and gcc 16.1.1: `meson setup --fatal-meson-warnings` is clean, `ninja` links `conmon`, and `ldd` shows neither libseccomp nor libdl.


kolyshkin:
Rebased on top of just-merged #669

jnovy (approved):
LGTM. All three commits are correct - deprecation fixes, CI job, and built-in option migration are well done.

One suggestion: `meson.build` has `src/ctr_logging.c` and `src/ctr_logging.h` listed twice in the `executable()` source list (pre-existing, not introduced by this PR). Meson silently deduplicates today, but since this PR adds `--fatal-meson-warnings` to CI, a future meson release that warns about duplicate sources would break the new job. Might be worth adding a 4th commit to deduplicate while you are already touching this file.

kolyshkin:
> One suggestion: `meson.build` has `src/ctr_logging.c` and `src/ctr_logging.h` listed twice in the `executable()` source list (pre-existing, not introduced by this PR). Meson silently deduplicates today, but since this PR adds `--fatal-meson-warnings` to CI, a future meson release that warns about duplicate sources would break the new job. Might be worth adding a 4th commit to deduplicate while you are already touching this file.

Makes sense, added (as 4th commit).

kolyshkin:
Added one more commit to fix meson after PR #671 (which broke it).

kolyshkin:
This only touches meson and meson-build ci job which is green, so merging

PR #671, opened 2026-08-15
Drop seccomp notify support

### LINKED ISSUE #275: errors occurred during building

### LINKED ISSUE #591: seccomp code review
There is broken logic in `conmon.c` `USE_SECCOMP` macro usage and `seccomp_notify.c` as found in issue #582. The purpose of this ticket is to investigate how we move forward with seccomp support in conmon.

kolyshkin:
As [suggested by @giuseppe](https://github.com/containers/conmon/pull/656#issuecomment-3387637838) in #656, completely drop conmon's seccomp notification code. It has never really worked, is not used by anything, and re-enabling it only brings regressions. It is easy to add back if we ever need it.

Worth spelling out how long it has been dead: the feature was effectively unreachable for about five years. 3a8c9138c8 ("make: only define use_seccomp if we're using it", released in v2.0.30, June 2021) inverted the conditional in `main()`, so:

* **built with libseccomp** (the usual distro build): `--seccomp-notify-socket` bailed out immediately with `seccomp support not present`;
* **built without libseccomp**: conmon set up the listener socket, then died in the stub `seccomp_accept_cb` (`seccomp support not available`) as soon as the runtime connected.

The code was still compiled and linked the whole time, just never reachable -- and nobody noticed until eef084b561 ("seccomp: fix inverted notify socket setup logic", #656) restored the conditional, at which point the regressions showed up.

Removed:
* the `--seccomp-notify-socket` and `--seccomp-notify-plugins` options, and the `CONMON_SECCOMP_NOTIFY_PLUGINS` environment variable;
* the plugin machinery (`src/seccomp_notify.[ch]`, `src/seccomp_notify_plugin.h`) and the seccomp listener socket;
* the libseccomp build dependency and its detection (`hack/seccomp-notify.sh`, `USE_SECCOMP`), including from the README, spec files, nix derivation, and CI.

The second commit is a follow-up cleanup: `setup_socket` was only ever called with a non-NULL path for the seccomp socket, so that branch is now dead code and the remainder is folded into its only caller, `setup_console_socket`.

Note one user-visible change: passing `--seccomp-notify-socket` now fails with glib's "unknown option" error rather than `seccomp support not present`. Both are hard failures, so no caller that works today starts silently misbehaving.

closes #591
closes #275


giuseppe (approved):
LGTM

kolyshkin:
ci failures seem unrelated (and are being fixed separately); merging

PR #672, opened 2026-08-15
ci: make CI green again

kolyshkin:
Assorted CI fixes to make CI green again:

1. Remove image-bundled `crun` and `runc`, ensuring the test is run with the ones we built.
2. Fix cri-tools version required by cri-o tests.
3. Fix bats version incompatibility with cri-o tests.

See individual commits for details.

### DIFF
--- a/hack/github-actions-setup
+++ b/hack/github-actions-setup
@@ -4,8 +4,9 @@ set -euo pipefail
 declare -A VERSIONS=(
     # renovate: datasource=github-releases depName=containernetworking/plugins
     ["cni-plugins"]=v1.9.1
-    # renovate: datasource=github-releases depName=bats-core/bats-core
-    ["bats"]=v1.14.0
+    # NOTE bats is deliberately not auto-updated: bats >= 1.14 errors out when
+    # a test suite ends up being empty, which breaks the cri-o test runner.
+    ["bats"]=v1.13.0
     # renovate: datasource=github-releases depName=opencontainers/runc
     ["runc"]=v1.5.1
     # renovate: datasource=github-releases depName=containers/crun
--- a/hack/github-actions-setup
+++ b/hack/github-actions-setup
@@ -17,6 +17,9 @@ main() {
     prepare_system
 
     install_packages
+    # NOTE this has to run after install_packages, which may install runtimes
+    # as podman dependencies.
+    remove_runtimes
     install_conmon
     install_bats
     install_critools
@@ -47,6 +50,15 @@ remove_packages() {
     sudo apt-get remove \
         conmon \
         containernetworking-plugins
+}
+
+# remove_runtimes removes all the pre-installed OCI runtimes, so they can't
+# shadow the ones we install. The runner image ships a static podman bundle in
+# /usr/local/bin (which comes first in PATH) containing its own runc and crun
+# -- notably a crun built without systemd support, which breaks all the cri-o
+# tests -- and docker brings its own /usr/bin/runc.
+remove_runtimes() {
+    sudo rm -f /usr/{local/,}{s,}bin/{runc,crun}
 }
 
 install_packages() {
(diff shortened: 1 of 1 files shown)


hack/github-actions-setup
```diff
@@ -31,6 +30,13 @@ prepare_system() {
     sudo systemctl stop docker
     sudo ufw disable
 
+    # Remove all pre-installed OCI runtimes, so they can't shadow the ones
+    # installed below. The runner image ships a static podman bundle in
+    # /usr/local/bin (which comes first in PATH) containing its own runc and
+    # crun -- notably a crun built without systemd support, which breaks all
+    # the cri-o tests -- and docker brings its own /usr/bin/runc.
+    sudo rm -f /usr/{local/bin,bin,sbin}/{runc,crun}
+
```
jnovy: This works today because `rm -f` deletes the binaries without uninstalling the apt packages, so dpkg's database still considers `crun`/`runc` installed and `apt install podman` (in `install_packages` below) sees its `crun | runc` dependency as satisfied.

It is a bit fragile though - if a future change to `install_packages` triggers a reinstall or upgrade that re-creates these files, `check_not_installed` would fire with an opaque failure. Moving the `rm -f` to run *after* `install_packages` (right before `install_runc`) would make the ordering self-evidently correct. Not blocking, just a thought.

hack/github-actions-setup
```diff
@@ -148,7 +161,19 @@ install_crun() {
     crun --version
 }
 
+# check_not_installed makes sure $1 is not available in PATH, so that the
+# binary we're about to install is not shadowed by a pre-installed one (those
+# are removed by prepare_system).
+check_not_installed() {
+    if command -v "$1"; then
+        echo "ERROR: $1 is already installed (see the path above), please remove it in prepare_system" >&2
```
jnovy: Nit: `command -v` prints the path to stdout before the error message goes to stderr. Under `set -x` the two can get interleaved. Slightly cleaner to capture the path:
```bash
check_not_installed() {
    local path
    if path=$(command -v "$1"); then
        echo "ERROR: $1 is already installed at $path, please remove it in prepare_system" >&2
        return 1
    fi
}
```
This makes the error self-contained even if stdout is redirected.

hack/github-actions-setup
```diff
@@ -126,17 +135,21 @@ install_cni_plugins() {
...
 
 install_runc() {
+    check_not_installed runc
+
     git clone --depth 1 --branch "${VERSIONS["runc"]}" https://github.com/opencontainers/runc
     pushd runc
     make RUNC_BUILDTAGS=-libpathrs
-    sudo install -D -m0755 runc /usr/sbin/runc
+    sudo install -D -m0755 runc /usr/bin/runc
     popd
```
jnovy, line 150: Good change - consistent with crun which also installs to `/usr/bin` via `make install prefix=/usr`.

hack/github-actions-setup
```diff
@@ -164,8 +189,19 @@ install_testdeps() {
     sudo cp test/policy.json /etc/containers
     sudo cp test/redhat_sigstore.yaml /etc/containers/registries.d/registry.access.redhat.com.yaml
     sudo cp test/registries.conf /etc/containers/registries.conf
+
+    # critest has to match the cri-o version we test against, otherwise it
+    # runs conformance tests for features cri-o does not implement yet, and
+    # those fail. Use the very same version cri-o itself is tested with.
+    critools=$(sed -n 's/^[[:space:]]*\["cri-tools"\]=\(.*\)$/\1/p' scripts/versions)
+    if [ -z "$critools" ]; then
```
jnovy, line 204: Verified the sed pattern matches cri-o's `scripts/versions` format (`    ["cri-tools"]=v1.36.0`). The empty-string guard below is good defensive coding.

jnovy (approved):
LGTM. Great root-cause analysis and clean fix. Two minor suggestions inline, neither blocking.

hack/github-actions-setup
kolyshkin: funny, this is how it was written and then I changed it to `! command -v` for simplicity.

hack/github-actions-setup
kolyshkin: Makes sense; done

hack/github-actions-setup
kolyshkin: changed

jnovy:
Filed https://github.com/cri-o/cri-o/pull/10244 to avoid downstream workarounds.

PR #673, opened 2026-08-19
test: fix some conmon test flakes/races

kolyshkin:
`attach: container finishes immediately without --stdin` is flaky: it waits for
the container to reach the `stopped` state and then reads the log, but the
container being stopped does not mean conmon is done writing the log, so the
test can read an empty one.

Seen in runc CI (the runc `validate / conmon` job runs this suite against a
runc build):

```
not ok 92 attach: container finishes immediately without --stdin
#   `assert "${output}" =~ "Container stopped!"  "'Container stopped!' found in the log"' failed
#|     FAIL: 'Container stopped!' found in the log
#| expected: =~ Container stopped\!
#|   actual:    ''
```

A rerun of the exact same commit passed.

conmon having exited does mean the log is complete, so this waits for that.

Waiting for a specific conmon needs its pid, and `$CONMON_PID_FILE` is a single
path per test that every conmon a test starts overwrites, so the first commit
gives each conmon started by `start_conmon_with_default_args` its own pidfile
and exposes the pid as `$CONMON_PID`. The rest builds on that: fix the flaky
test, then convert the remaining "wait for stopped, then read a log" places,
which have not been seen to fail but share the race.

Testing: the full suite passes locally with `-j $(nproc)`. Note that I could
not reproduce the race itself locally -- instrumenting the new helper showed
conmon had always already exited by the time it looked -- so this is verified
as "correct barrier, no regressions" rather than "observed to fix the flake".

The last commit fixes an unrelated but similar thing spotted on the way:
`04-runtime.bats` did `wait $CONMON_PID_FILE 2>/dev/null || true`, i.e. it
waited on a *filename* and swallowed the error, so it waited for nothing.

---

**Update**: rebased onto main, and the barrier is now in `run_conmon_with_default_args` itself (amended into "test: wait for conmon in the remaining log checks" rather than added as a separate commit).

That helper is the most common instance of the pattern -- 13 call sites across `02-ctr-logs.bats`, `04-runtime.bats` and `10-ctrl.bats`, every one of which reads a log right after -- so one barrier there covers all of them.

It also turns out this one is not merely theoretical: the same race fails the `coverage` job, where conmon is built with `--coverage` and is thus slow enough to lose it more often. Seen on an unrelated PR's run:

```
not ok 45 ctr logs: k8s partial message
#   `assert "${output}" =~ "stdout P"' failed
#| expected: =~ stdout P
#|   actual:    ''
```

Same signature as the `07-attach.bats` one: the runtime reports `stopped`, the log is read, and it is still empty. So this PR fixes two observed flakes, not one.

Re-tested after the rebase: `02-ctr-logs`, `04-runtime`, `07-attach`, `08-exec` and `10-ctrl` all pass locally.

---

**Update 2**: four more commits, all about CI not hanging for hours.

Today's runs kept getting stuck -- one job sat for 2.5 hours before being cancelled -- and it was never the same place twice: `hack/github-actions-setup` hung for 49 minutes on one run, the test step for 35 minutes on another. Both consist of network access with no timeout on it, and the GitHub Actions default job timeout is 6 hours, so nothing ever gave up.

5. `ci: set job timeouts` -- `timeout-minutes` on the jobs (conmon 20, cri-o 90, coverage 20, lint 10), a few times their normal runtime. A hang now fails in minutes instead of squatting a runner for the rest of the day.

6. `ci: time out and retry network access in the setup script` -- the script clones five repositories, downloads a tarball and a repo key, and runs apt. git has no timeout at all; curl has no total timeout and no retries; wget has a 900 second read timeout *and* 20 tries by default; apt retries nothing. All of them get a bound now, the clones via `GIT_HTTP_LOW_SPEED_LIMIT`/`GIT_HTTP_LOW_SPEED_TIME`.

7. `test: time out the image pull` -- `setup_suite` pulls the image the test rootfs comes from. It is the only network access in the suite, and a hang there stops everything before a single test runs, silently: bats shows `setup_suite` output only once it completes.

8. `test: bump the wait_for_runtime_status timeout` -- 5 seconds turned out to be too little on a loaded runner:

```
not ok 51 runtime: container execution with multiple log drivers
not ok 101 ctrl: resize the terminal, negative width and height
#| FAIL: timed out waiting for 'stopped' from conmon-test-...
```

The helper polls, so a longer limit costs nothing when things are healthy; 30 seconds now.

Commits 1-4 fix races (read a log before conmon wrote it); 5-8 stop CI hanging on stalled I/O. Happy to split the latter into their own PR if that reads better.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -12,6 +12,7 @@ jobs:
 
   coverage:
     runs-on: ubuntu-latest
+    timeout-minutes: 20
     steps:
       - uses: actions/checkout@v7
       - name: Install dependencies
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -12,6 +12,7 @@ jobs:
 
   conmon:
     runs-on: ubuntu-latest
+    timeout-minutes: 20
     steps:
       - uses: actions/checkout@v7
       - run: sudo hack/github-actions-setup
@@ -22,6 +23,7 @@ jobs:
 
   cri-o:
     runs-on: ubuntu-latest
+    timeout-minutes: 90
     strategy:
       matrix:
         go-version: [stable, oldstable]
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -13,6 +13,7 @@ jobs:
 
   lint:
     runs-on: ubuntu-latest
+    timeout-minutes: 10
     steps:
     - uses: actions/checkout@v7
     - name: Check C code formatting
--- a/hack/github-actions-setup
+++ b/hack/github-actions-setup
@@ -15,6 +15,11 @@ declare -A VERSIONS=(
 
 main() {
     set -x
+    # None of the git clones below has a timeout of its own, and a stalled
+    # one hangs this script until the job times out. Make git give up on a
+    # connection that transfers less than 1 KiB/s for a minute.
+    export GIT_HTTP_LOW_SPEED_LIMIT=1024 GIT_HTTP_LOW_SPEED_TIME=60
+
     prepare_system
 
     install_packages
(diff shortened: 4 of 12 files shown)


kolyshkin:
This is the order in which we need to merge test PRs:  #673 (this one, fixes conmon job) → #674 (fixes cri-o jobs) → #670 (fixes meson which was recently broken again).

kolyshkin:
OK this one is ready but I'll wait for review from @jnovy first

jnovy (approved):
LGTM - the core fix is sound, the refactoring is well-done, and the CI hardening is valuable.

Two minor suggestions:

1. **MEDIUM - Missing `assert_failure` in `test/02-ctr-logs.bats` (6 tests)**

Six tests switch from `start_conmon_with_default_args` to `run_conmon_expecting_failure` but none adds `assert_failure`. They only check `assert_output_contains`. Compare with `test/08-exec.bats` which correctly follows `run_conmon_expecting_failure` with `assert_failure`.

Risk: If conmon ever regresses to exit 0 while still printing the error message, these tests would pass even though the behavior is wrong. The function name `run_conmon_expecting_failure` implies the caller should verify the exit code. This is a pre-existing gap (the old code also didn't check), but the refactor makes the intent explicit - worth fixing while you're here.

2. **LOW-MEDIUM - `test/10-ctrl.bats` "rotate logs with --log-rotate" - missing `wait_for_conmon_exit` for main conmon**

The test starts main conmon via `start_conmon_with_default_args` (sets `CONMON_PID`), then runs exec conmon via `run_conmon_with_default_args` which overwrites `CONMON_PID` and waits for exec conmon to exit, then reads `$LOG_PATH` and `$LOG_PATH.1` (main conmon's output) without ever waiting for the main conmon.

The sibling test "rotate logs" and the `test_ctl_command` helper both correctly save `local main_conmon_pid=$CONMON_PID` before launching the second conmon, and call `wait_for_conmon_exit "$main_conmon_pid"`. This variant was missed and shares the exact same race the PR is fixing elsewhere.

Suggested fix:
```bash
# After start_conmon_with_default_args:
local main_conmon_pid=$CONMON_PID

# After run_conmon_with_default_args:
wait_for_conmon_exit "$main_conmon_pid"
```

kolyshkin:
Both good catches, thanks -- fixed, in the commits the lines came from rather than as fixups on top.

**1. Missing `assert_failure` in `test/02-ctr-logs.bats`.** Added right to `run_conmon_expecting_failure`.

**2. `ctrl: rotate logs with --log-rotate` missing the barrier for the main conmon.** Confirmed and fixed in `test: wait for conmon in the remaining log checks`, alongside the other barriers:

```bash
    wait_for_runtime_status "$CTR_ID" running
+   local main_conmon_pid=$CONMON_PID
...
    run_conmon_with_default_args \
        --log-path "k8s-file:$LOG_PATH.exec" \
        --exec \
        --exec-process-spec "${BUNDLE_PATH}/process.json"
+   # $LOG_PATH and $LOG_PATH.1, read below, are the main conmon's.
+   wait_for_conmon_exit "$main_conmon_pid"
```

`run_conmon_with_default_args` waits for the conmon it starts, i.e. the exec one, while `$LOG_PATH` and `$LOG_PATH.1` are written by the first -- exactly the race this series is about, and this variant was indeed missed. The commit message now calls it out specifically, since it is the one place where the barrier is not for the conmon just started.

Full suite green locally.

kolyshkin:
One day we'll have green CI and no flakes here. Probably not today though.

To merge (on top of this one)

1. Fixes:
 - #674 (to fix some cri-o failures)
 - #679 (to fix some podman failures that result in conmon job failures)

2. Improvements:
 - #670 (fix meson, add meson CI job)
 - #678 (install decent Go to avoid go downloading another go, and remove cri-o oldstable from matrix)
 - #676 (mostly cosmetic: log versions, grouping in CI logs, canceling old CI runs)

kolyshkin:
conmon still fails here, this is being fixed in #679. Everything else is fixed, so merging.

PR #676, opened 2026-08-19
ci, test: CI log readability, superseded runs, sync pipe race

kolyshkin:
Three independent readability fixes for CI logs, plus a fix for a flaky test the first CI run of this PR ran into.

### 1. `test: print the versions the tests run with`

Which versions a run gets varies a lot -- across the Ubuntu 22.04, 24.04 and 26.04 runner images, across architectures, and between this repo's CI and runc's, which runs this suite as well. A failure report that does not say which conmon, runtime, podman and kernel were involved is worth very little.

`test/run-tests.sh` already printed the conmon and runtime *paths*; this turns that into a proper block:

```
[INFO] Running tests with:
[INFO]   kernel:  Linux 6.11.0-1018-azure x86_64
[INFO]   distro:  Ubuntu 24.04.3 LTS
[INFO]   conmon:  /home/runner/work/conmon/conmon/bin/conmon: conmon version 2.2.1 commit: ...
[INFO]   runtime: /usr/bin/runc: runc version 1.5.1
[INFO]   bats:    /usr/local/bin/bats: Bats 1.13.0
[INFO]   podman:  /usr/local/bin/podman: podman version 5.8.4
[INFO]   podman:  /usr/bin/podman: podman version 4.9.3
```

podman gets listed in full rather than just resolved, because a runner may well have more than one: the runner image ships podman 5.x under `/usr/local`, while `apt install podman` adds Ubuntu's 4.x to `/usr/bin`. First in PATH wins, and it is worth being able to see which that was -- podman 4.x vs 5.x has already cost us once, in 0af8e86 ("test: don't use podman pull --policy").

This benefits runc's CI as much as ours, since it runs this very suite.

### 2. `ci: group the setup log`

`hack/github-actions-setup` runs under `set -x` and installs a dozen things; its log is a few thousand lines of apt, git and make output in which nothing can be found. Each step now goes into a collapsible [log group](https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#grouping-log-lines):

```
::group::install_crun
+ install_crun
...
::endgroup::
```

so the step that failed -- or the step that hangs, which is what prompted this -- is one click away instead of a scroll hunt. Outside GitHub Actions the markers are two harmless extra lines per step.

Verified locally: the version block renders as above, and the `step` wrapper keeps `set -x` inside the group while leaving the markers themselves untraced (they have to be plain lines for GitHub to fold them).

### 3. `ci: cancel superseded PR runs`

Pushing to a PR leaves its previous runs going. A branch pushed to a few times in a row ends up with several full runs of the same workflow competing for runners, all but the last of which nobody will look at -- and the integration workflow takes about 40 minutes.

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
  cancel-in-progress: true
```

`github.head_ref` is set for `pull_request` events only, so pushes to main and to tags fall back to the unique `run_id`, each landing in a group of its own and never being cancelled. `static.yml` is left alone: it runs on tags and `workflow_dispatch` only, where there is nothing to supersede.

### 4. `test: wait for the sync pipe output`

The first CI run of this PR hit

```
not ok 99 exec: --exec with _OCI_SYNCPIPE defined
...
FAIL: File /tmp/conmon-test-HAmjK4/syncpipe-output does not exist.
```

in the coverage job. The reader of the sync pipe runs in the background, appending to that file as the lines arrive, so it lags behind the conmon that wrote them -- and the tests checked for the file as soon as the container was reported stopped. The coverage job, the slowest of them all, loses that race often enough to matter.

The four affected tests now wait for the expected number of lines to show up instead. That also covers the `--api-version=1` test, which wants two of them (the pid and the exit status) and could previously have found the file with only the first one in it.

Unrelated to the rest of the PR, but it is what the PR's own CI failed on.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -8,6 +8,13 @@ on:
   pull_request:
 permissions: {}
 
+# Cancel a PR's older runs when it is pushed to again. github.head_ref is set
+# for pull_request events only, so pushes to main and tags fall back to the
+# unique run_id and are never cancelled.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
 jobs:
 
   coverage:
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -8,6 +8,13 @@ on:
   pull_request:
 permissions: {}
 
+# Cancel a PR's older runs when it is pushed to again. github.head_ref is set
+# for pull_request events only, so pushes to main and tags fall back to the
+# unique run_id and are never cancelled.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
 jobs:
 
   conmon:
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -9,6 +9,13 @@ on:
   pull_request:
 permissions: {}
 
+# Cancel a PR's older runs when it is pushed to again. github.head_ref is set
+# for pull_request events only, so pushes to main and tags fall back to the
+# unique run_id and are never cancelled.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
 jobs:
 
   lint:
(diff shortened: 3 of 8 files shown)


jnovy (approved):
LGTM. Clean, well-motivated CI improvement - no bugs, no security concerns, good shell scripting practices throughout. Commit structure is logical with proper separation of concerns. Ready to merge.

kolyshkin:
first all-green PR for quite some time; merging

PR #677, opened 2026-08-19
[test/DNM] ci: try to reproduce the conmon job hang

kolyshkin:
**Do not merge.** This is #673 plus one throwaway commit on top, to hunt a hang.

### What we are chasing

Twice today the `conmon` job went quiet mid-suite and stayed that way until it was killed:

```
03:48:35  ok 67 exec requires proper arguments
          <17 minutes of nothing>
04:05:24  The job has exceeded the maximum execution time of 20m0s
```

Test 68 is `integration: exec exit codes work correctly` in `test/06-exec-exit-status.bats`. It runs six podman commands, and until #673 exactly one of them -- the initial probe -- had a timeout. The prime suspect is the container creation:

```bash
container_id=$(podman --conmon $conmon_path run -dt "$UBI10_MICRO_IMAGE" sleep 30)
```

A command substitution waits for stdout to be closed, and conmon is started by podman with that very stdout. This test deliberately runs the *freshly built* conmon, so it is exactly the thing that might hold it open.

Nothing is printed while it hangs, because bats shows a test's output only once the test ends -- so from the outside the suite simply goes silent.

### What this branch does

* runs the `conmon` job **ten times in parallel** (`matrix.attempt`, `fail-fast: false`), since it does not reproduce locally at all and in CI hits maybe one run in several;
* drops the `cri-o` job and switches `coverage` and `validate` to `workflow_dispatch`, so nothing else competes for runners;
* keeps `timeout-minutes: 20` from #673, so a wedged attempt dies in 20 minutes rather than 6 hours;
* dumps the surviving `conmon`/`podman`/`runc`/`crun` processes and `podman ps -a` if the container creation times out.

### Reading the results

Both outcomes are useful:

* `not ok 68 integration: exec exit codes work correctly` in some attempts -> the suspect is confirmed, and the process dump says who was still alive;
* an attempt wedging for the full 20 minutes anyway -> the hang is somewhere else, and the log says after which test.
* all ten green -> it is rarer than one in ten, and the honest answer is that #673's timeouts contain it without explaining it.


kolyshkin:
Enough statistics collected: five runs of the 10x2 matrix, ~14 hangs of the podman cleanup deadlock with the bundled podman against 0 without it, and the exec pidfile race identified and fixed in #673 (0 occurrences in the last 20 jobs).

The findings live in #673 (test fixes and CI containment) and #679 (removing the bundled podman); the deadlock itself is written up for whoever takes it upstream. The branch stays around in case the experiment needs re-running.

PR #680, opened 2026-08-19
exec: report a runtime failure to the parent

kolyshkin:
When the runtime fails to start the exec'd process, it does not write the pid file. Conmon merely warned about that and exited, and since the failing conmon is the daemonized child, its exit status went nowhere: the caller saw the parent exit 0 and got nothing on the sync pipe, so an exec that never ran looked exactly like a successful one.

A non-zero runtime exit status can not be used to tell those apart for exec -- it is also how the exit status of the exec'd command is reported, which is why the existing check skips exec. A missing pid file, on the other hand, means the process was never started.

So report it: write the failure to the sync pipe, together with whatever the runtime printed to its stderr, and exit with an error rather than a warning.

With this, a failing exec produces

```
{"data": -1, "message": "... exec failed: unable to start container process: exec: \"/no/such/binary\": stat /no/such/binary: no such file or directory\n"}
```

instead of silence. A test for that is added.

### DIFF
--- a/src/conmon.c
+++ b/src/conmon.c
@@ -390,8 +390,27 @@ int main(int argc, char *argv[])
 	/* Read the pid so we can wait for the process to exit */
 	_cleanup_free_ char *contents = NULL;
 	if (!g_file_get_contents(opt_container_pid_file, &contents, NULL, &err)) {
-		nwarnf("Failed to read pidfile: %s", err->message);
-		exit(1);
+		/*
+		 * The runtime did not write the pid file, so the process was never started.
+		 * For exec this is the only way to tell "the runtime failed" from "the command
+		 * exited with a non-zero status", as the latter is a legitimate non-zero exit
+		 * of the runtime, checked (and skipped for exec) above. Report the failure to
+		 * the parent, which would otherwise see a successful exec, and pass along
+		 * whatever the runtime has to say about it.
+		 */
+		const char *error_msg = err->message;
+		if (mainfd_stderr >= 0) {
+			g_unix_set_fd_nonblocking(mainfd_stderr, TRUE, NULL);
+			num_read = read(mainfd_stderr, buf, BUF_SIZE - 1);
+			if (num_read > 0) {
+				buf[num_read] = '\0';
+				nwarnf("runtime stderr: %s", buf);
+				error_msg = buf;
+			}
+		}
+		if (sync_pipe_fd >= 0)
+			write_or_close_sync_fd(&sync_pipe_fd, -1, error_msg);
+		nexitf("Failed to read pidfile: %s", err->message);
 	}
 
 	container_pid = atoi(contents);
--- a/test/08-exec.bats
+++ b/test/08-exec.bats
@@ -186,3 +186,27 @@ teardown() {
     assert_json "${output}" =~ '"data": 0'
 }
 
+
+@test "exec: runtime failure is reported to the sync pipe" {
+    start_conmon_with_default_args --log-path "k8s-file:$LOG_PATH"
+    wait_for_runtime_status "$CTR_ID" running
+    start_oci_sync_pipe_reader
+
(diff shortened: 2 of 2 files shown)


src/conmon.c
```diff
@@ -390,8 +390,27 @@ int main(int argc, char *argv[])
...
+			num_read = read(mainfd_stderr, buf, BUF_SIZE - 1);
+			if (num_read > 0) {
+				buf[num_read] = '\0';
+				nwarnf("runtime stderr: %s", buf);
+				error_msg = buf;
+			}
+		}
+		if (sync_pipe_fd >= 0)
+			write_or_close_sync_fd(&sync_pipe_fd, -1, error_msg);
+		nexitf("Failed to read pidfile: %s", err->message);
```
TomSweeneyRedHat, line 413: I'm not 100% sure on this one, maybe this instead?

```suggestion
		nexitf("Failed to read pidfile: %s", error_msg);
```

TomSweeneyRedHat:
One question, otherwise LGTM

jnovy (approved):
LGTM. Clean, well-motivated fix for a real silent-failure bug.

**The problem**: when the runtime fails to start an exec'd process (e.g., binary does not exist), it never writes the pid file. The old code just warned and called `exit(1)`, but since this runs in the daemonized child, the exit status is lost - the parent saw a successful exit and got nothing on the sync pipe, making a failed exec indistinguishable from a successful one.

**The fix**: detect the missing pidfile, read whatever the runtime wrote to stderr, report it via the sync pipe as `{"data": -1, "message": "..."}`, and exit with an error. This is the right detection signal for exec because (as the comment explains) non-zero runtime exit status is ambiguous for exec - it also conveys the exec'd command's own exit code.

Code review notes:

- **Nonblocking read**: `g_unix_set_fd_nonblocking(mainfd_stderr, ...)` before the read is correct defensive coding. Notably, the existing stderr-reading block above (line 366 on main) does *not* set nonblocking, which could theoretically block - this new code is actually better in that regard.

- **`sync_pipe_fd >= 0`**: correct, since fd 0 is a valid file descriptor. This is consistent with line 407 (`opt_api_version >= 1` path). The pre-existing checks at lines 374/385 use `> 0`, which is a minor inconsistency in the existing code (not introduced by this PR).

- **Error message fallback**: `error_msg` defaults to `err->message` (the pidfile-read error) and is overridden only if runtime stderr has content - good fallback behavior so the sync pipe always carries something useful.

- **No resource issues**: `buf` is stack-allocated (`BUF_SIZE = 8192`), null-terminated after `read()`, and `nexitf()` terminates the process so there are no dangling state concerns.

- **`write_or_close_sync_fd` safety**: the function internally checks `*fd == -1` and returns early, so the `>= 0` guard plus the internal check makes this double-safe.

- **Test**: well-structured, follows the existing patterns in `08-exec.bats`. The `/no/such/binary` approach to trigger a runtime failure is clean, and asserting both `"data": -1` and `"exec failed"` validates both the structure and the content of the error report.

jnovy (approved):
LGTM - clean, well-motivated fix for a real silent-failure bug. The error propagation via sync pipe is correct, nonblocking read is properly set up, and the test covers the failure path well.

src/conmon.c
kolyshkin: Nah, those are two different errors:
 - `err->message` is the original error from Glib (something like "Failed to open file “/tmp/conmon-test-XXX/pidfile.2”: No such file or directory"), i.e. the conmon error.
 - `error_msg` is redefined inside `if` to be the actual runtime's stderr (something like "exec failed: unable to start container process: ...")
 and we want to see both the way we reveal those.

TomSweeneyRedHat: Oops!  Missed the read into the num_read buf, thanks for the explanation.

PR #681, opened 2026-08-19
Do not free the string owned by GLib

kolyshkin:
`g_get_tmp_dir()` returns a constant owned by GLib and cached in a static variable, which the caller must not free -- but `setup_console_socket` declares it `_cleanup_free_`, so it is freed on return. That leaves GLib's static pointing to freed memory, and the next `g_get_tmp_dir()` in the process hands out a dangling pointer.

It does not bite today, since `setup_console_socket` is called at most once per conmon process, but a use-after-free is a use-after-free.

### DIFF
--- a/src/conn_sock.c
+++ b/src/conn_sock.c
@@ -149,7 +149,7 @@ static void set_socket_buffers(int fd)
 
 char *setup_console_socket(void)
 {
-	_cleanup_free_ const char *tmpdir = g_get_tmp_dir();
+	const char *tmpdir = g_get_tmp_dir();
 
 	char *csname = g_build_filename(tmpdir, "conmon-term.XXXXXX", NULL);
 	/*


jnovy (approved):
LGTM. Correct fix for a use-after-free bug.

`g_get_tmp_dir()` returns a pointer to GLib's internal static cache - the caller does not own it. The `_cleanup_free_` attribute (which expands to `_cleanup_(freep)` via utils.h) calls `free()` on scope exit, corrupting GLib's internal state so that any subsequent `g_get_tmp_dir()` call returns a dangling pointer.

The fix is straightforward: drop the cleanup attribute since the pointer is borrowed, not owned. The only use of `tmpdir` is as input to `g_build_filename()`, which copies its arguments, so there is no lifetime concern.

As the commit message notes, this does not bite in practice today since `setup_console_socket` is called at most once per conmon process, but it is still the right thing to fix - a sanitizer build or a future refactor that calls the function path twice would turn this into a real crash.

PR #685, opened 2026-08-24
Moar test fixes

kolyshkin:
The coverage job hit this on #684, which touches nothing but the cri-o test selection:

```
not ok 120 ctrl: rotate logs with --log-rotate
  in test file test/10-ctrl.bats, line 192
  `assert "${output}" =~ "before rotation"' failed
```

### The race

`ctrl: rotate logs with --log-rotate` starts a container that writes `before rotation`, sends `2 1 1` to the control fifo to rotate the log, and then checks that the line is in the rotated file `$LOG_PATH.1` and that what comes after is in the new `$LOG_PATH`.

The test waits for the container to be *running* before rotating, but that says nothing about conmon having written the first line to the log yet. Lose that race and the rotation moves a log that does not contain the line, `before rotation` lands in the fresh log along with `after rotation`, and `$LOG_PATH.1` is there -- so the `assert_file_exists` above passes -- but empty. The coverage job is the slowest one, which is where this turns up.

So the test now waits for the line to be logged before rotating.

### The helper

While adding the fourth "poll until X" helper to `test_helper.bash`, the loop -- deadline from `$SECONDS`, check, sleep, `die` -- was written out three times already. It is now

```bash
retry <how_long> <interval> <command ...>
```

which returns non-zero instead of dying: it does not know what it was waiting for, the callers do, and they keep their own messages. `wait_for_runtime_status`, `wait_for_conmon_exit` and `wait_for_syncpipe_output` are converted to it, and the new `wait_for_log_line` uses it too.

No functional change from the conversion, other than the status check in `wait_for_runtime_status` no longer printing `expr`'s match count into the test log.

Ran the full suite locally: 120/120.

### The unrelated CI breakage

Also here, since the cri-o job was red for a reason of its own: go 1.27 (which `go-version: stable` started resolving to) does not build cri-o main.

In `golang.org/x/net` v0.54.0, which cri-o vendors, `http2` switches to a thin wrapper around `net/http` under go 1.27 (`server.go` is `//go:build !(go1.27 && !http2legacy)`), and `http2.TrailerPrefix` -- used by the vendored grpc -- is only defined in the non-wrapped build:

```
vendor/google.golang.org/grpc/internal/transport/handler_server.go:271:18: undefined: http2.TrailerPrefix
```

Upstream fixes this by vendoring x/net v0.55.0, which moves `TrailerPrefix` into a file built either way; see https://github.com/cri-o/cri-o/pull/10103#issuecomment-5400108050. Until that lands, go is pinned to 1.26 here.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -24,7 +24,10 @@ jobs:
       - uses: actions/checkout@v7
       - uses: actions/setup-go@v7
         with:
-          go-version: stable
+          # TODO: switch back to stable once cri-o vendors golang.org/x/net
+          # v0.55.0 or newer; until then it does not build with go 1.27.
+          # See https://github.com/cri-o/cri-o/pull/10103#issuecomment-5400108050
+          go-version: '1.26'
           cache: false
       - run: hack/github-actions-setup
       - name: Run conmon integration tests
@@ -45,7 +48,10 @@ jobs:
       - uses: actions/checkout@v7
       - uses: actions/setup-go@v7
         with:
-          go-version: stable
+          # TODO: switch back to stable once cri-o vendors golang.org/x/net
+          # v0.55.0 or newer; until then it does not build with go 1.27.
+          # See https://github.com/cri-o/cri-o/pull/10103#issuecomment-5400108050
+          go-version: '1.26'
           cache: false
       - run: hack/github-actions-setup
       - name: Run CRI-O integration tests (critest=${{ matrix.critest }})
--- a/test/10-ctrl.bats
+++ b/test/10-ctrl.bats
@@ -173,6 +173,11 @@ test_resize_command_ok() {
     wait_for_runtime_status "$CTR_ID" running
     local main_conmon_pid=$CONMON_PID
 
+    # The container being up does not mean its first line has been logged
+    # yet, and rotating before it is written leaves it in the new log rather
+    # than the rotated one, which is what the test checks for below.
+    wait_for_log_line "$LOG_PATH" "before rotation"
+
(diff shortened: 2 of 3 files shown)


jnovy (approved):
LGTM. Three well-scoped commits, each cleanly separated by concern.

**Commit 1 - retry helper.** The refactoring is faithful to the three original poll loops. The generic `retry <how_long> <interval> <command ...>` pattern is a nice de-duplication - returning non-zero instead of dying is the right call, since the callers know what they were waiting for and can produce a meaningful message. The helper predicates (`_runtime_status_is`, `_pid_is_gone`, `_file_has_lines`) each capture exactly the condition the original inline code tested. Bonus: `expr` match count no longer leaks into the test log from `_runtime_status_is` (redirected to `/dev/null`), and the `echo "$output"` is preserved so `bats` still captures runtime state output for debugging.

**Commit 2 - log rotate race fix.** The race analysis is spot-on: `wait_for_runtime_status ... running` says nothing about whether conmon has written the first line to the log yet. Rotating before that line lands puts it in the new log instead of the rotated one, which is exactly the failure mode the coverage job hit. `wait_for_log_line "$LOG_PATH" "before rotation"` closes the gap cleanly. The new `_file_has_line` helper using `grep -q --` is defensive against patterns that start with `-` - good practice.

**Commit 3 - pin go to 1.26.** Reasonable workaround for the `http2.TrailerPrefix` build breakage with go 1.27. The TODO comment with the upstream cri-o PR link makes the reversion path clear.

Shell quoting is correct throughout, all commits are signed off, and the commit messages are thorough with good root-cause explanations. No bugs, no security concerns.

PR #686, opened 2026-08-25
ci: add actionlint workflow

kolyshkin:
Lint the GitHub Actions workflow files, similar to what [runc does](https://github.com/opencontainers/runc/blob/main/.github/workflows/actionlint.yml).

This is deliberately a separate workflow, so that it still runs when another workflow file is broken.

Adapted to conmon conventions: `permissions: {}`, `ubuntu-latest` with a timeout, the same concurrency group as the other workflows, and an `all-done` job.

Checked locally with actionlint v1.7.12 (including `-shellcheck`): all current workflows pass, so this starts out green.


jnovy:
LGTM

PR #687, opened 2026-08-25
ci: fix zizmor findings, add zizmor workflow

kolyshkin:
Based on #686 (the first commit here is that PR); please merge that one first.

Run [zizmor](https://docs.zizmor.sh/) over the workflow files, fix everything it reports, and add a CI job so it stays fixed.

The findings, one commit each:

* **artipacked** (7): `actions/checkout` leaves the GitHub token in `.git/config` by default. None of the workflows use those credentials, so set `persist-credentials: false`.
* **unpinned-uses** (10): actions were referenced by mutable tags. Pin them to commit hashes, with the version in a trailing comment. Renovate keeps these updated.

The `zizmor` workflow is a separate file, for the same reason as `actionlint`: it still runs when another workflow file is broken. `advanced-security: false` is set so that findings fail the job instead of only appearing in the security tab.

Two findings are only reported by `--persona=pedantic` and are left alone:

* `template-injection` on `integration.yml` for `${{ matrix.critest }}` — the matrix is defined statically in the same file, so it is not attacker-controllable.
* `concurrency-limits` on `static.yml` — it only runs on tags and `workflow_dispatch`.

Verified locally with zizmor 1.22.0 (`--persona=regular`): no findings. `actionlint` is also clean.

### DIFF
--- a/.github/workflows/actionlint.yml
+++ b/.github/workflows/actionlint.yml
@@ -1,0 +1,36 @@
+# This is deliberately a separate workflow -- so that it still runs when
+# another workflow file is broken.
+name: actionlint
+on:
+  push:
+    tags:
+      - v*
+    branches:
+      - main
+      - release-*
+  pull_request:
+  workflow_dispatch:
+permissions: {}
+
+# Cancel a PR's older runs when it is pushed to again. github.head_ref is set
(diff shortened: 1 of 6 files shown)


jnovy:
LGTM

PR #690, opened 2026-08-26
ci: add static analysis and more compiler warnings

kolyshkin:
Until now the only check on the C code itself was clang-format. This adds
two static analyzers and a larger set of compiler warnings, plus the small
fixes needed to make the tree clean under them.

### Static analyzers

Two new make targets, so they can be run locally exactly the way CI runs
them, and two new blocking jobs in `validate.yml`:

- `make analyze` runs `clang --analyze` over every source file;
- `make cppcheck` runs cppcheck with the warning, performance and
  portability checks.

Three clang checkers are disabled and cppcheck's `style` checks are not
enabled; the Makefile says why. The short version is that the clang static
analyzer does not model `__attribute__((cleanup))`, which this code base
uses throughout, so every `_cleanup_free_` / `_cleanup_close_` /
`_cleanup_fclose_` variable looks either leaked or assigned but never read.
The CFG does carry a `CleanupFunction` element, but the engine never
evaluates it, and `ownership_takes` does not help, since the attribute
would refer to the address of the variable rather than to the pointer.

`gcc -fanalyzer` is deliberately not part of this. It does model the
cleanup attribute, but it does not know that `execv(2)` does not return,
which costs six false fd-leak reports around the fork/exec path.

### Warning flags

Everything added here is already clean on the tree, so it only guards
against new code. Flags that only some compilers have go through a
`cc-option` helper that probes the compiler first, because clang rejects
an unknown `-W` option outright. `-Wformat-signedness` and
`-Wjump-misses-init` exist in gcc and in clang 19 and later, but not in
clang 18, which is what ubuntu-latest still ships.

Three flags were considered and left out, each for a stated reason:
`-Wcast-align` (fires on the `CMSG_DATA()` casts, where the kernel
guarantees the alignment), `-Wcast-qual` (three of five hits cannot be
fixed: volatile hash table keys written by a signal handler, and the
non-const `iov_base` of `writev(2)`), and `-Wswitch-enum` (its only hit
already has a default case).

### Code changes

All of them are fixes for the above, and none is meant to change
behaviour:

- `socket_parent_dir()` used the only plain `strdup()` left in the tree;
- `disconnect_std_streams()` took two parameters that shadowed the globals
  of the same name, which both call sites passed in anyway;
- an outer `ret` in `main()` shadowed a genuinely local one further down;
- the inotify event buffer in `oom_cb_cgroup_v2()` was a variable-length
  array, because a `const size_t` is not an integer constant expression;
- arithmetic on the `void *` `iov_base` is a GNU extension;
- `remote_sock_shutdown()` switches over an `int`, so it wants a default;
- `st_uid` is unsigned and was printed with `%d`;
- a good deal of missing `const` on strings that are only ever read.

### Testing

`make`, `make analyze`, `make cppcheck` and the meson build were all run
against gcc 13 with clang 18 and gcc 16 with clang 22. The test suite was
not run locally, so CI is its first pass.


jnovy:
LGTM

PR #691, opened 2026-08-26
ci: add shfmt and shellcheck for shell scripts

kolyshkin:
Adds shfmt and shellcheck to CI, plus the fixes needed to make the tree clean under both.

Both run the way runc runs shfmt: a make target using a pinned container image, so a local run and a CI run cannot disagree, plus `localshfmt` / `localshellcheck` targets for whoever already has the tools installed. The CI jobs are then just `make shfmt` and `make shellcheck`. Pinning matters more for shellcheck than for shfmt, since what it reports moves between releases and ubuntu-latest is several versions behind.

### hack/get_ci_vm.sh

Removed first, because it is dead. It sets up a Cirrus-CI VM in GCP, and conmon stopped using Cirrus in e0c56ec; its `--setup` path sources `./contrib/cirrus/lib.sh` and runs `./contrib/cirrus/setup_environment.sh`, both deleted in 42cecdf. shellcheck is what pointed this out, via the SC1091 on that `source` line.

### shfmt

The `.editorconfig` is podman's file minus `space_redirects`, so redirections keep the shape they already have here. The `[[shell]]` section is what lets shfmt pick up `hack/github-actions-setup`, a shell script with a shebang and no extension; that section is an shfmt extension rather than something editorconfig itself understands. See podman-container-tools/podman#28785.

shfmt finds the files to format on its own, `.bats` files included, so no file list has to be maintained.

### shellcheck

One commit per theme: quoting (SC2086), declare-and-assign (SC2155), the printf padding trick (SC2140, SC2183), moving a helper above its callers (SC2218), and directives for the places shellcheck cannot know the context (SC2016, SC1091, SC2034, SC2154). No check is disabled globally.

None of the findings was an actual bug. The one that needed care is `get_conmon_journal_output()`, where the word splitting was load-bearing: its level filter is either empty or `-p <level>`, which has to reach journalctl as two arguments, so it becomes an array rather than a quoted string.

### Note

`test/Makefile` was reformatted in passing before this branch and lost the tabs in its `clean` recipe, which broke `make -C test clean` with `missing separator`. That is reverted here. It may be worth an `.editorconfig` section pinning Makefiles to tabs so it cannot happen again.


jnovy (approved):
LGTM

.editorconfig
```diff
@@ -0,0 +1,20 @@
...
+[**/vendor/**]
+ignore = true
+
+# Bash, bats, and sh files use 4 spaces for indentation.
+[*.{bash,bats,sh}]
+indent_style = space
+indent_size = 4
+
+# This is for shell scripts with shebang but no extensions.
+# Not yes supported by editorconfig[1], only by shfmt.
```
jnovy: nit: "Not yes" -> "Not yet"

kolyshkin: fixed

kolyshkin:
rebased (conflicts in Makefile resolved)

PR #692, opened 2026-08-26
ci: attach the static binaries to the release

kolyshkin:
Pushing a `v*` tag already builds the static binaries for all five architectures, but they are left as workflow artifacts: cutting a release means downloading five of them by hand and uploading them again, which is how `conmon.amd64` & co. ended up on v2.2.1. Do it in the workflow instead.

If no release exists for the tag yet — a tag pushed on its own, rather than one created along with a release from the web UI — the job creates a draft, so that the release notes can be written before anything is announced. Publishing the release is what triggers packit's downstream sync to Fedora and CentOS Stream, so it matters that the workflow does not publish anything by itself.

### Testing

The `upload` job only runs on a `v*` tag, so a PR run does not exercise it at all. It was tested by pushing a scratch tag to my fork, with the cachix push dropped for that run since a fork has no `CACHIX_AUTH_TOKEN` (reading from the cache still works). The tag, the branch and the draft release have since been deleted.

The draft was created, the five assets were attached under the same names v2.2.1 uses, and the downloaded `conmon.amd64` is a stripped, statically linked ELF that runs and reports its version.

### DIFF
--- a/.github/workflows/static.yml
+++ b/.github/workflows/static.yml
@@ -54,6 +54,45 @@ jobs:
           name: conmon-${{ matrix.name }}
           path: bin/conmon
 
+  upload:
+    # Attach the binaries built above to the release for this tag, so that
+    # cutting a release does not mean downloading five artifacts by hand.
+    # Manual runs have no tag to attach to, hence the condition.
+    if: startsWith(github.ref, 'refs/tags/v')
+    needs:
+      - static-binary
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


jnovy (approved):
LGTM. Clean, well-motivated CI improvement. Good security posture (scoped permissions, pinned actions), correct handling of the merge-gate job dependency, thoughtful draft-vs-published distinction for packit, and idempotent re-run behavior.

PR #695, opened 2026-08-28
ci: speed up the integration setup step

kolyshkin:
Two changes:

- **Shallow-clone cri-o and bats.** The cri-o clone is the only one in the script without a depth limit, and it is the largest repository involved. Only the default branch's tree is used (`scripts/versions`, the test suites, and the vendored ginkgo), so the history is fetched and discarded every run. bats is switched from clone-then-checkout to fetching the tag directly.

- **Cache GOCACHE and the module cache.** The script builds runc, ginkgo, critest and crictl from source, and the cri-o job builds cri-o itself; every run recompiles the standard library and the vendored trees from scratch. `setup-go`'s own cache is keyed on a `go.sum` this repository does not have, hence caching the two directories by hand. The key follows the versions pinned in the setup script, with `restore-keys` so that a version bump still gets an incremental first build, and includes the job name so the two jobs do not race for one cache entry.

  `.github/zizmor.yml` comes with it: zizmor's `cache-poisoning` audit fires because a tag push can trigger this workflow, and a poisoned cache could then reach a published artifact. This workflow publishes nothing -- the release build is in `static.yml`, which does not cache -- so the finding is ignored for this file only.

A third commit, oversubscribing the cores for the cri-o suite, has been dropped: cri-o/cri-o#10285 makes `test/test_runner.sh` default `JOBS` to `nproc * 2` itself.

Detailed measurements are in the comments below.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -31,6 +31,22 @@ jobs:
           # See https://github.com/cri-o/cri-o/pull/10103#issuecomment-5400108050
           go-version: '1.26'
           cache: false
+      # The setup script builds runc, ginkgo, critest and crictl from source,
+      # and the cri-o job builds cri-o itself; all of that is Go, and all of it
+      # recompiles the standard library and the vendored trees from scratch on
+      # every run. setup-go's own cache is keyed on a go.sum this repository
+      # does not have, hence caching GOCACHE and the module cache by hand.
+      #
+      # The key follows the dependency versions pinned in the setup script, so
+      # a bump gets a fresh cache; restore-keys then still seeds it from the
+      # previous one, which keeps that first build incremental.
+      - uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
+        with:
+          path: |
+            ~/.cache/go-build
+            ~/go/pkg/mod
+          key: ${{ runner.os }}-go-${{ github.job }}-${{ hashFiles('hack/github-actions-setup') }}
+          restore-keys: ${{ runner.os }}-go-${{ github.job }}-
       - run: hack/github-actions-setup
       - name: Run conmon integration tests
         run: |
@@ -57,6 +73,22 @@ jobs:
           # See https://github.com/cri-o/cri-o/pull/10103#issuecomment-5400108050
           go-version: '1.26'
           cache: false
+      # The setup script builds runc, ginkgo, critest and crictl from source,
+      # and the cri-o job builds cri-o itself; all of that is Go, and all of it
+      # recompiles the standard library and the vendored trees from scratch on
+      # every run. setup-go's own cache is keyed on a go.sum this repository
+      # does not have, hence caching GOCACHE and the module cache by hand.
+      #
+      # The key follows the dependency versions pinned in the setup script, so
+      # a bump gets a fresh cache; restore-keys then still seeds it from the
+      # previous one, which keeps that first build incremental.
+      - uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
+        with:
+          path: |
+            ~/.cache/go-build
+            ~/go/pkg/mod
+          key: ${{ runner.os }}-go-${{ github.job }}-${{ hashFiles('hack/github-actions-setup') }}
+          restore-keys: ${{ runner.os }}-go-${{ github.job }}-
       - run: hack/github-actions-setup
       - name: Run CRI-O integration tests (critest=${{ matrix.critest }})
         run: |
--- a/.github/zizmor.yml
+++ b/.github/zizmor.yml
@@ -1,0 +1,8 @@
+# cache-poisoning fires on integration.yml because a tag push can trigger it,
+# and a poisoned cache could then end up inside a published artifact. That
+# workflow only runs tests and publishes nothing -- the release build lives in
+# static.yml, which does not cache -- so the finding does not apply here.
+rules:
+  cache-poisoning:
+    ignore:
+      - integration.yml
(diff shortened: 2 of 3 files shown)


kolyshkin:
### Measurements

Re-measured after dropping the `JOBS` commit, so the numbers below are the effect of **this PR plus cri-o/cri-o#10285**, which now defaults `JOBS` to `nproc * 2` in cri-o's own `test/test_runner.sh`.

Baseline is the three most recent `integration` runs on `main` (9 jobs): [33389659768](https://github.com/containers/conmon/actions/runs/33389659768), [33207102160](https://github.com/containers/conmon/actions/runs/33207102160), [33016232632](https://github.com/containers/conmon/actions/runs/33016232632). The PR figures are run [34317505484](https://github.com/containers/conmon/actions/runs/34317505484), attempt 1 on a cold cache and attempt 2 on a confirmed hit. Medians, in seconds.

**`hack/github-actions-setup`, by function:**

| | `main` (n=9) | PR, cold cache (n=3) | PR, cache hit (n=3) |
|---|---|---|---|
| cri-o clone | 11 | 4 | 4 |
| `install_packages` | 21 | 18 | 16 |
| `install_runc` | 17 | 17 | 2 |
| `install_crun` | 45 | 49 | 45 |
| `install_testdeps` | 75 | 67 | 10 |
| **whole step** | **160** | **160** | **84** |
| **whole step, range** | **152-415** | **146-162** | **79-92** |

Both halves of the PR show up here, and they are worth reading separately.

The cache is the larger effect once it is warm: `install_testdeps` 75s to 10s, `install_runc` 17s to 2s, the whole step roughly halved. It restores in 4-5 seconds and never makes anything slower. The usual caveat applies, and it is the same one recorded earlier in this PR: attempt 2 is a re-run of attempt 1's commit, so the cri-o tree it cloned is byte-identical to the one the cache was built against. That is the best case. On a new commit, a few hours further along cri-o's `main`, the earlier runs on this branch measured `install_testdeps` at 63-68s and `install_runc` at 15-18s -- still an improvement, but a much smaller one. Sell this as roughly half a minute off the median, not 76 seconds.

The shallow clone is worth less on the median (11s to 4s) and everything in the tail. Of the nine baseline jobs, two spent 253 and 165 seconds in the cri-o clone, which is what produced the 415s and 330s setup steps in the range above; `main`'s median is 160s. With `--depth 1` the step has stayed inside 146-162s cold and 79-92s warm. That tail was the actual complaint.

One thing this run turned up: **the cache expires between merges.** Attempt 1 missed on both the exact key and the `restore-keys` prefix -- the August entries had been evicted after seven days idle -- which is why its whole-step median is identical to `main`'s. At the current merge rate the first run after a quiet week gets no cache. It costs nothing when it misses, so this is a note, not an objection.

**Job totals** (`main` median → PR cold → PR warm):

| | `main` | PR, cold | PR, warm |
|---|---|---|---|
| `conmon` | 4.4m | 4.1m | 2.8m |
| `cri-o (1)` | 5.6m | 5.4m | 3.1m |
| `cri-o (0)` | 31.4m | 24.5m | 22.6m |

`cri-o (0)` is dominated by the bats suite rather than by setup, and that is where cri-o/cri-o#10285 lands: the parallel pass goes from 28.0-28.7m on `main` to 20.8-21.7m here, with the same suite, all green, no test needing the retry `test_runner.sh` configures.

kolyshkin:
### `JOBS` result

Three `cri-o (0)` runs on this branch, plus the baseline. All green, no test failed, none needed its retry.

| | `JOBS` | tests | parallel pass | per test | serial pass |
|---|---|---|---|---|---|
| `main` [33016232632](https://github.com/containers/conmon/actions/runs/33016232632) | 4 | 306 | 25.6m | 5.0s | 1.4m |
| run 2 [33205171374](https://github.com/containers/conmon/actions/runs/33205171374) | 4 | 315 | 31.1m | 5.9s | 1.6m |
| run 3 [33210235032](https://github.com/containers/conmon/actions/runs/33210235032) | 8 | 315 | **19.4m** | 3.7s | 1.4m |
| run 4 [33212882666](https://github.com/containers/conmon/actions/runs/33212882666) | 16 | 315 | 18.3m | 3.5s | 1.3m |

`nproc * 2` is where the win is: against run 2, which ran the same 315 tests, 31.1m to 19.4m, a 38% cut. `cri-o (0)` as a whole went 35.7m to 25.5m.

`nproc * 4` was tried and rejected. It bought 1.1 further minutes, which is well inside the noise -- the two `JOBS=4` runs above differ by 5.5 minutes on their own -- and there is no reason to run sixteen container test suites at once on a four-core runner for a difference that cannot be distinguished from variance. The multiplier stays at 2.

The caveat from the previous comment still applies: three runs is not a distribution, and this needs watching after it merges. What makes it credible for now is that the `JOBS=8` gain (11.7m) is twice the spread seen between two identically configured runs, and that per-test cost fell from 5.9s to 3.7s while the serial pass, which is unaffected by `JOBS`, stayed flat at 1.3-1.6m across all four.

Two other things worth recording, both arguing for more samples rather than fewer:

- Setup was 138s in run 2, 162s in run 3 and 167s in run 4, on a confirmed cache hit every time. Runner variance, not the cache.
- `make -C "$CRIO_DIR" all test-binaries` took 0.3m in run 2 and 1.6m in run 3, also on a cache hit. cri-o is cloned from its `main`, so it moves between runs and the Go cache goes progressively stale against it. Either way it is about a minute, and not the thing worth optimising.

kolyshkin:
Instead of the last patch, I opened an upstream PR: https://github.com/cri-o/cri-o/pull/10285 which is more natural.

But let's keep it for now and we can revert it later once the above PR will land.

jnovy:
LGTM

